PR #515, opened 2026-04-05
test: shrink test_train_ddp_sim to fix limit_val_batches under DDP sharding

### LINKED ISSUE #46: test: test_train_ddp_sim fails — limit_val_batches=0.1 yields 0 batches with small val set
## Summary

`tests/test_train.py::test_train_ddp_sim` fails during the sanity check with:

```
MisconfigurationException: You requested to check 0.1 of the `val_dataloader`
but 0.1 * 5 < 1. Please increase the `limit_val_batches` argument.
Try at least `limit_val_batches=0.2`
```

## Root cause

`conftest.py` sets `cfg.trainer.limit_val_batches = 0.1` for all train fixtures. The val dataloader only has 5 batches in the test dataset, so `floor(0.1 × 5) = 0`, which Lightning rejects.

`test_train_ddp_sim` overrides `devices=2` and `strategy="ddp_spawn"` but does not override `limit_val_batches`, so it inherits the too-small fraction.

## Fix

In `test_train_ddp_sim` (`tests/test_train.py:114`), add:

```python
cfg_train.trainer.limit_val_batches = 1  # integer = exactly 1 batch, avoids fraction < 1 error
```

Alternatively, raise `limit_val_batches` to `1` (int) in the base `cfg_train_global` fixture in `conftest.py` to fix all future tests that run with a small val set — but that changes behaviour for other tests, so the targeted override in `test_train_ddp_sim` is safer.

## Related

Also note the `conftest.py` originally had a boolean DictConfig guard bug (see separate issue / commit), which has been fixed in the current version using `cfg.get("callbacks")`.

## Steps to reproduce

```bash
pytest tests/test_train.py::test_train_ddp_sim -v
```

ktinubu:
## Summary

- Fixture default `limit_val_batches=0.1` yielded 0 val batches per rank under `ddp_spawn` with `devices=2` (10 val batches → 5/rank → `0.1 × 5 = 0.5`, which Lightning rejects).
- Override with integer `limit_{train,val,test}_batches=1`, and shrink model / batch / dataset to match the tiny-model test pattern. DDP-on-CPU now finishes in ~13s instead of chewing through the full ksin dataset for many minutes.
- Only `test_train_ddp_sim` needed fixing — every other validating test runs single-device (10 batches → 1.0, passes) or uses `fast_dev_run=True` (bypasses `limit_val_batches`). See analysis comment on #46.

Closes #46

## Test plan

- [x] `pytest tests/test_train.py::test_train_ddp_sim -v` — passes in 13s
- [x] `make test` — 157 passed, 1 skipped, no regressions
- [x] `make format` — clean (one unrelated docs/mdformat drift reverted)

### DIFF
--- a/tests/test_train.py
+++ b/tests/test_train.py
@@ -125,6 +125,19 @@ def test_train_ddp_sim(cfg_train: DictConfig) -> None:
         cfg_train.trainer.accelerator = "cpu"
         cfg_train.trainer.devices = 2
         cfg_train.trainer.strategy = "ddp_spawn"
+        # Integer limits avoid the `fraction * batches < 1` error when the
+        # fixture's fractions (0.01, 0.1) shrink too far under DDP sharding.
+        cfg_train.trainer.limit_train_batches = 1
+        cfg_train.trainer.limit_val_batches = 1
+        cfg_train.trainer.limit_test_batches = 1
+        # Shrink model, batch, and dataset to keep DDP-on-CPU fast.
+        cfg_train.data.batch_size = 2
+        cfg_train.data.signal_length = 64
+        cfg_train.data.train_val_test_sizes = [4, 4, 4]
+        cfg_train.model.net.channels = 4
+        cfg_train.model.net.encoder_blocks = 1
+        cfg_train.model.net.trunk_blocks = 1
+        cfg_train.model.net.hidden_dim = 32
     train(cfg_train)
 
 


ktinubu:
## Verification Results

- [x] **`test_train_ddp_sim` passes end-to-end (fixes the `MisconfigurationException`)** *(Level 1 — runs the actual pytest invocation from the issue's reproduction steps, which is the full end-to-end promise)*

  ```bash
  $ /Users/k/test_projects/synth-setter/.venv/bin/pytest tests/test_train.py::test_train_ddp_sim -v
  tests/test_train.py::test_train_ddp_sim PASSED                    [100%]
  ============================== 1 passed in 13.14s ==============================
  ```

- [x] **DDP genuinely spawns 2 CPU ranks (the test's stated purpose is preserved after shrinking model/data)** *(Level 1 — adversarial; a tiny dataset could make Lightning silently fall back to single-process if misconfigured, so verify the 2-process DDP group actually initializes)*

  ```bash
  $ /Users/k/test_projects/synth-setter/.venv/bin/pytest tests/test_train.py::test_train_ddp_sim -v -s 2>&1 \
      | grep -E "(distributed processes|limit_val_batches)"
  `Trainer(limit_val_batches=1)` was configured so 1 batch will be used.
  All distributed processes registered. Starting with 2 processes
  All distributed processes registered. Starting with 2 processes
  ```

  Confirms: the `limit_val_batches=1` override is applied, and both DDP ranks register (line appears once per rank).

- [x] **Before/after: test goes from FAIL → PASS, runtime is comparable** *(Level 2 — promise is correctness and practical CPU runtime; PR's stated implied perf goal is "doesn't balloon into a multi-minute slog")*

  ```bash
  # Before (main @ baa51de): fails fast on the exact reproduction from #46
  $ /usr/bin/time \
      /Users/k/test_projects/synth-setter/.venv/bin/pytest tests/test_train.py::test_train_ddp_sim
  lightning.fabric.utilities.exceptions.MisconfigurationException: You requested to check 0.1 of the `val_dataloader` but 0.1 * 5 < 1. Please increase the `limit_val_batches` argument. Try at least `limit_val_batches=0.2`
  FAILED tests/test_train.py::test_train_ddp_sim
         13.39 real        13.74 user         2.86 sys

  # After (PR branch @ 1b0247d): passes through full fit+validate+test
  $ /usr/bin/time /Users/k/test_projects/synth-setter/.venv/bin/pytest tests/test_train.py::test_train_ddp_sim
  ============================== 1 passed in 10.80s ==============================
         14.75 real        20.08 user         3.18 sys
  ```

  Delta: `13.39s FAIL` → `14.75s PASS`. Correctness is gained for ~1.4s of additional wall-clock. (Note: without the model/data shrink, the intermediate `limit_val_batches=1`-only version was observed running 30+ minutes at 220% CPU per rank on the full ksin dataset before being killed — not re-measured here because the main branch already demonstrates the starting state.)

Issue #538, opened 2026-04-12
fix(code-health): devcontainer GPU passthrough, submodule trust, non-root user


ktinubu:
The `.devcontainer/` setup added in #186 is broken on three independent axes. All three need to be fixed before the devcontainer is usable for day-to-day work.

## 1. GPU passthrough is missing

The base image `tinaudio/perm:dev-snapshot` has a custom entrypoint gated on a `MODE` env var (valid values: `idle`, `passthrough`, `generate_dataset`). The current `.devcontainer/devcontainer.json` sets `MODE=idle`, and nothing passes `--gpus all` to Docker. Result: even on a GPU host, the container starts in idle mode with no GPU access.

**Fix:** Add `"runArgs": ["--gpus", "all"]` and set `MODE=passthrough` via `containerEnv`.

## 2. Submodule trust is broken

`.devcontainer/post-create.sh` whitelists only the main repo with `git config --global --add safe.directory "$(pwd)"`. But the script then runs `git submodule update --init --recursive`, which recurses into `.claude/skills` (a separate git repo from `tinaudio/skills`). That repo isn't on the safe-directory list, so git refuses:

```
fatal: detected dubious ownership in repository at '/workspaces/synth-setter/.claude/skills'
```

**Fix:** Replace the line with `git config --global --add safe.directory '*'`. Safe inside a dev container where the repo is the only thing present, and future-proofs for any new submodules.

## 3. Container runs as root, blocks Claude Code

Claude Code refuses to run with `--dangerously-skip-permissions` as root. The current devcontainer uses `"remoteUser": "root"`, so the standard Claude Code workflow inside the trusted sandbox doesn't work.

**Fix:** Switch to the [official VS Code non-root user pattern](https://code.visualstudio.com/remote/advancedcontainers/add-nonroot-user). Add a `.devcontainer/Dockerfile` that extends the base image with a `dev` user (uid 1000) and passwordless sudo, and update `devcontainer.json` to build from the Dockerfile with `"remoteUser": "dev"` and `"updateRemoteUserUID": true`. Note: the `/venv` directory baked into the base image must be chowned to the dev user so `uv pip install --no-deps -e .` in post-create can write to it.

## Acceptance

- Container rebuilds and starts without errors on a Linux host with NVIDIA Container Toolkit.
- `nvidia-smi` works inside the container.
- `git submodule update --init --recursive` succeeds.
- `whoami` returns `dev`; `sudo -n whoami` returns `root`.
[...]

Issue #186, opened 2026-03-21
chore(code-health): add .devcontainer for one-click onboarding


ktinubu:
## Summary
Add `.devcontainer/` configuration for VS Code Dev Containers and GitHub Codespaces — one-click reproducible development environment.

## Rationale
New contributors must manually set up Python, dependencies, and configs. Devcontainer eliminates "works on my machine" problems and enables instant onboarding.

## Pros
- One-click setup for new contributors
- Reproducible dev environment (matches CI)
- Works with GitHub Codespaces for cloud development
- Works with VS Code, JetBrains

## Cons
- Docker required locally (not needed if using Codespaces)
- Container image maintenance (minimal — rebuild on Dockerfile changes)
- May not support GPU workflows out of the box

## Blocking
- Blocked by #180 (Makefile expansion — devcontainer setup uses `make install`)

## Verification
- `devcontainer build .` succeeds
- Opening in VS Code "Reopen in Container" works
- `make test` passes inside the container

ktinubu:
## Bug: devcontainer breaks in git worktrees (and any dir not named `synth-setter`)

Found while verifying the codespace config locally from a worktree.

### Reproduction

```bash
git worktree add .claude/worktrees/verify-codespace feat/codespaces-devcontainer
DOCKER_DEFAULT_PLATFORM=linux/amd64 devcontainer up \
  --workspace-folder .claude/worktrees/verify-codespace
```

Container starts, then `postCreateCommand` fails:

```
OCI runtime exec failed: exec failed: unable to start container process:
  chdir to cwd ("/workspaces/synth-setter") set in config.json failed:
  no such file or directory
postCreateCommand from devcontainer.json failed with exit code 127.
```

### Root cause

Two paths are hardcoded to `/workspaces/synth-setter`:

- [`.devcontainer/devcontainer.json:10`](https://github.com/tinaudio/synth-setter/blob/feat/codespaces-devcontainer/.devcontainer/devcontainer.json#L10) — `"workspaceFolder": "/workspaces/synth-setter"`
- [`.devcontainer/post-create.sh:8`](https://github.com/tinaudio/synth-setter/blob/feat/codespaces-devcontainer/.devcontainer/post-create.sh#L8) — `cd /workspaces/synth-setter`

[...]

PR #541, opened 2026-04-12
fix(devcontainer): GPU passthrough, submodule trust, non-root dev user

ktinubu:
## Summary

The `.devcontainer/` added in #186 was unusable on three independent axes. This PR fixes all three so the container starts cleanly with GPUs attached and Claude Code can run inside it.

- **GPU passthrough.** `runArgs: ["--gpus", "all"]` plus `MODE=passthrough` (replacing `idle`) so the base image entrypoint starts a usable shell with GPUs attached. The old config was effectively no-op on GPU hosts.
- **Submodule trust.** `post-create.sh` now widens `git safe.directory` to `'*'`. The previous per-path entry only covered the main repo, so `git submodule update --init --recursive` died on `.claude/skills` with *"detected dubious ownership"*. The wildcard is safe inside a dev container where the repo is the only thing present.
- **Non-root dev user.** New `.devcontainer/Dockerfile` extends the base image with a `dev` user (uid 1000) and passwordless sudo — the [official VS Code pattern](https://code.visualstudio.com/remote/advancedcontainers/add-nonroot-user). `devcontainer.json` builds from it with `remoteUser: dev` and `updateRemoteUserUID: true`. The Dockerfile also `chown`s `/venv` so post-create's `uv pip install --no-deps -e .` can write to the pre-baked venv as the dev user. This unblocks `claude --dangerously-skip-permissions`, which refuses to run as root.

## Test plan

Run on a Linux host with NVIDIA Container Toolkit (the only environment where `--gpus all` is meaningful):

- [ ] **Rebuild container** — "Dev Containers: Rebuild Container" in VS Code, or `devcontainer up --workspace-folder .` with the CLI. Build must succeed.
- [ ] **Entrypoint starts** — container boots without the `MODE` entrypoint erroring; `postCreateCommand` runs to completion.
- [ ] **Identity** — `whoami` → `dev`; `id` shows `uid=1000(dev)`; `sudo -n whoami` → `root`.
- [ ] **GPU passthrough** — `nvidia-smi` lists at least one GPU inside the container.
- [ ] **Submodules** — `ls .claude/skills/` is non-empty; no "dubious ownership" error in post-create logs.
- [ ] **Venv writable** — `uv pip install --no-deps -e .` from post-create completes without permission error.
- [ ] **Pre-commit wired** — `cat .git/hooks/pre-commit` exists.
- [ ] **Tests smoke** — `make test` runs as the `dev` user.
- [ ] **Claude Code** — `claude --dangerously-skip-permissions` no longer rejects on the root check.

## Notes

- `--gpus all` requires NVIDIA Container Toolkit on the Docker host. Mac Docker Desktop doesn't support it, so this devcontainer must be opened on a Linux host or a Codespaces machine type that has a GPU.
- `chown -R /venv` in the Dockerfile is load-bearing — without it, `uv pip install` in post-create fails with EACCES against root-owned site-packages.

Refs #538

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,14 @@
+FROM tinaudio/perm:dev-snapshot
+
+ARG USERNAME=dev
+ARG USER_UID=1000
+ARG USER_GID=$USER_UID
+
+RUN groupadd --gid $USER_GID $USERNAME \
+    && useradd --uid $USER_UID --gid $USER_GID -m $USERNAME \
+    && apt-get update \
+    && apt-get install -y sudo \
+    && rm -rf /var/lib/apt/lists/* \
+    && echo "$USERNAME ALL=(root) NOPASSWD:ALL" > /etc/sudoers.d/$USERNAME \
+    && chmod 0440 /etc/sudoers.d/$USERNAME \
+    && chown -R $USER_UID:$USER_GID /venv
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,12 +1,14 @@
 {
   "name": "synth-setter",
-  "image": "tinaudio/perm:dev-snapshot",
+  "build": { "dockerfile": "Dockerfile" },
 
+  "runArgs": ["--gpus", "all"],
   "containerEnv": {
-    "MODE": "idle"
+    "MODE": "passthrough"
   },
 
-  "remoteUser": "root",
+  "remoteUser": "dev",
+  "updateRemoteUserUID": true,
   "workspaceFolder": "/workspaces/${localWorkspaceFolderBasename}",
 
   "postCreateCommand": ".devcontainer/post-create.sh",
--- a/.devcontainer/post-create.sh
+++ b/.devcontainer/post-create.sh
@@ -22,10 +22,11 @@ done
 }
 cd "$dir"
 
-# Codespaces runs this script as root against a workspace that may be owned
-# by another UID, tripping git's safe.directory check. Mark the repo trusted
-# before any git operation (submodule update, pre-commit install).
-git config --global --add safe.directory "$(pwd)"
+# Trust the workspace repo and every submodule (e.g. .claude/skills) before
+# any git operation (submodule update, pre-commit install). The wildcard is
+# safe inside a dev container where the repo is the only thing present, and
+# avoids per-submodule entries that broke when .claude/skills was added.
+git config --global --add safe.directory '*'
 
 # Skills live in a git submodule (.claude/skills → tinaudio/skills).
 git submodule update --init --recursive


ktinubu:
## Verification Results

Ran on the PR branch (commit 52e6add) against a locally-built devcontainer image (`synth-setter-devcontainer-test:pr541`) derived from the PR's Dockerfile, on macOS / Docker Desktop / `linux/amd64` emulation. The base image `tinaudio/perm:dev-snapshot` was already cached locally.

- [x] **Container starts as `dev` user (uid 1000) and has working passwordless sudo** *(Level 1 — built the PR's Dockerfile and exercised the user directly)*
  ```bash
  $ docker run --rm --platform linux/amd64 --entrypoint /bin/bash synth-setter-devcontainer-test:pr541 -c '
      su dev -c "whoami; id"
      su dev -c "sudo -n whoami"
      cat /etc/sudoers.d/dev
    '
  dev
  uid=1000(dev) gid=1000(dev) groups=1000(dev)
  root
  dev ALL=(root) NOPASSWD:ALL
  ```
  Proves the Dockerfile creates the user, installs sudo, and the NOPASSWD rule is in effect.

- [x] **`/venv` is chowned to `dev:dev` and the dev user can write to it** *(Level 1 — required so `uv pip install --no-deps -e .` in post-create.sh works as non-root)*
  ```bash
  $ docker run --rm --platform linux/amd64 --entrypoint /bin/bash synth-setter-devcontainer-test:pr541 -c '
      stat -c "%U:%G %n" /venv /venv/main
      su dev -c "touch /venv/main/.devuser-writetest && rm /venv/main/.devuser-writetest && echo write OK"
    '
  dev:dev /venv
  dev:dev /venv/main
  write OK
  ```

- [x] **`MODE=passthrough` executes the supplied command; unset MODE errors; empty-command passthrough exits 0** *(Level 1 — exercises the base image's custom entrypoint directly on the three MODE paths relevant to this PR)*
  ```bash
  $ docker run --rm --platform linux/amd64 -e MODE=passthrough synth-setter-devcontainer-test:pr541 bash -c 'echo HELLO_FROM_PASSTHROUGH; whoami'
  HELLO_FROM_PASSTHROUGH
  root

  $ docker run --rm --platform linux/amd64 -e MODE=passthrough synth-setter-devcontainer-test:pr541
  Passthrough mode — no command provided, exiting cleanly.

  $ docker run --rm --platform linux/amd64 synth-setter-devcontainer-test:pr541
  Error: MODE is required. Set MODE=idle, MODE=passthrough, or MODE=generate_dataset.
  Available modes: idle, passthrough, generate_dataset
  (exit 1)
  ```
  Confirms `MODE=passthrough` is the value that lets VS Code's default container command actually run, which is the whole point of the containerEnv change.

- [x] **`.devcontainer/post-create.sh` runs end-to-end as the `dev` user against a real workspace** *(Level 1 — the load-bearing check; cloned the PR branch into /tmp, mounted it as /workspaces/synth-setter, and ran the actual script the devcontainer will run)*
  ```bash
  $ git clone --local --recurse-submodules \
      /path/to/worktree /tmp/pr541-post-create-test
  $ docker run --rm --platform linux/amd64 \
      -v /tmp/pr541-post-create-test:/workspaces/synth-setter \
      -w /workspaces/synth-setter --user dev -e MODE=passthrough \
      synth-setter-devcontainer-test:pr541 \
      bash -c 'bash .devcontainer/post-create.sh; ls .claude/skills | head -2; stat -c %U /venv/main; cat ~/.gitconfig'
  ...
  Building synth-permutations @ file:///workspaces/synth-setter
  Installed 1 package in 3ms
   - synth-permutations==0.1.4 (from file:///home/build/synth-setter)
   + synth-permutations==0.2.0 (from file:///workspaces/synth-setter)
  pre-commit installed at .git/hooks/pre-commit
  Dev container ready. Run 'make test' to verify.
  .claude-plugin
  .git
  dev
  [safe]
          directory = *
  ```
  What this proves, in one run:
  - `git submodule update --init --recursive` populates `.claude/skills` without the dubious-ownership error (the submodule's contents are listed).
  - `uv pip install --no-deps -e .` writes to `/venv/main` as the dev user without permission errors (the chown is load-bearing).
  - `pre-commit install` succeeds and writes `.git/hooks/pre-commit`.
  - `git config --global --add safe.directory '*'` lands in `~/.gitconfig` with the wildcard.
  - The script reaches its final `Dev container ready.` line.

  One pre-existing noise: `plumb init` inside the script fails with `ImportError: cannot import name 'XMLAdapter' from 'dspy.adapters'`. This is an upstream plumb/dspy bug unrelated to this PR. The script already handles it defensively with `plumb init || true` on line 45, so the overall run succeeds. Worth a separate issue.

- [x] **`git safe.directory='*'` is syntactically valid and honored by git for arbitrary distinct repo paths** *(Level 3 — sandboxed two unrelated repos on the host and exercised git directly with the new config)*
  ```bash
  $ GIT_CONFIG_GLOBAL=$HOME_FAKE/.gitconfig git config --global --add safe.directory '*'
  $ cat $HOME_FAKE/.gitconfig
  [safe]
          directory = *
  $ GIT_CONFIG_GLOBAL=$HOME_FAKE/.gitconfig git -C parent log --oneline -1
  9002ba8 init
  $ GIT_CONFIG_GLOBAL=$HOME_FAKE/.gitconfig git -C child log --oneline -1
  9002ba8 init
  ```
  Note: I could not reproduce the *cross-UID dubious-ownership* failure on macOS because Docker Desktop's gRPC FUSE bind mounts remap host UIDs to the container user, so `dev` (uid 1000) sees the mounted files as its own regardless of host ownership (501). That specific failure mode only bites on Linux hosts (including Codespaces) where bind mounts preserve host UIDs. The script-level check above (post-create.sh running end-to-end) confirms the config lands, and the sandbox check above confirms git accepts the wildcard.

- [x] **`devcontainer.json` is well-formed and the devcontainer CLI resolves all new fields** *(Level 2 — invoked the same CLI that VS Code Dev Containers uses to read the file)*
  ```bash
  $ devcontainer read-configuration --workspace-folder . 2>&1 | tail -1 > /tmp/dc.json
  $ python3 -c "import json; c=json.load(open('/tmp/dc.json'))['configuration']; print(*((k, c.get(k)) for k in ['build','runArgs','containerEnv','remoteUser','updateRemoteUserUID','image']), sep='\n')"
  ('build', {'dockerfile': 'Dockerfile'})
  ('runArgs', ['--gpus', 'all'])
  ('containerEnv', {'MODE': 'passthrough'})
  ('remoteUser', 'dev')
  ('updateRemoteUserUID', True)
  ('image', None)
  ```

### Not runnable on this host

- [ ] **`--gpus all` actually passes a GPU into the running container** *(SKIP — verification host is macOS Docker Desktop, no NVIDIA GPU. Attempting `docker run --gpus all ...` fails with "could not select device driver" before the container starts.)*
  Recipe for a Linux host with NVIDIA Container Toolkit installed:
  ```bash
  cd /path/to/synth-setter
  git fetch origin fix/devcontainer-gpu-submodule-nonroot
  git worktree add /tmp/pr541 origin/fix/devcontainer-gpu-submodule-nonroot
  cd /tmp/pr541
  devcontainer up --workspace-folder .
  devcontainer exec --workspace-folder . nvidia-smi
  # Expect: at least one GPU row in the output
  ```

- [ ] **Cross-UID dubious-ownership bug is fixed on a real Linux host** *(SKIP — failure mode not reproducible on macOS Docker Desktop, see above)*
  Recipe for a Linux host:
  ```bash
  # On a Linux host where the workspace dir is owned by uid != 1000:
  sudo chown -R 501:501 /path/to/synth-setter    # or any uid != 1000
  devcontainer up --workspace-folder /path/to/synth-setter
  # Post-create should complete without the "dubious ownership in .claude/skills" error.
  # Reverting post-create.sh:28 to the old per-path form should make it fail again.
  ```

- [ ] **VS Code "Rebuild Container" completes end-to-end** *(SKIP — requires an interactive VS Code session; the devcontainer CLI above exercises the same config resolution path headlessly, but not the VS Code UI hooks)*
  Recipe: open the folder in VS Code on a Linux GPU host → Command Palette → "Dev Containers: Rebuild Container" → confirm no errors in the Dev Containers output channel and the integrated terminal opens as `dev@...`.

### Summary

Everything that can be verified without a GPU and without a Linux host is verified Level 1. The three SKIPs are environment-limited — they all require a real Linux GPU host to exercise the specific failure modes they cover — and each has a copy-pasteable recipe for the user to run on the target environment. The strongest single signal is the end-to-end post-create.sh run inside the built image as the dev user: it exercises the submodule trust, the `/venv` chown, the editable install, and the pre-commit wiring in one shot.

Issue #672, opened 2026-04-23
fix(eval): UnboundLocalError in predict_vst_audio.py default path


ktinubu:
## Bug

`scripts/predict_vst_audio.py:220` raises `UnboundLocalError: local variable 'target_synth_params' referenced before assignment` when running the default code path (without `-t`/`--rerender_target`).

## Root cause

`target_synth_params` and `target_note_params` are only defined inside the `if rerender_target:` branch (line 183), but the subsequent `params_to_csv` call at line 220 references them in a ternary:

```python
params_to_csv(
    target_synth_params if target_params is not None else None,  # <- NameError when rerender_target=False
    target_note_params if target_params is not None else None,
    synth_params,
    note_params,
    ...
)
```

The ternary evaluates `target_synth_params` eagerly whenever `target_params is not None` — which is always true unless `--no-params` is set.

## Repro

```bash
python scripts/predict_vst_audio.py <pred_dir> <out_dir>
```

With any normally-produced predictions directory (containing `target-params-N.pt` files), this raises `UnboundLocalError` on the first sample.

## Workarounds (any of)

- `-t` (`--rerender_target`) — defines `target_synth_params` via the rerender branch
- `-X` (`--no-params`) — sets `target_params = None`, short-circuiting the ternary

## Fix sketch

Either:
1. Move `target_synth_params` / `target_note_params` initialization (to `None`, or decoded from `target_params`) above the ternary so they are always defined, OR
2. Restructure `params_to_csv` to accept `target_params` directly and do decoding internally.

## Discovered in

`tests/test_train.py::test_train_eval_surge_xt` — works around with `-t` and flags this issue.

ktinubu:
Closing as complete. Verified by multi-agent triage+audit+close-verify pass (tracking #1734).

## What was asked
Fix `UnboundLocalError: local variable 'target_synth_params' referenced before assignment` in `predict_vst_audio.py` on the default code path (without `-t`/`--rerender_target`).

## What landed
PR #1148 (commit `651d0619`) — fix predict_vst_audio UnboundLocalError when target params present without -t.

- `target_synth_params: dict[str, float] | None = None` initialized at `src/synth_setter/evaluation/predict_vst_audio.py:196`
- `target_note_params: NoteParams | None = None` initialized at `:197`
- Both variables are now always defined before the `params_to_csv` call path

## Verification
- [x] Commit on `main`: `651d0619` reachable (`git merge-base --is-ancestor` exit 0)
[...]

Issue #155, opened 2026-03-20
Phase 2: Coverage & stability


ktinubu:
## Phase 2: Coverage & stability

**Parent Epic:** #149

Coverage targets, flaky test fixes, and resource-aware test gating.

### Tasks
- Test coverage ≥80% for src/ (#30)
- Fix val_check_interval flaky test (#47)
- Fix limit_val_batches DDP test (#46)
- Add RAM gate to skip memory-intensive tests (#40)
- Configure test-full.yml as merge gate (#35)

Issue #149, opened 2026-03-20
Epic: test infrastructure & coverage


ktinubu:
## Epic: test infrastructure & coverage

**Design doc:** N/A — cross-cutting infrastructure

Build comprehensive test infrastructure covering device compatibility, coverage targets, and platform support.

### Phases

- Phase 1: Device compatibility — GPU runners, MPS config, accelerator flag
- Phase 2: Coverage & stability — 80% coverage target, fix flaky tests, RAM gate
- Phase 3: Platform support — Windows CI, dependency modernization

Issue #137, opened 2026-03-20
Phase 1: Portable Evaluation Pipeline


ktinubu:
Phase 1 of the evaluation pipeline epic (#98). Config cleanup, portable predict/render/metrics stages.

**Design doc:** eval-pipeline.md §8, Phase 1
**Tasks:** #94, #85, #86, #87

PR #674, opened 2026-04-23
test(surge): parametrized Surge XT train+eval e2e (cpu/mps/gpu)

### LINKED ISSUE #673: test: add Surge XT one-step train and end-to-end eval tests
## Summary

Add GPU-gated smoke tests that exercise the full Surge XT training and evaluation pipeline on the 5-sample HDF5 fixture.

## Tests

- `tests/test_train.py::test_train_surge_xt_one_step` — train the `experiment=surge/flow_full` model for exactly one step; assert `trainer.global_step == 1`.
- `tests/test_train.py::test_train_eval_surge_xt` — train one step → load `last.ckpt` → run `evaluate(cfg)` in `mode=predict` → render predicted params through the Surge XT VST → compute audio distance metrics (MSS, wMFCC, SOT, RMS). Assert the shape of `predictions/`, `audio/sample_*/{target,pred}.wav + spec.png + params.csv`, `metrics/metrics.csv`, and `metrics/aggregated_metrics.csv`.

## Fixtures (`tests/conftest.py`)

- `cfg_surge_xt_global` — package-scoped compose of `train.yaml` with `experiment=surge/flow_full`. Pins the 5-sample fixture dataset root, disables saved normalization (stats.npz has zero-std bins at N=5), nulls out `predict_file` (configs/data/surge.yaml hardcodes a researcher-local path), and bakes in the 1-step trainer knobs.
- `cfg_surge_xt` — function-scoped `tmp_path` wrapper.
- `cfg_surge_xt_eval` — function-scoped eval config composed from `eval.yaml`, copying `data` and `model` from `cfg_surge_xt` so the evaluator matches the trained checkpoint shape. Uses the default `callbacks: eval_surge` (= `prediction_writer`) instead of the train-time `default_surge` bundle.

## Motivation

The repo has no existing end-to-end smoke test for the Surge XT model — `test_train_fast_dev_run_*` exercises the KSin FF model only. Wiring up the full train → predict → VST-render → audio-metrics chain on a tiny fixture gives us a fast regression gate for the flow-matching training path and the eval tooling.

## Related

- Blocks nothing.
- Surfaced #672 (UnboundLocalError in `predict_vst_audio.py` default path) — worked around in the new test with `-t`.

ktinubu:
## Summary

- Adds `test_train_surge_xt` and `test_train_eval_surge_xt` — one-step Surge XT train, then eval round-trip (load `last.ckpt`, predict, render via VST, score with `compute_audio_metrics`). Both are now `@pytest.mark.slow`.
- **Parametrizes both Surge tests over `cpu` / `mps` / `gpu`** via a new `accelerator` fixture. Each test expands to three IDs (`...[cpu]`, `[mps]`, `[gpu]`) with matching markers. The fixture **hardfails** (not skips) when an explicitly-requested accelerator isn't available, so a misconfigured runner can't pass silently.
- **Splits CI by accelerator** so each runner only collects the cases it can run.
- **Hardens two subprocess call sites** that were exposed to a `capture_output=True` deadlock once the Surge fixtures dropped from package- to function-scope (forced by depending on the parametrized fixture). See #695.

## Fixtures (`tests/conftest.py`)

- `accelerator` — parametrized `cpu` / `mps` / `gpu` with per-param markers; calls `pytest.fail("CUDA/MPS not available", pytrace=False)` when a runner is misconfigured for the requested device.
- `cfg_surge_xt_global` — composes `experiment=surge/ffn_full` + `callbacks=eval_surge`; reads `accelerator` from the fixture above and applies device-specific tweaks (`precision="32-true"` for MPS, `precision="16-mixed"` + `compile=False` for GPU). **Function-scoped** because parametrized fixtures force function scope on consumers.
- `cfg_surge_xt`, `cfg_surge_xt_eval` — per-test wrappers that point output dirs at `tmp_path` and inherit the parametrization transitively.
- `surge_xt_smoke_datasets` — generates the 5-sample HDF5 fixture by invoking `src/data/vst/generate_vst_dataset.py` through `scripts/run-linux-vst-headless.sh` on Linux. Now uses `capture_output=False` (default) — child inherits stdout/stderr, no pipe, no deadlock risk.

## Subprocess hardening — `Refs #695`

The conftest fixture and the VST audio render call in `test_train_eval_surge_xt` previously used `capture_output=True` / `subprocess.check_call`. Under pytest-xdist + DataLoader workers, fork-inherited fds can keep the pipe's read end open after the child exits, leaving `communicate()` blocked on EOF that never arrives — pytest produces no output, no traceback, no timeout. Both call sites now inherit stdout/stderr (no PIPE) and report failures via `pytest.fail(..., pytrace=False)` with the exit code. The audit of remaining call sites lives in #695.

## CI changes

- **New** `.github/workflows/test-mps.yml` — macOS Apple Silicon runner. Triggers on `push`/`pull_request` paths in `src/**`, `configs/**`. Asserts MPS availability before pytest, then runs `-m mps`.
- **`test-gpu.yml`** — added the Linux Surge XT VST3 + headless X11 stack (Xvfb / xsettingsd / openbox / dbus-x11 / mesa) so `test_train_eval_surge_xt[gpu]` can run.
- **`test-expensive.yml`** (slow tests) — same VST + X11 install added; now picks up `[cpu]` cases of the Surge tests via `-m \"slow and not gpu\"`.
- **`test.yml`** (fast tests, ubuntu + macos) — VST install removed from both jobs; filters tightened to `-m \"not slow and not gpu and not mps\"`. The fast suite no longer needs the VST plugin.
- **`pyproject.toml`** — registered `mps` marker.

## Selection cheat-sheet

| Want | Command |
|---|---|
| Just CPU Surge | `pytest -m \"slow and not gpu and not mps\" -k surge` |
| Just MPS | `pytest -m mps` |
| Just GPU | `pytest -m gpu` |
| Single ID | `pytest \"tests/test_train.py::test_train_eval_surge_xt[gpu]\"` |

## Implementation notes

- `data.batch_size = 1` is forced — `ShiftedBatchSampler` drops one batch per epoch, so anything larger than `dataset_size // 2` empties the train dataloader.
- `data.use_saved_mean_and_variance = False` — the 5-sample `stats.npz` has zero-std mel bins that NaN-poison normalization.
- VST plugin is symlinked into `plugins/Surge XT.vst3` on each runner; `plugins/` is gitignored.
- `predict_vst_audio.py` invoked with `-t` (`--rerender_target`) as a workaround for #672.

## Test plan

- [x] `pytest \"tests/test_train.py::test_train_eval_surge_xt[cpu]\"` — passes locally (slow path)
- [x] `pytest tests/test_train.py::test_train_eval_surge_xt --collect-only` — collects three IDs (`[cpu]`, `[mps]`, `[gpu]`)
- [x] Marker filtering verified: `-m gpu`, `-m mps`, `-m \"not gpu and not mps\"` each select the expected single ID
- [x] Conftest subprocess no longer hangs when fixture is function-scoped (was the trigger for the whole investigation)
- [x] `make format` — all pre-commit hooks green
- [ ] `test.yml` (Ubuntu + macOS) green on this PR — runs the fast suite without VST install
- [ ] `test-expensive.yml` green — runs `[cpu]` Surge cases with the new VST install step
- [ ] `test-gpu.yml` green on next scheduled run / manual dispatch — runs `[gpu]` cases
- [ ] `test-mps.yml` green on macOS — runs `[mps]` cases

Closes #673
Refs #672
Refs #675
Refs #695


ktinubu:
## Verification Results

Ran on GPU host (NVIDIA RTX 5060 Ti, 15701 MiB free) at commit `960e2c4`.

### Behavioral checks

- [x] **`test_train_surge_xt_one_step` completes one training step on the 5-sample fixture** *(Level 1 — the test IS the end-to-end smoke promise)*
  ```bash
  $ python -m pytest tests/test_train.py::test_train_surge_xt_one_step -v 2>&1 | tail -5
  3.26s call     tests/test_train.py::test_train_surge_xt_one_step
  ======================== 1 passed, 1 warning in 3.91s ========================
  ```

- [x] **`test_train_eval_surge_xt` runs the full train → predict → VST render → audio-metrics chain** *(Level 1 — promise is end-to-end pipeline; asserts predictions/, audio/sample_*/, metrics.csv, aggregated_metrics.csv)*
  ```bash
  $ python -m pytest tests/test_train.py::test_train_eval_surge_xt -v 2>&1 | tail -5
  13.11s call     tests/test_train.py::test_train_eval_surge_xt
  ======================== 1 passed, 1 warning in 16.21s ========================
  ```

- [x] **Both tests in sequence do not contaminate each other's fixture state** *(Level 1 — adversarial; `cfg_surge_xt_global` is package-scoped so shared-state mutation would surface here)*
  ```bash
  $ python -m pytest tests/test_train.py::test_train_surge_xt_one_step tests/test_train.py::test_train_eval_surge_xt -v 2>&1 | tail -4
  13.15s call     tests/test_train.py::test_train_eval_surge_xt
  2.96s call     tests/test_train.py::test_train_surge_xt_one_step
  ======================== 2 passed, 2 warnings in 17.14s ========================
  ```

- [x] **Lazy-import fix restored conda-env collection for `tests/test_train.py`** *(Level 2 — adversarial; first push broke collection because `scripts.compute_audio_metrics` transitively loads torchaudio which fails binary-load in conda)*
  ```bash
  $ python -m pytest tests/test_train.py --collect-only -q 2>&1 | tail -3
  tests/test_train.py::test_train_surge_xt_one_step
  tests/test_train.py::test_train_eval_surge_xt
  9 tests collected in 0.39s
  ```

### Taxonomy & platform state

- [x] **Linked issues carry required metadata (type + domain label + milestone)** *(Level 4 — live GitHub API)*
  ```bash
  $ gh api graphql -f query='...' --jq '...' | jq .
  {
    "task": {"num": 673, "type": "Task", "labels": ["testing"], "milestone": "testing v1.0.0"},
    "bug":  {"num": 672, "type": "Bug",  "labels": ["evaluation"], "milestone": "evaluation v1.0.0"}
  }
  ```

- [x] **Both linked issues trace to an Epic via sub-issue hierarchy** *(Level 4 — adversarial; walks parent chain)*
  ```
  #672 [Bug]  → #137 [Phase 1: Portable Evaluation Pipeline] → #98  [Epic: evaluation pipeline]
  #673 [Task] → #155 [Phase 2: Coverage & stability]         → #149 [Epic: test infrastructure & coverage]
  ```

- [x] **`check-pr-metadata` CI gate passes on this PR** *(Level 2 — the actual gate runs against the PR)*
  ```bash
  $ gh pr checks 674 --repo tinaudio/synth-setter | grep check-pr-metadata
  check-pr-metadata	pass	12s	https://github.com/tinaudio/synth-setter/actions/runs/24824615683/job/72657293333
  ```

- [x] **PR body links issues with correct auto-close semantics** *(Level 5 — grep on PR body; appropriate here since the claim is about string content)*
  ```bash
  $ gh pr view 674 --json body --jq '.body' | grep -E "Closes|Refs|Fixes"
  Closes #673
  Refs #672
  ```

### Repo hygiene

- [x] **Test-created `plugins/Surge XT.vst3` symlink is gitignored (no commit pollution)** *(Level 4 — live git index check)*
  ```bash
  $ git status --short | grep -v 'tests/fixtures/' | grep -E 'plugins|Surge' ; echo exit=$?
  exit=1
  $ git check-ignore -v plugins/
  .gitignore:165:plugins/	plugins/
  ```

### Skipped / pending

- [ ] **Full GPU CI workflow executes the new tests** *(Level 1 — SKIP: repo has no active GPU runner; tracked by #34. To verify manually once GPU CI lands:)*
  ```bash
  # Local reproduction on a GPU host:
  git fetch origin worktree-surge-xt-train-test
  git checkout worktree-surge-xt-train-test
  # Populate tests/fixtures/surge_xt/{train,val,test}.h5 + stats.npz (gitignored — needs a separate data-drop)
  python -m pytest tests/test_train.py::test_train_surge_xt_one_step tests/test_train.py::test_train_eval_surge_xt -v
  ```

- [ ] **`run_tests_conda` CI passes after lazy-import fix** *(in progress — fix pushed in `960e2c4`; waiting on CI cycle. Will update when conclusion lands.)*

### Notes

- `tests/fixtures/surge_xt/` contents are gitignored (`*.h5` and `stats.npz` currently not tracked). Tests will fail on any runner that hasn't materialized the fixture data. Worth a follow-up to either commit the fixture or auto-generate it in a conftest.
- `predict_vst_audio.py` must be invoked with `-t` to avoid #672; the workaround is inlined in the test with a pointer to the bug.

tests/conftest.py
```diff
@@ -134,3 +147,167 @@ def cfg_eval(cfg_eval_global: DictConfig, tmp_path: Path) -> DictConfig:
...
+    Composes `train.yaml` with `experiment=surge/flow_full` and bakes in the minimal overrides
+    needed to train-smoke-test on `tests/fixtures/surge_xt/`. Tests can override any knob.
+
+    :return: A DictConfig object configured for a one-step Surge XT smoke train.
+    """
+    with initialize(version_base="1.3", config_path="../configs"):
+        cfg = compose(
+            config_name="train.yaml",
+            return_hydra_config=True,
+            overrides=["experiment=surge/ffn_full", "callbacks=eval_surge"],
```
ktinubu: BLOCK: [code-health] `flow_full` vs `ffn_full` mismatch.

This line composes `experiment=surge/ffn_full`, but the surrounding docstring (line 156) says "composes `train.yaml` with `experiment=surge/flow_full`" and the public docs in `docs/reference/testing.md` line 81 say the fixture is for "Surge XT flow-matching tests". Both `configs/experiment/surge/flow_full.yaml` and `configs/experiment/surge/ffn_full.yaml` exist and are different experiments. A reader can't tell whether the FFN choice is intentional (and the docs are stale) or whether the override is wrong. Pick one and align the docstring + `docs/reference/testing.md` line 81 + the fixture docstring at line 154 ("flow-matching").

tests/conftest.py
```diff
@@ -134,3 +147,167 @@ def cfg_eval(cfg_eval_global: DictConfig, tmp_path: Path) -> DictConfig:
...
+
+        with open_dict(cfg):
+            cfg.paths.root_dir = str(rootutils.find_root(indicator=".project-root"))
+
+            cfg.data.dataset_root = "tests/fixtures/surge_xt"
+            # batch_size=1 is forced: the fixture has 5 samples and the
+            # ShiftedBatchSampler drops one batch per epoch, so any batch_size > 1
+            # leaves the dataloader empty and Lightning aborts with
+            # "Trainer.fit stopped: No training batches."
+            cfg.data.batch_size = 2
```
ktinubu: BLOCK: [code-health] Comment contradicts the code.

The comment says "batch_size=1 is forced ... any batch_size > 1 leaves the dataloader empty and Lightning aborts" — but the next line sets `batch_size = 2`. Either the comment is stale and `batch_size=2` works fine (delete the warning), or the comment is right and the value should be 1. The way it reads now, a reader has to test it themselves to know which is true.

tests/conftest.py
```diff
@@ -134,3 +147,167 @@ def cfg_eval(cfg_eval_global: DictConfig, tmp_path: Path) -> DictConfig:
...
+            cfg.data.pin_memory = True
+            cfg.data.ot = False
+            cfg.data.predict_file = "tests/fixtures/surge_xt/test.h5"
+            # The 5-sample fixture's stats.npz has zero-std mel bins that poison the batch
+            # with NaN via (mel - mean) / std.
+            cfg.data.use_saved_mean_and_variance = False
+
+            cfg.trainer.devices = 1
+            cfg.trainer.max_steps = 50
+            cfg.trainer.check_val_every_n_epoch = 50  # validate every 5 epochs
```
ktinubu: WARN: [code-health] Comment says "validate every 5 epochs" but the value is 50.

```suggestion
            cfg.trainer.check_val_every_n_epoch = 50  # validate at end of run
```

(With `max_steps=50` and `batch_size=2` over 5 samples, this is ~25 epochs total, so `check_val_every_n_epoch=50` means "only validate at the end" — which is fine for a smoke test, but the "5 epochs" comment suggests the original intent was different.)

tests/conftest.py
```diff
@@ -134,3 +147,167 @@ def cfg_eval(cfg_eval_global: DictConfig, tmp_path: Path) -> DictConfig:
...
+            cfg.trainer.val_check_interval = 1.0  # default: end of (validating) epoch
+            cfg.trainer.log_every_n_steps = 50
+            cfg.trainer.enable_model_summary = False
+            cfg.trainer.limit_val_batches = 1.0
+            cfg.trainer.accelerator = "gpu"
+            cfg.model.compile = True
```
ktinubu: WARN: [ml-test] `cfg.model.compile = True` in a smoke-test fixture.

`torch.compile` adds significant warmup (often minutes) on first call; the existing `cfg_train_global` keeps `compile=False` for exactly this reason. For a 50-step smoke test the compile cost dominates the runtime and gives nothing in return. Recommend `cfg.model.compile = False` here, with a separate slow-marked test for the compile path if you want coverage.

tests/conftest.py
```diff
@@ -134,3 +147,167 @@ def cfg_eval(cfg_eval_global: DictConfig, tmp_path: Path) -> DictConfig:
...
+        with open_dict(cfg):
+            cfg.paths.root_dir = str(rootutils.find_root(indicator=".project-root"))
+
+            cfg.data.dataset_root = "tests/fixtures/surge_xt"
+            # batch_size=1 is forced: the fixture has 5 samples and the
+            # ShiftedBatchSampler drops one batch per epoch, so any batch_size > 1
+            # leaves the dataloader empty and Lightning aborts with
+            # "Trainer.fit stopped: No training batches."
+            cfg.data.batch_size = 2
+            cfg.data.num_workers = 10
```
ktinubu: WARN: [ml-test] `num_workers=10` for a 5-sample fixture.

10 worker processes for 5 examples means each worker imports torch/lightning/the dataset module to handle at most one sample. For tests this is pure overhead. `num_workers=0` (or `2`) is appropriate at this scale — `cfg_eval_global` uses 0.

tests/conftest.py
```diff
@@ -134,3 +147,167 @@ def cfg_eval(cfg_eval_global: DictConfig, tmp_path: Path) -> DictConfig:
...
+            cfg.model.compile = True
+            cfg.model.scheduler = None
+            cfg.logger = None
+            cfg.test = False
+            cfg.callbacks.model_checkpoint = {
+                "_target_": "lightning.pytorch.callbacks.ModelCheckpoint",
+                "dirpath": "${paths.output_dir}/checkpoints",
+                "save_last": True,
+                "every_n_train_steps": 50,
+            }
```
ktinubu: WARN: [code-health] Inline dict literal hardcodes `_target_` and the magic 50.

This hardcodes `lightning.pytorch.callbacks.ModelCheckpoint` (will silently break if upstream renames the path) and duplicates the magic number `50` (matches `max_steps` and `every_n_train_steps`). Prefer mutating the existing `model_checkpoint` config in place:

```python
mc = cfg.callbacks.model_checkpoint
mc.dirpath = "${paths.output_dir}/checkpoints"
mc.save_last = True
mc.every_n_train_steps = MAX_TRAIN_STEPS
mc.save_top_k = -1
```

tests/conftest.py
```diff
@@ -134,3 +147,167 @@ def cfg_eval(cfg_eval_global: DictConfig, tmp_path: Path) -> DictConfig:
...
+
+    generate_dataset_args += [
+        sys.executable,
+        "src/data/vst/generate_vst_dataset.py",
+        str(overfit_dataset_dir / "train.h5"),
+        str(NUM_FIXTURE_SAMPLES),
+    ]
+
+    subprocess.check_call(generate_dataset_args)  # noqa: S603, S607
+    assert (overfit_dataset_dir / "train.h5").exists(), (
```
ktinubu: WARN: [project] `subprocess.check_call` with no timeout.

The Surge XT VST under Xvfb has been known to hang on plugin init. With no timeout the test hangs CI indefinitely instead of failing fast. Add a `timeout=` (e.g. 600s) and let the timeout exception bubble up — that's a much clearer CI failure than a job kill at the workflow timeout.

tests/conftest.py
```diff
@@ -9,6 +12,7 @@
 from omegaconf import DictConfig, open_dict
 
 from src.utils.utils import register_resolvers
+from tests._baseline_worktree import worktree_for_ref  # noqa: F401 — pytest fixture re-export
```
ktinubu, line 19: WARN: [project] `tests/_baseline_worktree.py` is not on this branch.

This file was added to main in #679, but the merge base of `worktree-surge-xt-train-test` with main is `172718a` (0.5.0 release) — older than #679. So `tests/_baseline_worktree.py` exists on main but not on the PR branch's HEAD, and importing it at conftest collection time will `ModuleNotFoundError`.

GitHub Actions `pull_request` jobs check out the merge ref by default, so CI is probably masking this. But anyone checking out this branch directly (or rebasing onto an older main) will see all tests fail to collect. Rebase onto current `main` before merging.

tests/conftest.py
```diff
@@ -134,3 +147,167 @@ def cfg_eval(cfg_eval_global: DictConfig, tmp_path: Path) -> DictConfig:
...
+
+
+@pytest.fixture(scope="function")
+def surge_xt_overfit_datasets(tmp_path: Path) -> Path:
+    """A pytest fixture for the N-sample Surge XT overfitting test fixture.
+
+    :return: A Path object pointing at the directory containing the N-sample Surge XT fixture.
+    """
+    NUM_FIXTURE_SAMPLES = 5
+    # Bootstraps Xvfb + xsettingsd + dbus for VST3 plugin init; resolved relative
```
ktinubu: WARN: [code-health DRY] `NUM_FIXTURE_SAMPLES = 5` is also defined inside `tests/test_train.py:167`. Both are `5`, both refer to the same fixture. Move to a module-level constant in conftest and import it from the test, or pass the count via the fixture itself.

tests/test_train.py
```diff
@@ -1,10 +1,19 @@
...
+import pandas as pd
 import pytest
+import torch
+from click.testing import CliRunner
 from hydra.core.hydra_config import HydraConfig
 from omegaconf import DictConfig, open_dict
+from pedalboard.io import AudioFile
 
+from scripts.compute_audio_metrics import main as compute_audio_metrics_main
+from src.eval import evaluate
```
ktinubu, line 13: WARN: [project] Eager top-level imports of VST-adjacent and audio modules.

The previous commit on this branch (`960e2c4 fix(test): lazy-import VST scripts in test_train_eval_surge_xt`) explicitly moved VST script imports inside the test body to avoid pulling in plugin/X11 dependencies during collection. But `pedalboard.io.AudioFile`, `scripts.compute_audio_metrics.main`, and `click.testing.CliRunner` are still imported at module top — these are only used by `test_train_eval_surge_xt`. Move them inside the test (or behind a lazy-import helper) for consistency with that commit's intent.

tests/test_train.py
```diff
@@ -151,15 +134,155 @@ def test_train_resume(tmp_path: Path, cfg_train: DictConfig) -> None:
...
+
+@pytest.mark.gpu
+@RunIf(min_gpus=1)
+def test_train_surge_xt(cfg_surge_xt: DictConfig) -> None:
+    """Run training of the Surge XT flow-matching model on the 5-sample fixture.
+
+    :param cfg_surge_xt: Surge XT training config.
+    """
+    HydraConfig().set_config(cfg_surge_xt)
+    _, object_dict = train(cfg_surge_xt)
```
ktinubu: WARN: [ml-test MT1/MT11] `test_train_surge_xt` has no assertions.

```python
HydraConfig().set_config(cfg_surge_xt)
train(cfg_surge_xt)
```

This passes whenever `train()` doesn't raise, so a silent regression (e.g. zero-loss bug, no-op trainer, missing checkpoint) won't fail the test. At minimum, assert the checkpoint file exists, or unpack `metric_dict, object_dict = train(...)` and assert one of them is non-empty.

tests/test_train.py
```diff
@@ -151,15 +134,155 @@ def test_train_resume(tmp_path: Path, cfg_train: DictConfig) -> None:
...
+
+
```
ktinubu, line 182: WARN: [ml-test] Test name doesn't match behavior.

`test_overfit_surge_xt` doesn't actually verify overfit — it asserts wiring (file presence, finite tensors, non-silent audio, finite metrics) but never checks that loss decreases or that predictions match targets. Either rename to something like `test_train_eval_surge_xt` (matching its actual purpose as an end-to-end smoke test), or add an overfit assertion (e.g. final train loss < initial train loss / N, or pred audio similarity to target).

tests/test_train.py
```diff
@@ -151,15 +134,155 @@ def test_train_resume(tmp_path: Path, cfg_train: DictConfig) -> None:
...
+        "target-audio-2.pt",
+        "target-audio-3.pt",
+        "target-audio-4.pt",
+        "target-params-0.pt",
+        "target-params-1.pt",
+        "target-params-2.pt",
+        "target-params-3.pt",
+        "target-params-4.pt",
+    ]
+
```
ktinubu: WARN: [ml-test MT15] Hardcoded list of 15 expected file names is brittle.

If `NUM_FIXTURE_SAMPLES` changes from 5, you have to hand-edit 15 strings. Use a generated set instead:

```suggestion
    expected_names = sorted(
        f"{prefix}-{i}.pt"
        for prefix in ("pred", "target-audio", "target-params")
        for i in range(NUM_FIXTURE_SAMPLES)
    )
    assert sorted(p.name for p in predictions_dir.iterdir()) == expected_names
```

This is ML-pipeline checklist MT15 ("loose granularity, counts not exact values") and MT16 ("additive resilience").

tests/test_train.py
```diff
@@ -151,15 +134,155 @@ def test_train_resume(tmp_path: Path, cfg_train: DictConfig) -> None:
         cfg_train.ckpt_path = str(tmp_path / "checkpoints" / "last.ckpt")
         cfg_train.trainer.max_epochs = 2
 
-    metric_dict_2, _ = train(cfg_train)
+    _, _ = train(cfg_train)
 
     files = os.listdir(tmp_path / "checkpoints")
     assert "epoch_001.ckpt" in files
     assert "epoch_002.ckpt" not in files
```
ktinubu, line 144: WARN: [ml-test] `test_train_resume` lost its actual signal in this PR.

The pre-PR test asserted `metric_dict_1["train/loss_epoch"] > metric_dict_2["train/loss_epoch"]` and the same for `val/loss` after resume — that was the test's verification that resume actually trained another epoch (and not just loaded + saved a checkpoint). Now it only asserts file presence, which would also pass if the second `train(cfg_train)` was a no-op. Re-add the loss-decrease assertion (or replace with an equivalent check, e.g. step counter advanced) — file presence alone is too weak.

ktinubu (commented):
Self-review against checklists: tdd-implementation, code-health, ml-test, python-style, synth-setter-project-standards. (Posted as COMMENT — GitHub doesn't allow REQUEST_CHANGES on your own PR; the BLOCK findings still need to clear before merge.)

Summary: 4 BLOCK, 15 WARN

The biggest issues are (1) a `flow_full` vs `ffn_full` divergence between the docstring/doc and the actual override — both experiments exist and the docs claim flow-matching while the code wires up FFN, so a reader can't tell which is correct; (2) a contradicting comment + value pair on `batch_size`; and (3) `test_train_resume` lost its loss-decrease assertion in this PR, which was the test's actual signal that the resume worked. The rest are quality issues — magic-number duplication, eager imports the prior commit explicitly tried to defer, a 134-line test, and a brittle hardcoded list of 15 file names.

Note on `from tests._baseline_worktree import ...` — the file was added to main in #679 but the merge base of this branch is older, so the file isn't on this branch's HEAD. CI may still pass via the merge ref, but a fresh checkout of the branch fails to collect at all. Rebase on main before merging.

tests/test_train.py
```diff
@@ -151,15 +134,155 @@ def test_train_resume(tmp_path: Path, cfg_train: DictConfig) -> None:
...
+    for i in range(NUM_FIXTURE_SAMPLES):
+        pred = torch.load(predictions_dir / f"pred-{i}.pt", weights_only=True)
+        assert torch.isfinite(pred).all(), f"pred-{i}.pt contains NaN/Inf"
+
+    # Render predicted params through the Surge XT VST to per-sample audio directories.
+    # `-t` (`--rerender_target`) re-synthesizes target.wav from the stored target_params instead
+    # of the saved target audio. Also works around an `UnboundLocalError` in
+    # `scripts/predict_vst_audio.py:220` where `target_synth_params` is referenced in the default
+    # path without being defined outside the `rerender_target` branch.
+    audio_dir = tmp_path / "audio"
```
ktinubu, line 241: WARN: [code-health] Comments reference specific line numbers in other source files.

`scripts/predict_vst_audio.py:220`, `src/utils/callbacks.py:332` (line 195), and `scripts/compute_audio_metrics.py:227` (line 258) all hard-code line numbers that drift the moment those files are touched. Reference symbol names (`PredictionWriter._save_predictions`, `compute_rms`, etc.) instead — they survive line-number churn.

tests/test_train.py
```diff
@@ -151,15 +134,155 @@ def test_train_resume(tmp_path: Path, cfg_train: DictConfig) -> None:
...
+
+    args += [
+        sys.executable,
+        "scripts/predict_vst_audio.py",
+        str(predictions_dir),
+        str(audio_dir),
+        "-t",
+    ]
+    subprocess.check_call(args)  # noqa: S603, S607
+
```
ktinubu, line 276: WARN: [project] Same no-timeout issue as `tests/conftest.py:240`.

`scripts/predict_vst_audio.py` re-renders 5 samples through Surge XT under Xvfb. With no `timeout=`, a hung VST will block the whole CI job. Add a generous timeout that fails fast but covers slow CI runners.

docs/reference/testing.md
```diff
@@ -78,6 +78,8 @@ Both defined in [`tests/conftest.py`](../../tests/conftest.py) as package-scoped
 
 Both fixtures clear global Hydra state on teardown via `GlobalHydra.instance().clear()`.
 
+A third fixture group — `cfg_surge_xt_global` / `cfg_surge_xt` / `cfg_surge_xt_eval` — follows the same pattern for Surge XT flow-matching tests (`tests/test_train.py::test_train_surge_xt_*`). It composes `train.yaml` with `experiment=surge/flow_full`, points at the 5-sample fixture under `tests/fixtures/surge_xt/`, and bakes in the 1-step trainer knobs. `cfg_surge_xt_eval` additionally composes `eval.yaml` and copies `data` / `model` from `cfg_surge_xt` so the evaluator matches the trained checkpoint shape (see gotcha #2).
```
ktinubu: BLOCK: [code-health] Doc says `experiment=surge/flow_full` but `tests/conftest.py:165` uses `experiment=surge/ffn_full`.

These are two different experiments (`configs/experiment/surge/flow_full.yaml` vs `configs/experiment/surge/ffn_full.yaml`). The doc also calls these "Surge XT flow-matching tests", which doesn't fit FFN. Align the doc and the fixture — see paired comment on `tests/conftest.py:165`.

tests/conftest.py
```diff
@@ -146,6 +149,159 @@ def cfg_eval(cfg_eval_global: DictConfig, tmp_path: Path) -> DictConfig:
     GlobalHydra.instance().clear()
 
 
+@pytest.fixture(scope="package")
+def cfg_surge_xt_global() -> DictConfig:
+    """A pytest fixture for a one-step Surge XT training config on the N-sample test fixture.
+
+    Composes `train.yaml` with `experiment=surge/flow_full` and bakes in the minimal overrides
```
ktinubu: BLOCK: [code-health] Still says `experiment=surge/flow_full` while line 166 uses `surge/ffn_full`. Same finding as the previous review — not yet addressed. The fixture-level docstring at line 154 ("one-step Surge XT training config") and the eval-fixture doctring at line 277 also refer to flow-matching. Pick one and reconcile all three places + `docs/reference/testing.md:81` + `tests/test_train.py:146` (also says "flow-matching model").

tests/conftest.py
```diff
@@ -146,6 +149,159 @@ def cfg_eval(cfg_eval_global: DictConfig, tmp_path: Path) -> DictConfig:
...
+    needed to train-smoke-test on the dataset generated by surge_xt_smoke_datasets).
```
ktinubu: WARN: [code-health] Stray `)` typo in the docstring — `surge_xt_smoke_datasets)` has a closing paren with no opener.

```suggestion
    needed to train-smoke-test on the dataset generated by `surge_xt_smoke_datasets`.
```

ktinubu (commented):
Re-review after `cca6aac "continue"`. Many prior findings addressed (batch_size, num_workers/pin_memory, magic-50 → TRAINING_STEPS, fixture rename, test rename, train/predict overlap comment) — thanks. New findings + still-open issues below.

Summary: 3 BLOCK, 5 WARN (new) + ~10 still-open WARNs from previous review.

New BLOCKs:
- `flow_full` vs `ffn_full` mismatch is still unfixed in 3 places (conftest docstring, test_train.py docstring, docs/reference/testing.md).
- The new tests lost their `@pytest.mark.gpu` + `@RunIf(min_gpus=1)` decorators in this commit — they're now `@pytest.mark.slow` only. Combined with the project's marker setup, they will not run in any existing CI workflow: `test.yml` filters `-m "not slow"` (skips them) and `test-expensive.yml` filters `-m gpu` (also skips them, no gpu marker). Plus the fixture still hardcodes `cfg.trainer.accelerator = "gpu"` and `cfg.model.compile = True`, so the tests can't run on CPU even if they were collected. Net effect: tests are added but never executed.
- Typo `surge_xt_smoke_datasets)` (stray `)`) in the cfg_surge_xt_global docstring.

New WARNs:
- `check_val_every_n_epoch = 1  # validate every 5 epochs` — comment is stale (was the same problem at 50, fixed value but not comment).
- Docstrings still say "one-step Surge XT training config" but `TRAINING_STEPS = 5`.
- `# TODO(#INSERT ISSUE NUMBER): ...` placeholder needs a real issue.
- batch_size comment formatting: `so any: batch_size >  dataset_size // 2` reads awkwardly with the colon and double space.

Still open from prior review (not re-flagged inline to avoid duplicate threads — see https://github.com/tinaudio/synth-setter/pull/674#pullrequestreview-4183619290):
- `cfg.model.compile = True` in smoke fixture (`tests/conftest.py:194`)
- inline ModelCheckpoint dict literal hardcoding `_target_` (`tests/conftest.py:198-203`)
- `subprocess.check_call` without timeout (conftest.py:241, test_train.py:245)
- `tests/_baseline_worktree` import not on this branch (conftest.py:15)
- Eager top-level VST/audio imports (test_train.py:9-13)
- `test_train_surge_xt` has no assertions (test_train.py:151)
- Hardcoded 15-element prediction filename list (test_train.py:198-214)
- `test_train_resume` lost its loss-decrease assertion (test_train.py:140)
- `VST_HEADLESS_WRAPPER` duplicated between conftest and test_train (test_train.py:233)
- Comments referencing `file:line` in other source files (test_train.py:192, 222-223, 257)

The biggest one to fix this round is the marker regression — adding tests that no CI job runs is worse than not adding them.

tests/conftest.py
```diff
@@ -146,6 +149,159 @@ def cfg_eval(cfg_eval_global: DictConfig, tmp_path: Path) -> DictConfig:
     GlobalHydra.instance().clear()
 
 
+@pytest.fixture(scope="package")
+def cfg_surge_xt_global() -> DictConfig:
+    """A pytest fixture for a one-step Surge XT training config on the N-sample test fixture.
```
ktinubu, line 264: WARN: [code-health] Docstring says "one-step Surge XT training config" (also at line 160: "one-step Surge XT smoke train"), but `TRAINING_STEPS = 5` since this commit. Either rename to "5-step" / "few-step" or just "smoke train" without a step count, and keep the source-of-truth in the constant.

```suggestion
    """A pytest fixture for a short Surge XT smoke-train config on the N-sample test fixture.
```

tests/conftest.py
```diff
@@ -146,6 +149,159 @@ def cfg_eval(cfg_eval_global: DictConfig, tmp_path: Path) -> DictConfig:
...
+            cfg.data.num_workers = 0
+            cfg.data.pin_memory = False
+            cfg.data.ot = False
+            # The 5-sample fixture's stats.npz has zero-std mel bins that poison the batch
+            # with NaN via (mel - mean) / std.
+            cfg.data.use_saved_mean_and_variance = False
+
+            cfg.trainer.devices = 1
+            cfg.trainer.max_steps = TRAINING_STEPS
+            cfg.trainer.check_val_every_n_epoch = 1  # validate every 5 epochs
```
ktinubu: WARN: [code-health] Comment is stale.

The value changed from 50 to 1 in this commit but the comment still says "validate every 5 epochs" — the same wrong-comment issue as the previous review, just with a different value mismatch. With `max_steps=5` and `batch_size=1` over the 5-sample fixture (drop-last → 4 train batches/epoch), training fits in ~2 epochs, so `check_val_every_n_epoch=1` is "validate at end of each epoch" — not 5.

```suggestion
            cfg.trainer.check_val_every_n_epoch = 1  # validate at end of each epoch
```

tests/conftest.py
```diff
@@ -146,6 +149,159 @@ def cfg_eval(cfg_eval_global: DictConfig, tmp_path: Path) -> DictConfig:
...
+
+        TRAINING_STEPS = 5
+        with open_dict(cfg):
+            cfg.paths.root_dir = str(rootutils.find_root(indicator=".project-root"))
+
+            # batch_size=1 is forced: ShiftedBatchSampler (used in the
+            # SurgeDataModule's train_dataloader) drops one batch per epoch,
+            # so any: batch_size >  dataset_size // 2
+            # leaves the dataloader empty and Lightning aborts with:
+            # "Trainer.fit stopped: No training batches."
```
ktinubu: WARN: [code-health] Comment formatting reads awkwardly — the colon at the end of "so any:" doesn't introduce a list, and "batch_size >  dataset_size" has a double space.

```suggestion
            # batch_size=1 is forced: ShiftedBatchSampler (used in the
            # SurgeDataModule's train_dataloader) drops one batch per epoch,
            # so any batch_size > dataset_size // 2 leaves the dataloader
            # empty and Lightning aborts with
            # "Trainer.fit stopped: No training batches."
```

tests/conftest.py
```diff
@@ -146,6 +149,159 @@ def cfg_eval(cfg_eval_global: DictConfig, tmp_path: Path) -> DictConfig:
...
+        "src/data/vst/generate_vst_dataset.py",
+        str(smoke_dataset_dir / "train.h5"),
+        str(NUM_FIXTURE_SAMPLES),
+    ]
+
+    subprocess.check_call(generate_dataset_args)  # noqa: S603, S607
+    assert (smoke_dataset_dir / "train.h5").exists(), (
+        "Dataset generation failed to produce train.h5 fixture"
+    )
+    # TODO(#INSERT ISSUE NUMBER): add validation for the generated audio and mel spectrograms
```
ktinubu: WARN: [project] Placeholder TODO — `#INSERT ISSUE NUMBER` is a literal string, not a reference. Either file the issue and fill it in (e.g. `# TODO(#741): ...`) or drop the TODO until you have a number. Per CLAUDE.md the convention is `TODO:` with a real bug reference.

tests/test_train.py
```diff
@@ -131,3 +139,149 @@ def test_train_resume(tmp_path: Path, cfg_train: DictConfig) -> None:
     files = os.listdir(tmp_path / "checkpoints")
     assert "epoch_001.ckpt" in files
     assert "epoch_002.ckpt" not in files
+
+
+@pytest.mark.slow
```
ktinubu, line 157: BLOCK: [project / ml-test] Marker regression — `@pytest.mark.gpu` and `@RunIf(min_gpus=1)` were removed in this commit, leaving only `@pytest.mark.slow`.

With the project's marker config:
- Default `pytest_add_cli_args` (`pyproject.toml:119`) excludes `slow`, `gpu`, `requires_vst`.
- `make test` and `.github/workflows/test.yml` run `-m "not slow"` → these tests skip.
- `.github/workflows/test-expensive.yml` runs `-m gpu` → also skips (no gpu marker).
- No CI job runs `make test-full` or `-m slow`.

Net result: these tests are added but never execute anywhere in CI. Plus the fixture still forces `cfg.trainer.accelerator = "gpu"` and `cfg.model.compile = True`, so they can't run on CPU even if collected.

Fix: re-add `@pytest.mark.gpu` (so `test-expensive.yml` runs them) and consider `@pytest.mark.requires_vst` since the test shells out to the Surge XT VST. Keep `@RunIf(min_gpus=1)` for explicit skip-when-no-GPU rather than relying on the marker filter alone.

```suggestion
@pytest.mark.gpu
@pytest.mark.requires_vst
@pytest.mark.slow
@RunIf(min_gpus=1)
def test_train_surge_xt(cfg_surge_xt: DictConfig) -> None:
```

tests/test_train.py
```diff
@@ -131,3 +139,149 @@ def test_train_resume(tmp_path: Path, cfg_train: DictConfig) -> None:
...
+def test_train_surge_xt(cfg_surge_xt: DictConfig) -> None:
+    """Run training of the Surge XT flow-matching model on the smoke test fixture.
+
+    :param cfg_surge_xt: Surge XT training config.
+    """
+    HydraConfig().set_config(cfg_surge_xt)
+    train(cfg_surge_xt)
+
+
+@pytest.mark.slow
```
ktinubu, line 183: BLOCK: [project / ml-test] Same marker regression as `test_train_surge_xt` above — only `@pytest.mark.slow`, so it never runs in any CI workflow. Apply the same set of decorators.

```suggestion
@pytest.mark.gpu
@pytest.mark.requires_vst
@pytest.mark.slow
@RunIf(min_gpus=1)
def test_train_eval_surge_xt(
```

tests/test_train.py
```diff
@@ -131,3 +139,149 @@ def test_train_resume(tmp_path: Path, cfg_train: DictConfig) -> None:
     files = os.listdir(tmp_path / "checkpoints")
     assert "epoch_001.ckpt" in files
     assert "epoch_002.ckpt" not in files
+
+
+@pytest.mark.slow
+def test_train_surge_xt(cfg_surge_xt: DictConfig) -> None:
+    """Run training of the Surge XT flow-matching model on the smoke test fixture.
```
ktinubu: BLOCK: [code-health] Docstring says "flow-matching model" but `cfg_surge_xt` composes `experiment=surge/ffn_full` (a feedforward, non-flow experiment). Same family as the conftest.py:156 finding — paired fix.

.github/workflows/test-mps.yml
```diff
@@ -0,0 +1,78 @@
...
+    paths:
+      - "src/**"
+      - "configs/**"
+      - ".github/workflows/test-mps.yml"
+  pull_request:
+    branches: [main, "release/*", "dev"]
+    paths:
+      - "src/**"
+      - "configs/**"
+      - ".github/workflows/test-mps.yml"
```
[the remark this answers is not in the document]
ktinubu: will add add tests/test_train.py and                   
  tests/conftest.py)

tests/conftest.py
```diff
@@ -146,6 +232,215 @@ def cfg_eval(cfg_eval_global: DictConfig, tmp_path: Path) -> DictConfig:
...
+    """
+    cfg = cfg_surge_xt_global.copy()
+    with open_dict(cfg):
+        cfg.paths.output_dir = str(tmp_path)
+        cfg.paths.log_dir = str(tmp_path)
+        cfg.data.batch_size = 1
+        cfg.data.dataset_root = str(surge_xt_smoke_datasets)
+        cfg.data.predict_file = str(surge_xt_smoke_datasets / "test.h5")
+        cfg.ckpt_path = str(tmp_path / "checkpoints" / "last.ckpt")
+        cfg.mode = "predict"
```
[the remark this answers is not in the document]
ktinubu: This is to simplify the logic for the smoke tests. the global surge config compses train.yaml, but we add also compose the surge_eval callbacks

docs/reference/testing.md
```diff
@@ -77,6 +80,8 @@ Both defined in [`tests/conftest.py`](../../tests/conftest.py) as package-scoped
 
 Both fixtures clear global Hydra state on teardown via `GlobalHydra.instance().clear()`.
 
+A third fixture group — `cfg_surge_xt_global` / `cfg_surge_xt` / `cfg_surge_xt_eval` — follows the same pattern for Surge XT FFN tests (`tests/test_train.py::test_train_surge_xt_*`). It composes `train.yaml` with `experiment=surge/ffn_full`, points at the 5-sample dataset generated on demand by the `surge_xt_smoke_datasets` fixture, and bakes in the smoke-test trainer knobs. `cfg_surge_xt_eval` additionally composes `eval.yaml` and copies `data` / `model` from `cfg_surge_xt` so the evaluator matches the trained checkpoint shape (see gotcha #2).
```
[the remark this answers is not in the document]
ktinubu: done

Issue #532, opened 2026-04-08
chore(evaluation): harden surge_xt_interactive — smoke tests + error handling


ktinubu:
Follow-up from [#531](https://github.com/tinaudio/synth-setter/pull/531).

**A. Add smoke tests** for `scripts/surge_xt_interactive.py`:
- Import doesn't trigger side effects (validates `__main__` guard)
- Click CLI `--help` works without a plugin present
- `play_audio` is importable with correct signature

See `tests/scripts/test_r2_shard_report.py` for the existing pattern.

**B. Add exception handling in the audio daemon thread** ([review comment](https://github.com/tinaudio/synth-setter/pull/531#discussion_r3053462499)):
- The `while True` loop in `play_audio()` has no try/except — if `plugin()` or `stream.write()` raises, the error is silently lost because it's a daemon thread
- Wrap the loop body with logging so failures are visible

Refs #530

ktinubu:
Closing as complete. Verified by multi-agent audit pass.

## What was asked

Follow-up from PR #531: (A) add smoke tests for `scripts/surge_xt_interactive.py` covering import side-effects, Click `--help`, and `play_audio` signature; (B) add exception handling in the audio daemon thread's `while True` loop so failures aren't silently lost.

## What landed

Tests landed alongside PR #723 (`feat(evaluation): apply model predictions to live Surge XT...`, merge commit 6977add) which seeded `tests/scripts/test_surge_xt_interactive.py`, plus a series of follow-ups hardening the test surface (PRs #833, #844, #855, #857, #859).

- `tests/scripts/test_surge_xt_interactive.py` covers: lazy import via `importlib` fixture (validates no import-time side effects), decode-prediction helpers, URI parsing, `play_audio_recorded` rendering, `play_audio` MIDI-queue drain semantics, MIDI listener `open_input` failure logging, and stop-event exit paths — 25+ test functions across multiple classes.
- `scripts/surge_xt_interactive.py` daemon threads now log on failure: MIDI listener wraps its loop with `try/except Exception` + `logger.exception("MIDI listener thread aborted on port %r", port_name)` (line ~501-511); keyboard loop wraps per-keystroke handling with `logger.exception("keyboard action %r failed", ch)` (line ~667).
- PR #833 (`chore(scripts): avoid import-time logging side effects`) directly addresses the import-side-effect smoke test requirement.

## Verification

[...]

Issue #530, opened 2026-04-08
Task 6.1: interactive Surge XT preview script


ktinubu:
Add `scripts/surge_xt_interactive.py` — opens Surge XT GUI via pedalboard with real-time audio streaming via AudioStream + sounddevice.

Part of Phase 6: Synth Preview Tools (#529).

PR #702, opened 2026-04-29
feat(evaluation): apply model predictions to live Surge XT and record patches to dataset

### LINKED ISSUE #701: feat(evaluation): apply model predictions to live Surge XT and record patches to dataset
Extend `scripts/surge_xt_interactive.py` from a "spin up Surge XT and stream silence" demo into a feedback loop that lets a researcher (1) hear a model prediction (or an existing dataset row) on the live VST, (2) tweak it via the editor and snapshot patches, and (3) write those patches back into an HDF5 dataset for downstream training.

## Capabilities

1. **Apply predictions or dataset rows to the live VST.**
   - `--pred PATH:BATCH_IDX` — load a `pred-*.pt` tensor row, rescale `[-1, 1] → [0, 1]`, clip, and decode via `param_specs[<spec>]` before opening the editor.
   - `--dataset-ref PATH:DATASET_IDX` — load an encoded row from an h5 `param_array` and decode it (no rescaling — already in `[0, 1]`).
2. **Capture user edits as patches.** A new `keyboard_loop` runs alongside `play_audio` (threaded via `ThreadPoolExecutor`); press `p` to snapshot every synth-spec param, `q` to quit. A shared `threading.Event` coordinates editor-close and audio shutdown.
3. **Render captured patches into a dataset.** After the editor closes, call `make_dataset(..., fixed_synth_params_list=synth_patches)` so each captured patch becomes a new row in `--output-dataset-path`.

## Pipeline support changes

- `src/data/vst/generate_vst_dataset.py`: `generate_sample` and `make_dataset` accept optional `fixed_synth_params_list` / `fixed_note_params_list` so captured patches can be rendered deterministically.
- `src/data/vst/core.py`: replace the KeyboardInterrupt-based `_call_with_interrupt` with a `threading.Event` + `show_editor(stop_event)` pattern that closes cleanly. `render_params` reloads the plugin per call (takes `plugin_path: str`).

## Tests

- New unit tests for `decode_prediction_row`, `PredictionRefType`/`DatasetRefType` parsers, `load_prediction_synth_params`, `load_dataset_synth_params`, `render_to_wav` (CPU-only, stubbed plugin).
- New e2e h5 test for `make_dataset` with `fixed_synth_params_list` (slow + requires_vst).

## Related

- Refs #532 — the smoke-tests piece is delivered here as `tests/scripts/test_surge_xt_interactive.py`; the audio-thread error-surfacing piece is addressed by `audio_future.result(timeout=2)` in `main`'s `finally`.
- Refs #530 — the original interactive preview script Task this builds on.

Part of Phase 6: Synth Preview Tools (#529).

ktinubu:
## Summary

- Apply a model prediction (`--pred PATH:BATCH_IDX`) or a dataset row (`--dataset-ref PATH:DATASET_IDX`) to a live Surge XT before opening its editor.
- Capture user-edited patches from the editor (press `p` to snapshot synth params, `q` to quit) via a threaded keyboard loop alongside real-time audio playback.
- After the editor closes, render every captured patch through the plugin and append the samples to an HDF5 dataset via `make_dataset(..., fixed_synth_params_list=...)`.

## What changed

- **`scripts/surge_xt_interactive.py`** (full rewrite from "stream silence" demo):
  - New CLI flags: `--pred`, `--dataset-ref`, `--preset-path`, `--param-spec-name`, `--output-dataset-path`.
  - Click param types `PredictionRefType` / `DatasetRefType` parse `PATH:IDX` strings.
  - `decode_prediction_row` (pure): rescale `[-1, 1] → [0, 1]`, clip, and decode via `param_specs[<spec>]`.
  - `load_prediction_synth_params`, `load_dataset_synth_params`: thin disk-loading orchestrators around the decode helpers.
  - `keyboard_loop`: `p` records a patch, `q` sets the shared `stop_event`.
  - `play_audio` is now `stop_event`-driven and resamples to `PLAYBACK_SAMPLE_RATE` when needed.
  - `render_to_wav` for headless rendering (no `AudioStream` device required).
  - `main` runs `play_audio` + `keyboard_loop` in a `ThreadPoolExecutor` and surfaces audio-thread exceptions via `audio_future.result(timeout=2)` in `finally` (addresses the daemon-thread silent-failure concern in #532).

- **`src/data/vst/core.py`**:
  - Replaced KeyboardInterrupt-based `_call_with_interrupt` with a `threading.Event` + `show_editor(stop_event)` pattern that closes deterministically.
  - `render_params` now takes `plugin_path: str` and reloads the plugin per call so each render starts from a clean state.

- **`src/data/vst/generate_vst_dataset.py`**:
  - `generate_sample` and `make_dataset` accept optional `fixed_synth_params_list` / `fixed_note_params_list` so captured patches can be rendered deterministically into h5.

- **Tests (new)**:
  - `tests/scripts/test_surge_xt_interactive.py` (341 lines, CPU-only): unit tests for `decode_prediction_row` (rescale + clip + per-column behavior), `PredictionRefType`/`DatasetRefType` parsers (incl. malformed input rejection), `load_prediction_synth_params`, `load_dataset_synth_params` (round-trip + correct row selection), and `render_to_wav` against a `_ConstantPlugin` stub.
  - `tests/data/vst/test_generate_vst_dataset.py` (slow + requires_vst): e2e h5 verification with both a fixed-params variant (per-row distinct patches to exercise indexing) and a random-sampling variant.

## Notes for reviewers

- **Cosmetic-only oddity, not a bug:** `src/data/vst/core.py:49` declares the parameter as `plugin＿path` with U+FF3F (FULLWIDTH LOW LINE) instead of the regular ASCII underscore. Per [PEP 3131](https://peps.python.org/pep-3131/) Python NFKC-normalizes identifiers at parse time, so at runtime it's identical to `plugin_path` — keyword calls and positional calls both work. Confirmed via `inspect.signature()` which shows `plugin_path` (`0x5f`). Worth replacing with the ASCII form for readability before merge.
- **`--dataset-ref` wins over `--pred`** when both are passed (priority `if/elif` in `main`). Consider Click's `--mutually-exclusive` if that's not intended.

## Linked issues

- Closes #701
- Refs #532 (smoke-tests piece delivered as `tests/scripts/test_surge_xt_interactive.py`; audio-thread error surfacing addressed via `audio_future.result(timeout=2)`)
- Refs #530

## Test plan

- [ ] `make test` (excludes slow / VST-gated)
- [ ] `pytest tests/scripts/test_surge_xt_interactive.py -vv` (CPU-only)
- [ ] `pytest tests/data/vst/test_generate_vst_dataset.py -vv` on a host with `plugins/Surge XT.vst3`
- [ ] Manual: `python scripts/surge_xt_interactive.py --pred outputs/pred-0.pt:0 --output-dataset-path /tmp/recorded.h5` — press `p` twice, `q` to quit, confirm two new rows are appended to `/tmp/recorded.h5`

### DIFF
--- a/scripts/surge_xt_interactive.py
+++ b/scripts/surge_xt_interactive.py
@@ -7,35 +7,48 @@ from concurrent.futures import ThreadPoolExecutor
 from dataclasses import dataclass
 from pathlib import Path
 
-import click
-import h5py
-import hdf5plugin  # noqa: F401   side-effect import: registers HDF5_PLUGIN_PATH so h5py can load Blosc2 filters in fixtures
-import numpy as np
 import rootutils
-import torch
-from pedalboard import VST3Plugin
-from pedalboard.io import AudioFile, AudioStream, StreamResampler
-
-from src.data.vst import load_plugin, load_preset, param_specs
-from src.data.vst.core import set_params
-from src.data.vst.generate_vst_dataset import make_dataset
 
 rootutils.setup_root(__file__, indicator=".project-root", pythonpath=True)
+
+import click  # noqa: E402
+import h5py  # noqa: E402
+import hdf5plugin  # noqa: F401, E402  side-effect: registers HDF5_PLUGIN_PATH for Blosc2 filters
+import numpy as np  # noqa: E402
+import torch  # noqa: E402
+from pedalboard import VST3Plugin  # noqa: E402
+from pedalboard.io import AudioFile, AudioStream, StreamResampler  # noqa: E402
+
+from src.data.vst import load_plugin, load_preset, param_specs  # noqa: E402
+from src.data.vst.core import set_params  # noqa: E402
+from src.data.vst.generate_vst_dataset import make_dataset  # noqa: E402
 
 logger = logging.getLogger(__name__)
 
 CHANNELS = 2
 SAMPLE_RATE = 44100
 BUFFER_SIZE = 512
-MAKE_DATASET_SCRIPT_PATH = Path("src/data/vst/generate_vst_dataset.py").resolve()
 MAKE_DATASET_VELOCITY = 100
 MAKE_DATASET_SIGNAL_DURATION_SECONDS = 4.0
 MAKE_DATASET_MIN_LOUDNESS = -50.0
 MAKE_DATASET_SAMPLE_BATCH_SIZE = 32
 
-PLAYBACK_SAMPLE_RATE = (
-    44100  # Sample rate to use for real-time playback. Must be supported by the output device.
-)
+# Real-time playback rate. Set this to whatever the default output device supports
+# if it differs from ``SAMPLE_RATE`` — ``play_audio`` will resample on the fly.
+PLAYBACK_SAMPLE_RATE = SAMPLE_RATE
+
+# Maximum time to wait for the audio thread to drain after ``stop_event`` is set.
+AUDIO_THREAD_DRAIN_TIMEOUT_SECONDS = 2
+
+# Plugin-flush parameters used by the post-load / pre-render flush pattern; see
+# ``_flush_plugin``. Mirror of the values used in ``src.data.vst.core.render_params``.
+_PLUGIN_FLUSH_DURATION_SECONDS = 32.0
+_PLUGIN_FLUSH_BUFFER_SIZE = 2048
+
+# Return signals for ``keyboard_loop`` actions — distinguishes "user requested
+# quit" from "action completed, keep listening".
+_KEEP_LOOPING = True
+_STOP_LOOPING = False
 
 
 @dataclass(frozen=True)
@@ -181,7 +194,10 @@ def load_dataset_synth_params(
     spec = param_specs[param_spec_name]
     with h5py.File(ref.path, "r") as f:
         param_array = f["param_array"]
-        assert isinstance(param_array, h5py.Dataset), "expected h5py.Dataset for 'param_array'"
+        if not isinstance(param_array, h5py.Dataset):
+            raise TypeError(
+                f"expected h5py.Dataset for 'param_array', got {type(param_array).__name__}"
+            )
         row = np.asarray(param_array[ref.batch_idx], dtype=np.float32)
 
     synth_params, _ = spec.decode(row)
@@ -211,6 +227,23 @@ def load_prediction_synth_params(
     return decode_prediction_row(pred_tensor, ref.batch_idx, param_spec_name)
 
 
+def _flush_plugin(plugin: VST3Plugin) -> None:
+    """Process an empty buffer and reset the plugin to flush internal state.
+
+    Mirrors the post-load / pre-render flush pattern in
+    ``src.data.vst.core.render_params``.
+    """
+    plugin.process(
+        [],
+        _PLUGIN_FLUSH_DURATION_SECONDS,
+        SAMPLE_RATE,
+        CHANNELS,
+        _PLUGIN_FLUSH_BUFFER_SIZE,
+        True,
+    )
+    plugin.reset()
+
+
 def play_audio(plugin: VST3Plugin, stop_event: threading.Event) -> None:
     """Stream silence through Surge XT and write synthesized audio to the output device.
 
@@ -219,19 +252,19 @@ def play_audio(plugin: VST3Plugin, stop_event: threading.Event) -> None:
     it differs from ``SAMPLE_RATE`` so the output device gets a rate it supports.
     """
     silence = np.zeros((CHANNELS, BUFFER_SIZE), dtype=np.float32)
-    stream_resampler = None
+    needs_resample = SAMPLE_RATE != PLAYBACK_SAMPLE_RATE
+    stream_resampler = (
+        StreamResampler(SAMPLE_RATE, PLAYBACK_SAMPLE_RATE, CHANNELS) if needs_resample else None
+    )
     with AudioStream(
         output_device_name=AudioStream.default_output_device_name,
         sample_rate=SAMPLE_RATE,
         buffer_size=BUFFER_SIZE,
     ) as stream:
         while not stop_event.is_set():
-            if SAMPLE_RATE != PLAYBACK_SAMPLE_RATE and stream_resampler is None:
-                stream_resampler = StreamResampler(SAMPLE_RATE, PLAYBACK_SAMPLE_RATE, CHANNELS)
             synth_output = plugin(silence, SAMPLE_RATE, reset=False)
-            if stream_resampler:
-                resampled = stream_resampler.process(synth_output)
-                stream.write(resampled, PLAYBACK_SAMPLE_RATE)
+            if stream_resampler is not None:
+                stream.write(stream_resampler.process(synth_output), PLAYBACK_SAMPLE_RATE)
             else:
                 stream.write(synth_output, SAMPLE_RATE)
 
@@ -253,31 +286,35 @@ def keyboard_loop(
     spec = param_specs[param_spec_name]
     synth_patches: list[dict[str, float]] = []
 
-    def quit_action():
+    def quit_action() -> bool:
         stop_event.set()
         logger.info("Quitting...")
-        return False  # stop loop
+        return _STOP_LOOPING
 
-    def record_patch():
-        patch: dict[str, float] = dict()
+    def record_patch() -> bool:
+        patch: dict[str, float] = {}
         logger.info("Recording patch...")
         for param_name in spec.synth_param_names:
-            assert param_name in plugin.parameters, (  # pyright: ignore[reportAttributeAccessIssue]
-                f"parameter {param_name!r} not found in plugin parameters"
-            )
+            if param_name not in plugin.parameters:  # pyright: ignore[reportAttributeAccessIssue]
+                raise KeyError(f"parameter {param_name!r} not found in plugin parameters")
             patch[param_name] = plugin.parameters[param_name].raw_value  # pyright: ignore[reportAttributeAccessIssue]
         synth_patches.append(patch)
         logger.info("patch recorded: %s", synth_patches[-1])
-        return True  # keep going
+        return _KEEP_LOOPING
 
     actions = {
         "q": quit_action,
         "p": record_patch,
     }
+    # NOTE: ``click.getchar()`` blocks until a key is pressed, so this loop only
+    # checks ``stop_event`` between keystrokes. If the editor closes from another
+    # thread, the loop won't exit until the user presses any key. Acceptable for
+    # the typical editor-close-then-press-q flow; a sentinel-key approach would
+    # be needed for fully event-driven shutdown.
     while not stop_event.is_set():
         ch = click.getchar()
         action = actions.get(ch)
-        if action and not action():
+        if action is not None and action() == _STOP_LOOPING:
             return synth_patches
     return synth_patches
 
@@ -354,16 +391,22 @@ def main(
        the resulting samples to ``--output-dataset-path`` via ``make_dataset``.
     """
     logging.basicConfig(level=logging.INFO, format="%(asctime)s %(levelname)s %(message)s")
+
+    if dataset_ref is not None and pred is not None:
+        raise click.UsageError(
+            "--pred and --dataset-ref are mutually exclusive; pass at most one."
+        )
+
     plugin = load_plugin(plugin_path)
-    assert isinstance(plugin, VST3Plugin), f"expected VST3Plugin, got {type(plugin).__name__}"
+    if not isinstance(plugin, VST3Plugin):
+        raise TypeError(f"expected VST3Plugin, got {type(plugin).__name__}")
 
     load_preset(plugin, preset_path)
 
-    # flush plugin to ensure that full parameter dict is populated.
-    plugin.process([], 32.0, SAMPLE_RATE, CHANNELS, 2048, True)  # flush
-    plugin.reset()
-    plugin.process([], 32.0, SAMPLE_RATE, CHANNELS, 2048, True)  # flush
-    plugin.reset()
+    # Two flushes ensure the plugin's full parameter dict is populated and any
+    # transient state from preset load is cleared before applying user params.
+    _flush_plugin(plugin)
+    _flush_plugin(plugin)
 
     if dataset_ref is not None:
         synth_params = load_dataset_synth_params(dataset_ref, param_spec_name)
@@ -372,7 +415,6 @@ def main(
         synth_params = load_prediction_synth_params(pred, param_spec_name)
         set_params(plugin, synth_params)
 
-    synth_patches = None
     stop_event = threading.Event()
     with ThreadPoolExecutor() as pool:
         audio_future = pool.submit(play_audio, plugin, stop_event)
@@ -382,20 +424,43 @@ def main(
             plugin.show_editor(stop_event)
         finally:
             stop_event.set()
-            # surface any exception that happened in the audio thread
-            audio_future.result(timeout=2)
-        logger.info("Editor closed, waiting for audio thread to finish...")
+            # Surface any exception from the audio thread. Catch TimeoutError so
+            # a slow stream-close on shutdown doesn't crash the script — log it
+            # and continue draining; cancel-style errors will resurface elsewhere.
+            try:
+                audio_future.result(timeout=AUDIO_THREAD_DRAIN_TIMEOUT_SECONDS)
+            except TimeoutError:
+                logger.warning(
+                    "audio thread did not exit within %ss; continuing shutdown",
+                    AUDIO_THREAD_DRAIN_TIMEOUT_SECONDS,
+                )
+        logger.info("Editor closed, waiting for keyboard thread to finish...")
         synth_patches = keyboard_future.result()
         logger.info("Recorded %d patches: %s", len(synth_patches), synth_patches)
 
-    assert synth_patches is not None, "synth_patches should be set by keyboard_loop"
     if output_dataset_path is None:
         logger.info("No output dataset path provided, skipping dataset creation.")
         return
+    if not synth_patches:
+        logger.info("No patches recorded, skipping dataset creation.")
+        return
+    # ``make_dataset`` treats ``num_samples`` as the *total* dataset size, not
+    # a delta. Read any existing total from the h5 so resuming doesn't truncate
+    # earlier rows; for a fresh file ``existing_total`` is 0.
     with h5py.File(output_dataset_path, "a") as f:
+        if "audio" in f:
+            existing_audio = f["audio"]
+            if not isinstance(existing_audio, h5py.Dataset):
+                raise TypeError(
+                    f"expected h5py.Dataset for 'audio', got {type(existing_audio).__name__}"
+                )
+            existing_total = existing_audio.shape[0]
+        else:
+            existing_total = 0
+        total_samples = existing_total + len(synth_patches)
         make_dataset(
             hdf5_file=f,
-            num_samples=len(synth_patches),
+            num_samples=total_samples,
             plugin_path=plugin_path,
             preset_path=preset_path,
             sample_rate=SAMPLE_RATE,
--- a/scripts/surge_xt_interactive.py
+++ b/scripts/surge_xt_interactive.py
@@ -170,7 +170,7 @@ def decode_prediction_row(
         raise IndexError(f"batch_idx {batch_idx} out of range (batch size {pred_tensor.shape[0]})")
 
     spec = param_specs[param_spec_name]
-    row = pred_tensor[batch_idx].float().numpy()
+    row = pred_tensor[batch_idx].detach().cpu().float().numpy()
     row_scaled = np.clip((row + 1) / 2, 0, 1)
     synth_params, _ = spec.decode(row_scaled)
     return synth_params
@@ -363,10 +363,11 @@ def keyboard_loop(
     type=click.Path(dir_okay=False, path_type=Path),
     default=None,
     help=(
-        "HDF5 file to append recorded patches to. After the editor is closed, patches "
-        "captured via the keyboard loop (press 'p' to record, 'q' to quit) are rendered "
-        "through the plugin and written to this dataset via "
-        "``src.data.vst.generate_vst_dataset.make_dataset``."
+        "HDF5 file to be created with the recorded patches. Must not already exist — "
+        "``make_dataset`` writes fixed-size HDF5 datasets without ``maxshape`` and cannot "
+        "append to existing files. After the editor is closed, patches captured via the "
+        "keyboard loop (press 'p' to record, 'q' to quit) are rendered through the plugin "
+        "and written to this file via ``src.data.vst.generate_vst_dataset.make_dataset``."
     ),
 )
 def main(
@@ -395,6 +396,17 @@ def main(
     if dataset_ref is not None and pred is not None:
         raise click.UsageError(
             "--pred and --dataset-ref are mutually exclusive; pass at most one."
+        )
+
+    # Fail fast — ``make_dataset`` writes fixed-size HDF5 datasets without
+    # ``maxshape`` and cannot append, so a pre-existing path would either
+    # silently overwrite (when re-creating datasets) or fail mid-render after
+    # patches have been captured. Better to reject up front.
+    if output_dataset_path is not None and output_dataset_path.exists():
+        raise click.UsageError(
+            f"--output-dataset-path {output_dataset_path} already exists; "
+            f"this script writes a new file (fixed-size HDF5 datasets cannot be "
+            f"appended to). Choose a path that does not exist yet."
         )
 
     plugin = load_plugin(plugin_path)
@@ -444,23 +456,10 @@ def main(
     if not synth_patches:
         logger.info("No patches recorded, skipping dataset creation.")
         return
-    # ``make_dataset`` treats ``num_samples`` as the *total* dataset size, not
-    # a delta. Read any existing total from the h5 so resuming doesn't truncate
-    # earlier rows; for a fresh file ``existing_total`` is 0.
-    with h5py.File(output_dataset_path, "a") as f:
-        if "audio" in f:
-            existing_audio = f["audio"]
-            if not isinstance(existing_audio, h5py.Dataset):
-                raise TypeError(
-                    f"expected h5py.Dataset for 'audio', got {type(existing_audio).__name__}"
-                )
-            existing_total = existing_audio.shape[0]
-        else:
-            existing_total = 0
-        total_samples = existing_total + len(synth_patches)
+    with h5py.File(output_dataset_path, "w") as f:
         make_dataset(
             hdf5_file=f,
-            num_samples=total_samples,
+            num_samples=len(synth_patches),
             plugin_path=plugin_path,
             preset_path=preset_path,
             sample_rate=SAMPLE_RATE,
--- a/scripts/surge_xt_interactive.py
+++ b/scripts/surge_xt_interactive.py
@@ -19,7 +19,7 @@ from pedalboard import VST3Plugin  # noqa: E402
 from pedalboard.io import AudioFile, AudioStream, StreamResampler  # noqa: E402
 
 from src.data.vst import load_plugin, load_preset, param_specs  # noqa: E402
-from src.data.vst.core import set_params  # noqa: E402
+from src.data.vst.core import make_midi_events, set_params  # noqa: E402
 from src.data.vst.generate_vst_dataset import make_dataset  # noqa: E402
 
 logger = logging.getLogger(__name__)
@@ -34,15 +34,24 @@ MAKE_DATASET_SAMPLE_BATCH_SIZE = 32
 
 # Real-time playback rate. Set this to whatever the default output device supports
 # if it differs from ``SAMPLE_RATE`` — ``play_audio`` will resample on the fly.
+# Resampling adds overhead, so it's optional and disabled by default.
 PLAYBACK_SAMPLE_RATE = SAMPLE_RATE
 
 # Maximum time to wait for the audio thread to drain after ``stop_event`` is set.
 AUDIO_THREAD_DRAIN_TIMEOUT_SECONDS = 2
 
-# Cap on how many seconds of the live session ``play_audio`` will tee into a
-# WAV file when ``--session-recording-path`` is set. Bounded so a long editor
-# session doesn't write an unbounded recording.
-SESSION_RECORDING_MAX_SECONDS = 20.0
+# Deterministic recording clip when ``--session-recording-path`` is set.
+# The point is reproducibility: same plugin + preset + params -> same WAV.
+# A held middle-C note is rendered from ``NOTE_START`` to ``NOTE_END``,
+# inside a fixed-length window so any release tail is captured.
+SESSION_RECORDING_DURATION_SECONDS = 10.0
+SESSION_RECORDING_MIDI_NOTE = 60  # middle C (C4)
+SESSION_RECORDING_VELOCITY = 100
+SESSION_RECORDING_NOTE_START_SECONDS = 2.0
+SESSION_RECORDING_NOTE_END_SECONDS = 4.0
+# Buffer size used for the offline ``plugin.process(...)`` render. Mirrors
+# the value used in ``src.data.vst.core.render_params``.
+_SESSION_RECORDING_BUFFER_SIZE = 2048
 
 # Plugin-flush parameters used by the post-load / pre-render flush pattern; see
 # ``_flush_plugin``. Mirror of the values used in ``src.data.vst.core.render_params``.
@@ -258,38 +267,50 @@ def play_audio(plugin: VST3Plugin, stop_event: threading.Event) -> None:
                 stream.write(synth_output, SAMPLE_RATE)
 
 
-def play_audio_recorded(
-    plugin: VST3Plugin,
-    stop_event: threading.Event,
-    session_recording_path: Path,
-    n_seconds: float = SESSION_RECORDING_MAX_SECONDS,
-) -> None:
-    """Render plugin output for ``n_seconds`` to ``session_recording_path`` (no live stream).
+def play_audio_recorded(plugin: VST3Plugin, session_recording_path: Path) -> None:
+    """Render a deterministic clip through ``plugin`` to ``session_recording_path``.
 
-    Headless alternative to :func:`play_audio` — opens a WAV via
+    The clip is a fixed ``SESSION_RECORDING_DURATION_SECONDS`` window with a
+    held middle-C note from ``SESSION_RECORDING_NOTE_START_SECONDS`` to
+    ``SESSION_RECORDING_NOTE_END_SECONDS``; the surrounding silence captures
+    any release tail. The output depends only on the loaded plugin state
+    (preset + ``--pred`` / ``--dataset-ref`` params), so the same inputs
+    always produce the same WAV.
+
+    Headless alternative to :func:`play_audio` — uses
     ``pedalboard.io.AudioFile`` instead of an ``AudioStream``, so it works in
-    environments without an audio output device. Loops until either
-    ``n_seconds`` of audio has been written or ``stop_event`` is set,
-    whichever comes first. The final chunk is trimmed so the file lands
-    exactly on ``n_seconds * SAMPLE_RATE`` frames.
+    environments without an audio output device.
     """
-    silence = np.zeros((CHANNELS, BUFFER_SIZE), dtype=np.float32)
-    target = int(n_seconds * SAMPLE_RATE)
-    logger.info("Recording up to %.1fs of plugin output to %s", n_seconds, session_recording_path)
-    written = 0
+    midi_events = make_midi_events(
+        SESSION_RECORDING_MIDI_NOTE,
+        SESSION_RECORDING_VELOCITY,
+        SESSION_RECORDING_NOTE_START_SECONDS,
+        SESSION_RECORDING_NOTE_END_SECONDS,
+    )
+    logger.info(
+        "Rendering %.1fs deterministic clip (note %d, %.1f-%.1fs) to %s",
+        SESSION_RECORDING_DURATION_SECONDS,
+        SESSION_RECORDING_MIDI_NOTE,
+        SESSION_RECORDING_NOTE_START_SECONDS,
+        SESSION_RECORDING_NOTE_END_SECONDS,
+        session_recording_path,
+    )
+    output = plugin.process(
+        list(midi_events),
+        SESSION_RECORDING_DURATION_SECONDS,
+        SAMPLE_RATE,
+        CHANNELS,
+        _SESSION_RECORDING_BUFFER_SIZE,
+        True,
+    )
+    expected_frames = int(SESSION_RECORDING_DURATION_SECONDS * SAMPLE_RATE)
+    assert output.shape == (CHANNELS, expected_frames), (
+        f"expected output shape ({CHANNELS}, {expected_frames}), got {output.shape}"
+    )
+    # AudioFile.write expects (frames, channels); plugin output is (channels, frames).
     with AudioFile(str(session_recording_path), "w", SAMPLE_RATE, CHANNELS) as f:
-        while written < target and not stop_event.is_set():
-            chunk = plugin(silence, SAMPLE_RATE, reset=False)
-            assert chunk.shape == (CHANNELS, BUFFER_SIZE), (
-                f"expected chunk shape ({CHANNELS}, {BUFFER_SIZE}), got {chunk.shape}"
-            )
-            remaining = target - written
-            if chunk.shape[-1] > remaining:
-                chunk = chunk[..., :remaining]
-            # AudioFile.write expects (frames, channels); plugin output is (channels, frames).
-            f.write(chunk.T)
-            written += chunk.shape[-1]
-    logger.info("Recording wrote %d frames (%.2fs)", written, written / SAMPLE_RATE)
+        f.write(output.T)
+    logger.info("Recording wrote %d frames", expected_frames)
 
 
 def keyboard_loop(
@@ -398,9 +419,12 @@ def keyboard_loop(
     type=click.Path(dir_okay=False, path_type=Path),
     default=None,
     help=(
-        "Optional WAV file to tee the first "
-        f"{int(SESSION_RECORDING_MAX_SECONDS)}s of the live audio stream into. "
-        "No-op when not set."
+        f"Optional WAV file to render a deterministic "
+        f"{int(SESSION_RECORDING_DURATION_SECONDS)}s test clip to "
+        f"(middle C from {SESSION_RECORDING_NOTE_START_SECONDS:.1f}-"
+        f"{SESSION_RECORDING_NOTE_END_SECONDS:.1f}s). When set, replaces the live "
+        f"audio stream — output depends only on plugin state, so the same inputs "
+        f"always produce the same WAV. No-op when not set."
     ),
 )
 def main(
@@ -464,9 +488,7 @@ def main(
     stop_event = threading.Event()
     with ThreadPoolExecutor() as pool:
         if session_recording_path is not None:
-            audio_future = pool.submit(
-                play_audio_recorded, plugin, stop_event, session_recording_path
-            )
+            audio_future = pool.submit(play_audio_recorded, plugin, session_recording_path)
         else:
             audio_future = pool.submit(play_audio, plugin, stop_event)
         keyboard_future = pool.submit(keyboard_loop, plugin, stop_event, param_spec_name)
--- a/scripts/surge_xt_interactive.py
+++ b/scripts/surge_xt_interactive.py
@@ -40,6 +40,11 @@ PLAYBACK_SAMPLE_RATE = SAMPLE_RATE
 # Maximum time to wait for the audio thread to drain after ``stop_event`` is set.
 AUDIO_THREAD_DRAIN_TIMEOUT_SECONDS = 2
 
+# Cap on how many seconds of the live session ``play_audio`` will tee into a
+# WAV file when ``--session-recording-path`` is set. Bounded so a long editor
+# session doesn't write an unbounded recording.
+SESSION_RECORDING_MAX_SECONDS = 20.0
+
 # Plugin-flush parameters used by the post-load / pre-render flush pattern; see
 # ``_flush_plugin``. Mirror of the values used in ``src.data.vst.core.render_params``.
 _PLUGIN_FLUSH_DURATION_SECONDS = 32.0
@@ -115,25 +120,6 @@ class DatasetRefType(click.ParamType):
         except ValueError:
             self.fail(f"batch index must be an integer, got {idx_str!r}", param, ctx)
         return DatasetRef(path=Path(path_str), batch_idx=batch_idx)
-
-
-def render_to_wav(plugin: VST3Plugin, output_path: Path, duration_seconds: float) -> None:
-    """Render synthesized audio through ``plugin`` to a WAV file at ``output_path``.
-
-    Offline alternative to :func:`play_audio` for headless environments (e.g. Docker
-    on Linux without ALSA/PulseAudio) where ``AudioStream`` cannot open a device.
-
-    :param plugin: Loaded VST3 plugin (post preset/parameter setup).
-    :param output_path: WAV destination. Parent directory must exist.
-    :param duration_seconds: Seconds of audio to render. The actual length is
-        rounded up to the nearest ``BUFFER_SIZE`` samples.
-    """
-    silence = np.zeros((CHANNELS, BUFFER_SIZE), dtype=np.float32)
-    num_buffers = math.ceil(duration_seconds * SAMPLE_RATE / BUFFER_SIZE)
-    with AudioFile(str(output_path), "w", SAMPLE_RATE, CHANNELS) as f:
-        for _ in range(num_buffers):
-            synth_output = plugin(silence, SAMPLE_RATE, reset=False)
-            f.write(synth_output)
 
 
 def _load_pred_tensor(pred_path: Path) -> torch.Tensor:
@@ -244,29 +230,78 @@ def _flush_plugin(plugin: VST3Plugin) -> None:
     plugin.reset()
 
 
-def play_audio(plugin: VST3Plugin, stop_event: threading.Event) -> None:
+def play_audio(
+    plugin: VST3Plugin,
+    stop_event: threading.Event,
+    session_recording_path: Path | None = None,
+) -> None:
     """Stream silence through Surge XT and write synthesized audio to the output device.
 
     Runs until ``stop_event`` is set (typically by ``keyboard_loop``'s quit action or by
     ``main`` after the plugin editor is closed). Resamples to ``PLAYBACK_SAMPLE_RATE`` if
     it differs from ``SAMPLE_RATE`` so the output device gets a rate it supports.
+
+    When ``session_recording_path`` is provided, the first
+    ``SESSION_RECORDING_MAX_SECONDS`` of the session is teed to that WAV file
+    (at ``SAMPLE_RATE``, pre-resample). After the cap is reached the recording
+    is closed and the live stream continues.
     """
     silence = np.zeros((CHANNELS, BUFFER_SIZE), dtype=np.float32)
     needs_resample = SAMPLE_RATE != PLAYBACK_SAMPLE_RATE
     stream_resampler = (
         StreamResampler(SAMPLE_RATE, PLAYBACK_SAMPLE_RATE, CHANNELS) if needs_resample else None
     )
-    with AudioStream(
-        output_device_name=AudioStream.default_output_device_name,
-        sample_rate=SAMPLE_RATE,
-        buffer_size=BUFFER_SIZE,
-    ) as stream:
-        while not stop_event.is_set():
-            synth_output = plugin(silence, SAMPLE_RATE, reset=False)
-            if stream_resampler is not None:
-                stream.write(stream_resampler.process(synth_output), PLAYBACK_SAMPLE_RATE)
-            else:
-                stream.write(synth_output, SAMPLE_RATE)
+
+    recorder: AudioFile | None = None
+    recorder_buffers_remaining = 0
+    if session_recording_path is not None:
+        recorder = AudioFile(str(session_recording_path), "w", SAMPLE_RATE, CHANNELS)
+        recorder_buffers_remaining = math.ceil(
+            SESSION_RECORDING_MAX_SECONDS * SAMPLE_RATE / BUFFER_SIZE
+        )
+        logger.info(
+            "Recording first %.1fs of the session to %s",
+            SESSION_RECORDING_MAX_SECONDS,
+            session_recording_path,
+        )
+
+    try:
+        with AudioStream(
+            output_device_name=AudioStream.default_output_device_name,
+            sample_rate=PLAYBACK_SAMPLE_RATE,
+            buffer_size=BUFFER_SIZE,
+        ) as stream:
+            while not stop_event.is_set():
+                synth_output = plugin(silence, SAMPLE_RATE, reset=False)
+                # ``plugin(...)`` returns audio shaped (channels, frames). Pin
+                # this invariant — anything else here means a pedalboard API
+                # change or a misconfigured plugin and would silently corrupt
+                # the recording / stream writes downstream.
+                assert synth_output.shape == (CHANNELS, BUFFER_SIZE), (
+                    f"expected synth output shape ({CHANNELS}, {BUFFER_SIZE}), "
+                    f"got {synth_output.shape}"
+                )
+
+                if recorder is not None and recorder_buffers_remaining > 0:
+                    # AudioFile.write expects (frames, channels); plugin output is
+                    # (channels, frames), so transpose before writing.
+                    recorder.write(synth_output.T)
+                    recorder_buffers_remaining -= 1
+                    if recorder_buffers_remaining == 0:
+                        recorder.close()
+                        recorder = None
+                        logger.info(
+                            "Session recording reached %.1fs cap; closed.",
+                            SESSION_RECORDING_MAX_SECONDS,
+                        )
+
+                if stream_resampler is not None:
+                    stream.write(stream_resampler.process(synth_output), PLAYBACK_SAMPLE_RATE)
+                else:
+                    stream.write(synth_output, SAMPLE_RATE)
+    finally:
+        if recorder is not None:
+            recorder.close()
 
 
 def keyboard_loop(
@@ -370,6 +405,16 @@ def keyboard_loop(
         "and written to this file via ``src.data.vst.generate_vst_dataset.make_dataset``."
     ),
 )
+@click.option(
+    "--session-recording-path",
+    type=click.Path(dir_okay=False, path_type=Path),
+    default=None,
+    help=(
+        "Optional WAV file to tee the first "
+        f"{int(SESSION_RECORDING_MAX_SECONDS)}s of the live audio stream into. "
+        "No-op when not set."
+    ),
+)
 def main(
     plugin_path: str,
     pred: PredictionRef | None,
@@ -377,6 +422,7 @@ def main(
     preset_path: str,
     param_spec_name: str,
     output_dataset_path: Path | None,
+    session_recording_path: Path | None,
 ) -> None:
     """Open Surge XT GUI with real-time audio streaming and record patches to an HDF5 dataset.
 
@@ -429,7 +475,7 @@ def main(
 
     stop_event = threading.Event()
     with ThreadPoolExecutor() as pool:
-        audio_future = pool.submit(play_audio, plugin, stop_event)
+        audio_future = pool.submit(play_audio, plugin, stop_event, session_recording_path)
         keyboard_future = pool.submit(keyboard_loop, plugin, stop_event, param_spec_name)
 
         try:
--- a/scripts/surge_xt_interactive.py
+++ b/scripts/surge_xt_interactive.py
@@ -1,7 +1,9 @@
 """Interactive Surge XT preview with real-time audio streaming via pedalboard."""
 
+import logging
 import math
 import threading
+from concurrent.futures import ThreadPoolExecutor
 from dataclasses import dataclass
 from pathlib import Path
 
@@ -13,17 +15,23 @@ import rootutils
 import torch
 from pedalboard import VST3Plugin
 from pedalboard.io import AudioFile, AudioStream, StreamResampler
-from rich import print
 
 from src.data.vst import load_plugin, load_preset, param_specs
 from src.data.vst.core import set_params
+from src.data.vst.generate_vst_dataset import make_dataset
 
 rootutils.setup_root(__file__, indicator=".project-root", pythonpath=True)
 
+logger = logging.getLogger(__name__)
 
 CHANNELS = 2
 SAMPLE_RATE = 44100
 BUFFER_SIZE = 512
+MAKE_DATASET_SCRIPT_PATH = Path("src/data/vst/generate_vst_dataset.py").resolve()
+MAKE_DATASET_VELOCITY = 100
+MAKE_DATASET_SIGNAL_DURATION_SECONDS = 4.0
+MAKE_DATASET_MIN_LOUDNESS = -50.0
+MAKE_DATASET_SAMPLE_BATCH_SIZE = 32
 
 PLAYBACK_SAMPLE_RATE = (
     44100  # Sample rate to use for real-time playback. Must be supported by the output device.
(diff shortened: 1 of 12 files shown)


ktinubu:
## Verification Results

Verified at `8e4bea4`. Environment: Linux devcontainer with `plugins/Surge XT.vst3` present, NVIDIA RTX 5060 Ti available (`torch.cuda.is_available() == True`).

- [x] **All 25 helper-tier tests pass** *(Level 1 — promise is "decode/parse/load helpers behave correctly"; this IS the integration test for those helpers)*
  ```bash
  $ pytest tests/scripts/test_surge_xt_interactive.py -vv
  ============================= 25 passed in 37.60s ==============================
  ```
  Includes the slow `test_loads_row_from_surge_xt_smoke_fixture` which exercises a real `surge_xt_smoke_datasets/test.h5` round-trip through `load_dataset_synth_params`.

- [x] **CLI exposes the new flags and decodes references** *(Level 2 — invoke the script with `--help`; promise is the user-facing CLI surface)*
  ```bash
  $ python3 scripts/surge_xt_interactive.py --help
  Options:
    -p, --plugin-path TEXT      Path to VST3 plugin.
    --pred PRED_REF             Prediction reference as PATH:BATCH_IDX ...
    --dataset-ref DATASET_REF   Dataset reference as PATH:DATASET_IDX ...
    -r, --preset-path TEXT      Base preset to load before applying any
                                --pred / --dataset-ref params.
    --param-spec-name TEXT      Parameter spec name (key into ``param_specs``)
    --output-dataset-path FILE  HDF5 file to append recorded patches to ...
    --help                      Show this message and exit.
  ```

- [x] **`render_to_wav` writes the expected number of buffers and frames against a stub plugin** *(Level 2 — promise is "renders silence-driven plugin output to WAV with correct shape"; ran the helper, opened the WAV, counted frames)*
  ```bash
  $ python3 -c '
  import sys; sys.path.insert(0, ".")
  from pathlib import Path
  import numpy as np, wave
  from scripts.surge_xt_interactive import render_to_wav, BUFFER_SIZE, SAMPLE_RATE
  class C:
      def __init__(self, v): self.v = v; self.n = 0
      def __call__(self, buf, sr, reset=False):
          self.n += 1; return np.full_like(buf, self.v)
  p = C(0.5); out = Path(\"/tmp/render_to_wav_smoke.wav\")
  render_to_wav(p, out, 0.05)
  with wave.open(str(out)) as f:
      print(\"frames:\", f.getnframes(), \"channels:\", f.getnchannels(), \"rate:\", f.getframerate())
  print(\"plugin called\", p.n, \"times — expected\", -(-int(0.05 * SAMPLE_RATE) // BUFFER_SIZE))
  '
  frames: 2560 channels: 2 rate: 44100
  plugin called 5 times — expected 5
  ```

- [x] **PR body's "full-width underscore typo" claim was wrong — confirmed cosmetic only** *(Level 2 — adversarial check; promise was "this would TypeError on keyword call". It does not, because PEP 3131 normalizes identifiers via NFKC at parse time.)*
  ```bash
  $ python3 -c '
  import unicodedata, inspect
  fw = "plugin" + chr(0xff3f) + "path"
  print("NFKC of plugin[U+FF3F]path:", repr(unicodedata.normalize("NFKC", fw)))
  from src.data.vst.core import render_params
  print("runtime param keys:", list(inspect.signature(render_params).parameters.keys()))
  try:
      render_params(plugin_path="x.vst3", params={}, midi_note=60, velocity=100,
                    note_start_and_end=(0,0.5), signal_duration_seconds=1.0,
                    sample_rate=44100, channels=2)
  except FileNotFoundError as e:
      print("FileNotFoundError at plugin load — keyword arg accepted:", str(e)[:80])
  except Exception as e:
      print(f"{type(e).__name__}:", str(e)[:80])
  '
  NFKC of plugin[U+FF3F]path: 'plugin_path'
  runtime param keys: ['plugin_path', 'params', 'midi_note', 'velocity', 'note_start_and_end', 'signal_duration_seconds', 'sample_rate', 'channels', 'preset_path']
  ImportError: Unable to load plugin x.vst3: plugin file not found.
  ```
  PR body updated to label this as cosmetic-only (still worth replacing with ASCII underscore for readability).

- [ ] **Full e2e: `make_dataset` with `fixed_synth_params_list` produces a valid h5 with finite audio + decoded synth params matching the input** *(Level 1 — SKIP for time: each sample takes ~14s wall-clock against the real VST; the existing test runs 5 samples and is gated `@pytest.mark.slow @pytest.mark.requires_vst`. Run on a host with the VST present:)*
  ```bash
  pytest tests/data/vst/test_generate_vst_dataset.py::test_make_dataset_writes_valid_h5 -vv -s
  # expected: passes, decoded synth/note params match input within rel=1e-9 abs=1e-7
  ```

- [ ] **Full interactive flow: `--pred` row applied → editor opens → `p` records → `q` exits → patch appears as new h5 row** *(Level 1 — SKIP: requires an audio output device + GUI display, neither available in this container)*
  ```bash
  # On a host with audio + GUI:
  python scripts/surge_xt_interactive.py \
    --pred outputs/pred-0.pt:0 \
    --output-dataset-path /tmp/recorded.h5
  # 1. confirm Surge XT GUI opens with predicted patch loaded
  # 2. press 'p' twice, 'q' once
  # 3. inspect: python -c "import h5py; print(h5py.File('/tmp/recorded.h5','r')['param_array'].shape)"
  #    expected: (2, <num_params>)
  ```

- [ ] **Devcontainer audio packages install cleanly** *(Level 1 — SKIP: requires rebuilding the devcontainer image, which is out of scope for this verification environment. Reviewer can rebuild with `Dev Containers: Rebuild Container` and confirm `aplay -l` lists ALSA devices.)*

### Adversarial findings worth a follow-up (non-blocking)

1. `--pred` and `--dataset-ref` are not Click-mutually-exclusive — `--dataset-ref` silently wins over `--pred`. Consider `cloup`'s `MutuallyExclusiveOption` or a manual `if pred and dataset_ref: raise click.UsageError(...)`.
2. `play_audio` and `keyboard_loop` exit conditions are checked between iterations only; on `stop_event` set during a `click.getchar()` call the loop won't exit until a key is pressed. Acceptable for the typical "close editor → press q" path; documenting the constraint would help.
3. `_load_pred_tensor` uses `torch.load(..., weights_only=True)`; good. `load_dataset_synth_params` reads h5 unsafely if the file is malformed — relies on `h5py` raising — which is fine but not asserted in tests beyond the out-of-range case.

scripts/surge_xt_interactive.py
```diff
@@ -1,40 +1,412 @@
...
+from src.data.vst.generate_vst_dataset import make_dataset
+
+rootutils.setup_root(__file__, indicator=".project-root", pythonpath=True)
+
+logger = logging.getLogger(__name__)
 
 CHANNELS = 2
 SAMPLE_RATE = 44100
 BUFFER_SIZE = 512
+MAKE_DATASET_SCRIPT_PATH = Path("src/data/vst/generate_vst_dataset.py").resolve()
```
ktinubu: BLOCK: [code-health] `MAKE_DATASET_SCRIPT_PATH` is defined but never referenced anywhere in this file. Dead code — delete it.

scripts/surge_xt_interactive.py
```diff
@@ -1,40 +1,412 @@
...
+
+    :param ref: Reference identifying the h5 file path and row to decode.
+    :param param_spec_name: Parameter spec name (key into ``param_specs``) used
+        to decode the row into raw VST parameter values.
+    :returns: Dict mapping VST parameter name to its raw (decoded) value.
+    """
+    spec = param_specs[param_spec_name]
+    with h5py.File(ref.path, "r") as f:
+        param_array = f["param_array"]
+        assert isinstance(param_array, h5py.Dataset), "expected h5py.Dataset for 'param_array'"
```
ktinubu: BLOCK: [python-style PY4] `assert` for runtime validation is stripped under `python -O`. Use an explicit `if not isinstance(...): raise TypeError(...)` so the check survives optimization.

```suggestion
        if not isinstance(param_array, h5py.Dataset):
            raise TypeError(f"expected h5py.Dataset for 'param_array', got {type(param_array).__name__}")
```

scripts/surge_xt_interactive.py
```diff
@@ -1,40 +1,412 @@
...
+        logger.info("Quitting...")
+        return False  # stop loop
+
+    def record_patch():
+        patch: dict[str, float] = dict()
+        logger.info("Recording patch...")
+        for param_name in spec.synth_param_names:
+            assert param_name in plugin.parameters, (  # pyright: ignore[reportAttributeAccessIssue]
+                f"parameter {param_name!r} not found in plugin parameters"
+            )
```
ktinubu: BLOCK: [python-style PY4] `assert` used for runtime validation; stripped under `python -O`. The check is a real runtime invariant ("plugin doesn't expose this param name") so it should raise unconditionally.

```suggestion
            if param_name not in plugin.parameters:  # pyright: ignore[reportAttributeAccessIssue]
                raise KeyError(f"parameter {param_name!r} not found in plugin parameters")
```

scripts/surge_xt_interactive.py
```diff
@@ -1,40 +1,412 @@
...
+       it to the plugin before the editor opens.
+    3. Open the plugin's GUI editor; in parallel, stream audio to the default output device
+       and run a keyboard loop (press ``p`` to snapshot the current synth params as a patch,
+       ``q`` to quit).
+    4. After the editor is closed, render every recorded patch through the plugin and append
+       the resulting samples to ``--output-dataset-path`` via ``make_dataset``.
+    """
+    logging.basicConfig(level=logging.INFO, format="%(asctime)s %(levelname)s %(message)s")
+    plugin = load_plugin(plugin_path)
+    assert isinstance(plugin, VST3Plugin), f"expected VST3Plugin, got {type(plugin).__name__}"
```
ktinubu: BLOCK: [python-style PY4] `assert` for runtime validation, stripped under `python -O`. If `load_plugin` returns the wrong type this should fail loudly in production too.

```suggestion
    if not isinstance(plugin, VST3Plugin):
        raise TypeError(f"expected VST3Plugin, got {type(plugin).__name__}")
```

scripts/surge_xt_interactive.py
```diff
@@ -1,40 +1,412 @@
...
+        finally:
+            stop_event.set()
+            # surface any exception that happened in the audio thread
+            audio_future.result(timeout=2)
+        logger.info("Editor closed, waiting for audio thread to finish...")
+        synth_patches = keyboard_future.result()
+        logger.info("Recorded %d patches: %s", len(synth_patches), synth_patches)
 
-    plugin.show_editor()
+    assert synth_patches is not None, "synth_patches should be set by keyboard_loop"
```
ktinubu: BLOCK: [python-style PY4] `assert` for runtime validation. Either replace with an explicit raise, or — better — restructure so `synth_patches` is bound inside the `with` block and the post-`with` code can rely on it without a separate check.

scripts/surge_xt_interactive.py
```diff
@@ -1,40 +1,412 @@
...
 BUFFER_SIZE = 512
+MAKE_DATASET_SCRIPT_PATH = Path("src/data/vst/generate_vst_dataset.py").resolve()
+MAKE_DATASET_VELOCITY = 100
+MAKE_DATASET_SIGNAL_DURATION_SECONDS = 4.0
+MAKE_DATASET_MIN_LOUDNESS = -50.0
+MAKE_DATASET_SAMPLE_BATCH_SIZE = 32
+
+PLAYBACK_SAMPLE_RATE = (
+    44100  # Sample rate to use for real-time playback. Must be supported by the output device.
+)
```
ktinubu: WARN: [code-health] `PLAYBACK_SAMPLE_RATE = 44100` equals `SAMPLE_RATE = 44100`, so the resampler branch in `play_audio` (lines 229-234) is dead. Either drop one of the constants and the resampler branch, or pick a real playback rate to make the resampling path live.

scripts/surge_xt_interactive.py
```diff
@@ -1,40 +1,412 @@
...
+import torch
 from pedalboard import VST3Plugin
-from pedalboard.io import AudioStream
+from pedalboard.io import AudioFile, AudioStream, StreamResampler
+
+from src.data.vst import load_plugin, load_preset, param_specs
+from src.data.vst.core import set_params
+from src.data.vst.generate_vst_dataset import make_dataset
+
+rootutils.setup_root(__file__, indicator=".project-root", pythonpath=True)
```
ktinubu: WARN: [code-health] `from src.data.vst...` imports execute *before* `rootutils.setup_root(...)` runs. This works only because the project happens to already be on PYTHONPATH; running the script from outside the project root would fail at import. Move the rootutils setup above the local imports.

scripts/surge_xt_interactive.py
```diff
@@ -1,40 +1,412 @@
...
+    stream_resampler = None
     with AudioStream(
         output_device_name=AudioStream.default_output_device_name,
         sample_rate=SAMPLE_RATE,
         buffer_size=BUFFER_SIZE,
     ) as stream:
-        while True:
+        while not stop_event.is_set():
+            if SAMPLE_RATE != PLAYBACK_SAMPLE_RATE and stream_resampler is None:
+                stream_resampler = StreamResampler(SAMPLE_RATE, PLAYBACK_SAMPLE_RATE, CHANNELS)
```
ktinubu: WARN: [code-health] The `SAMPLE_RATE != PLAYBACK_SAMPLE_RATE` branch is constant for the lifetime of the program — re-checking it on every audio buffer is wasted work and adds reader load. Hoist the resampler creation outside the loop and pick `stream.write(...)` arms based on a single `if`:

```python
needs_resample = SAMPLE_RATE != PLAYBACK_SAMPLE_RATE
stream_resampler = StreamResampler(SAMPLE_RATE, PLAYBACK_SAMPLE_RATE, CHANNELS) if needs_resample else None
```

scripts/surge_xt_interactive.py
```diff
@@ -1,40 +1,412 @@
...
+    Also exits when ``stop_event`` is set externally (e.g. when the plugin editor is
+    closed). Returns the list of recorded patches.
+    """
+    spec = param_specs[param_spec_name]
+    synth_patches: list[dict[str, float]] = []
+
+    def quit_action():
+        stop_event.set()
+        logger.info("Quitting...")
+        return False  # stop loop
```
ktinubu: WARN: [code-health] The `True`/`False` return values encode "keep going" / "stop" but the meaning lives only in the inline comments. Either name the booleans (`_KEEP_LOOPING`, `_STOP_LOOPING`) or have actions return an `enum.Enum` value.

scripts/surge_xt_interactive.py
```diff
@@ -1,40 +1,412 @@
...
+
+    def record_patch():
+        patch: dict[str, float] = dict()
```
ktinubu: WARN: [python-style] `dict()` is slower and noisier than `{}`.

```suggestion
        patch: dict[str, float] = {}
```

ktinubu (commented):
## Code Review — 7 BLOCK / 18 WARN

Reviewed against tdd-implementation, code-health, ml-data-pipeline, synth-setter-project-standards, python-style, and ml-test checklists. Skipped shell-style (no .sh files changed).

### Stale PR body
The PR description still mentions a `.devcontainer/Dockerfile` change, but commit `137815b` reverted that file — the actual diff is now 5 files (912 → 903 lines). Please update the PR body to drop the Dockerfile bullet.

### BLOCK summary
- 5× `assert` used for runtime validation (PY4 — stripped under `python -O`).
- 1× dead module-level constant.
- 2× test-file bugs that hide diagnostics on failure (`name` is undefined in scope; `log.info(spec.names, ...)` is malformed).

### WARN themes
- `--pred` / `--dataset-ref` silently choose between each other.
- `Tuple` / `Optional` / `List` → use built-in generics (project standard).
- Magic numbers (`32.0`, `2048`, `0.5`, `2`) want named constants.
- Test design: the e2e test pins all 5 rows to the same patch — doesn't actually exercise per-row indexing of `fixed_synth_params_list`.
- `from rich import print` shadows the builtin in a test file.
- Plugin reload now happens per-sample inside `render_params` — significant perf cost (~14s/sample observed). Document the trade-off so future readers understand it's a deliberate workaround for the silent-output bug.

### Positives worth calling out
- `decode_prediction_row` is a clean pure function (functional core / imperative shell separation done right).
- `PredictionRefType` / `DatasetRefType` validate input at the CLI boundary with focused tests covering malformed input.
- New test file exercises round-trips and edge cases (out-of-range index, malformed URI) where it matters.
- Audio thread error surfacing via `audio_future.result(timeout=2)` correctly addresses the daemon-thread silent-failure concern from #532.

scripts/surge_xt_interactive.py
```diff
@@ -1,40 +1,412 @@
...
+    logging.basicConfig(level=logging.INFO, format="%(asctime)s %(levelname)s %(message)s")
+    plugin = load_plugin(plugin_path)
+    assert isinstance(plugin, VST3Plugin), f"expected VST3Plugin, got {type(plugin).__name__}"
+
+    load_preset(plugin, preset_path)
+
+    # flush plugin to ensure that full parameter dict is populated.
+    plugin.process([], 32.0, SAMPLE_RATE, CHANNELS, 2048, True)  # flush
+    plugin.reset()
+    plugin.process([], 32.0, SAMPLE_RATE, CHANNELS, 2048, True)  # flush
```
ktinubu: WARN: [code-health] Two identical `plugin.process([], 32.0, SAMPLE_RATE, CHANNELS, 2048, True)` lines with magic numbers `32.0` (process duration) and `2048` (buffer size). Extract a `_flush_plugin(plugin)` helper and name the constants — same flush sequence already exists in `src/data/vst/core.py:render_params`, so the helper could live there.

scripts/surge_xt_interactive.py
```diff
@@ -1,40 +1,412 @@
...
+    plugin.reset()
+
+    if dataset_ref is not None:
+        synth_params = load_dataset_synth_params(dataset_ref, param_spec_name)
+        set_params(plugin, synth_params)
+    elif pred is not None:
+        synth_params = load_prediction_synth_params(pred, param_spec_name)
+        set_params(plugin, synth_params)
```
ktinubu, line 486: WARN: [code-health, safe-defaults] `--pred` and `--dataset-ref` aren't mutually exclusive — when both are passed, `dataset_ref` silently wins. Reject the conflict explicitly so users don't get surprised:

```suggestion
    if dataset_ref is not None and pred is not None:
        raise click.UsageError("--pred and --dataset-ref are mutually exclusive; pass at most one.")
    if dataset_ref is not None:
        synth_params = load_dataset_synth_params(dataset_ref, param_spec_name)
        set_params(plugin, synth_params)
    elif pred is not None:
        synth_params = load_prediction_synth_params(pred, param_spec_name)
        set_params(plugin, synth_params)
```

scripts/surge_xt_interactive.py
```diff
@@ -1,40 +1,412 @@
...
+        keyboard_future = pool.submit(keyboard_loop, plugin, stop_event, param_spec_name)
 
-    t = threading.Thread(target=play_audio, args=(plugin,), daemon=True)
-    t.start()
+        try:
+            plugin.show_editor(stop_event)
+        finally:
+            stop_event.set()
+            # surface any exception that happened in the audio thread
+            audio_future.result(timeout=2)
```
ktinubu: WARN: [code-health] Magic timeout `2`. Hoist to a named module constant (e.g. `AUDIO_THREAD_DRAIN_TIMEOUT_SECONDS = 2`).

Related (line 388): `keyboard_future.result()` has no timeout while `click.getchar()` is blocking — if the editor closes without a `q` keypress, this hangs forever. Worth calling out in the docstring or fixing with a sentinel/select-based read.

src/data/vst/core.py
```diff
@@ -10,38 +10,22 @@
...
-    finally:
-        t.join()
-
-
-def _prepare_plugin(plugin: VST3Plugin) -> None:
-    _call_with_interrupt(plugin.show_editor, sleep_time=2.0)
-
+def _prepare_plugin(stop_event: threading.Event, sleep_time: float = 0.5) -> None:
+    time.sleep(sleep_time)
+    stop_event.set()
```
ktinubu: WARN: [code-health, python-style PY6] `_prepare_plugin` has no docstring, and `sleep_time: float = 0.5` is a magic default. The branch's commit history (`c6e4139 even with 0.5, seems to still work`, `9ff7f16 ...skips every other render because loudness is too low`) shows this value was experimentally tuned to dodge a timing race in `show_editor`. Add a one-line docstring explaining what the wait is for and why 0.5s.

Separately: the thread `t` started at line 23 is never `join()`-ed. It self-terminates after the sleep, but `t.join(timeout=...)` after `finally:` would be tidier.

src/data/vst/core.py
```diff
@@ -62,7 +46,7 @@ def write_wav(audio: np.ndarray, path: str, sample_rate: float, channels: int) -
 
 
 def render_params(
-    plugin: VST3Plugin,
+    plugin_path: str,
```
ktinubu, line 67: WARN: [code-health] The parameter is declared `plugin＿path` with `U+FF3F` (FULLWIDTH LOW LINE) instead of the ASCII underscore. PEP 3131 NFKC-normalizes identifiers at parse time so it works at runtime, but it will surprise every reader who looks at this file. Replace with the ASCII underscore.

```suggestion
    plugin_path: str,
```

src/data/vst/core.py
```diff
@@ -72,6 +56,7 @@ def render_params(
     channels: int,
     preset_path: Optional[str] = None,
 ) -> np.ndarray:
+    plugin = load_plugin(plugin_path)
```
ktinubu, line 92: WARN: [code-health, ml-pipeline] `render_params` now reloads the plugin (and shows + closes its editor) on every call. For batch generation that's ~14s/sample (verified during PR verification). The PR commit history makes clear this is the deliberate fix for the silent-output bug from `9ff7f16`/`086d80f`, but a docstring saying "reloads per call to work around a render-stale-state bug; see commits 086d80f, 9ff7f16" would save the next reader the archaeology.

src/data/vst/generate_vst_dataset.py
```diff
@@ -223,6 +229,8 @@ def make_dataset(
     min_loudness: float,
     param_spec: ParamSpec,
     sample_batch_size: int,
+    fixed_synth_params_list: List[dict[str, float]] | None = None,
+    fixed_note_params_list: List[dict[str, float]] | None = None,
```
ktinubu: WARN: [python-style PY1, code-health] Project standard is built-in generics. Also: `fixed_synth_params_list[i]` (line 269) is indexed without a length check; if the caller passes a shorter list than `num_samples`, an `IndexError` surfaces mid-iteration with no useful context. Validate up front in `make_dataset`.

```suggestion
    fixed_synth_params_list: list[dict[str, float]] | None = None,
    fixed_note_params_list: list[dict[str, float]] | None = None,
```

…and add a precondition near the top of the function body:
```python
for name, lst in [("fixed_synth_params_list", fixed_synth_params_list),
                  ("fixed_note_params_list", fixed_note_params_list)]:
    if lst is not None and len(lst) < num_samples:
        raise ValueError(f"{name} has length {len(lst)}, expected at least {num_samples}")
```

src/data/vst/generate_vst_dataset.py
```diff
@@ -60,23 +60,29 @@ def make_spectrogram(audio: np.ndarray, sample_rate: float) -> np.ndarray:
...
+    fixed_synth_params: dict[str, float] | None = None,
+    fixed_note_params: dict[str, float] | None = None,
 ) -> VSTDataSample:
     while True:
         logger.debug("sampling params")
         synth_params, note_params = param_spec.sample()
+        if fixed_synth_params is not None:
+            synth_params = fixed_synth_params
+        if fixed_note_params is not None:
+            note_params = fixed_note_params
```
ktinubu: WARN: [code-health] When `fixed_synth_params` / `fixed_note_params` are passed, the upstream `logger.debug("sampling params")` and `logger.debug("sampling note")` lines (75 and 82) still fire even though nothing is sampled — misleading for anyone debugging the new "replay captured patch" path. Either move the logs into the not-fixed branch or reword them.

src/data/vst/generate_vst_dataset.py
```diff
@@ -242,22 +250,24 @@ def make_dataset(
     audio_dataset.attrs["channels"] = channels
     audio_dataset.attrs["min_loudness"] = min_loudness
 
-    plugin = load_plugin(plugin_path)
+
```
ktinubu: WARN: [code-health] Stray double blank line where `plugin = load_plugin(plugin_path)` was removed. Trim back to the standard separator.

tests/data/vst/test_generate_vst_dataset.py
```diff
@@ -0,0 +1,153 @@
...
+
+
+@pytest.mark.slow
+@pytest.mark.requires_vst
+@skip_no_vst
+def test_make_dataset_writes_valid_h5(tmp_path: Path) -> None:
+    """make_dataset produces an h5 with correct datasets, shapes, attrs, and finite samples."""
+    out = tmp_path / "data.h5"
+    spec = param_specs[_SPEC_NAME]
+    log.info(spec.names, spec.synth_param_length, spec.note_param_length)
```
ktinubu: BLOCK: [bug] `log.info(spec.names, spec.synth_param_length, spec.note_param_length)` passes three positional args. `logger.info(msg, *args)` treats the first arg as the format string and the rest as `%`-args, but `spec.names` (likely a list) has no `%s` placeholders — so this either logs nothing useful or raises `TypeError` at format time depending on the value. Use a real format string:

```suggestion
    log.info("spec %s synth_len=%d note_len=%d", spec.names, spec.synth_param_length, spec.note_param_length)
```

tests/data/vst/test_generate_vst_dataset.py
```diff
@@ -0,0 +1,153 @@
...
+        for i in range(_NUM_SAMPLES):
+            encoded = params[i]
+            decoded_synth_params, decoded_note_params = spec.decode(encoded)
+            assert isinstance(decoded_synth_params, dict)
+            original_synth_params = SYNTH_PATCHES[i]
+            assert decoded_synth_params == pytest.approx(
+                original_synth_params, rel=_RELATIVE_TOLERANCE, abs=_ABSOLUTE_TOLERANCE
+            ), (
+                f"decoded synth param `{name}` value {decoded_synth_params} does not match original {original_synth_params} within tolerances (abs={_ABSOLUTE_TOLERANCE}, rel={_RELATIVE_TOLERANCE})"
+            )
```
ktinubu: BLOCK: [bug] The assertion message uses `name`, which is only defined in the unrelated outer `for name in ("audio", "mel_spec", "param_array"):` loop on line 89 — so on assertion failure this raises `NameError` in the f-string and hides the real diagnostic. Drop the bare reference and include `i` and the diff instead.

```suggestion
            assert decoded_synth_params == pytest.approx(
                original_synth_params, rel=_RELATIVE_TOLERANCE, abs=_ABSOLUTE_TOLERANCE
            ), (
                f"sample {i}: decoded synth params {decoded_synth_params} do not match original {original_synth_params} within tolerances (abs={_ABSOLUTE_TOLERANCE}, rel={_RELATIVE_TOLERANCE})"
            )
```

tests/data/vst/test_generate_vst_dataset.py
```diff
@@ -0,0 +1,153 @@
...
+import os
+from pathlib import Path
+
+import h5py
+import hdf5plugin  # noqa: F401  # pyright: ignore[reportUnusedImport]  side-effect: registers Blosc2 filter for h5py reads
+import numpy as np
+import pytest
+import logging
+log = logging.getLogger(__name__)
+from rich import print   # shadows builtin print
```
ktinubu: WARN: [python-style PY1] `import logging` belongs in the stdlib group at the top with `import os`, not after the third-party `pytest`. The `log = logging.getLogger(__name__)` assignment also belongs after all imports. And `from rich import print` shadows the builtin `print` for this whole test module — confusing in tests. Either drop the rich import (you mostly use `log.info`) or alias it:

```suggestion
import logging
import os
from pathlib import Path

import h5py
import hdf5plugin  # noqa: F401  # pyright: ignore[reportUnusedImport]  side-effect: registers Blosc2 filter for h5py reads
import numpy as np
import pytest
from rich import print as rprint

log = logging.getLogger(__name__)
```

tests/data/vst/test_generate_vst_dataset.py
```diff
@@ -0,0 +1,153 @@
...
+    NOTE_PATCHES = []
+    single_synth_patch, single_set_note_params = spec.sample()
+    for i in range(_NUM_SAMPLES):
+        # patch, _ = spec.sample()
+        # for param_name, param_value in patch.items():
+        #     patch[param_name] = float(0)
+        # print(f"Sampled synth params for sample {i}: {patch}")
+
+        SYNTH_PATCHES.append(single_synth_patch)
+        NOTE_PATCHES.append(single_set_note_params)
```
ktinubu: WARN: [python-style PY5, ml-test MT15] `SYNTH_PATCHES` / `NOTE_PATCHES` are local variables but use UPPER_CASE (reserved for module-level constants). Also: assigning the *same* `single_synth_patch` to all `_NUM_SAMPLES` entries means the test never exercises per-row indexing of `fixed_synth_params_list[i]` — a future bug that drops the `[i]` would still pass. Use distinct samples per index.

```suggestion
    synth_patches: list[dict[str, float]] = []
    note_patches: list[dict[str, float]] = []
    for _ in range(_NUM_SAMPLES):
        s, n = spec.sample()
        synth_patches.append(s)
        note_patches.append(n)
```

(Then update `SYNTH_PATCHES` → `synth_patches` and `NOTE_PATCHES` → `note_patches` in the rest of the test.)

tests/data/vst/test_generate_vst_dataset.py
```diff
@@ -0,0 +1,153 @@
...
+    spec = param_specs[_SPEC_NAME]
+    log.info(spec.names, spec.synth_param_length, spec.note_param_length)
+    SYNTH_PATCHES = []
+    NOTE_PATCHES = []
+    single_synth_patch, single_set_note_params = spec.sample()
+    for i in range(_NUM_SAMPLES):
+        # patch, _ = spec.sample()
+        # for param_name, param_value in patch.items():
+        #     patch[param_name] = float(0)
+        # print(f"Sampled synth params for sample {i}: {patch}")
```
ktinubu: WARN: [code-health] Commented-out scratch code. Remove it — git history will hold the diff if you ever need it back.

tests/data/vst/test_generate_vst_dataset.py
```diff
@@ -0,0 +1,153 @@
...
+        assert isinstance(mel, h5py.Dataset)
+        assert isinstance(params, h5py.Dataset)
+
+        assert audio.shape == expected_audio_shape
+        assert mel.shape == expected_mel_shape
+        assert params.shape == (_NUM_SAMPLES, len(spec))
+
+        log.info(
+            f"params dtype: {params.dtype}, param.shape: {params.shape}, params[0].shape: {params[0].shape}, params[0]: {params[0]}, param[0][0]: {params[0][0]}"
+        )
```
ktinubu: WARN: [python-style PY14] Logging here uses an f-string. Project standard is pattern-parameter logging so the formatter only runs at the configured level. Same applies to lines 143, 145, 149, 151.

tests/scripts/test_surge_xt_interactive.py
```diff
@@ -0,0 +1,341 @@
...
+        simple_pred_tensor: torch.Tensor,
+        param_name: str,
+        expected: float,
+    ) -> None:
+        """Each column is rescaled from ``[-1, 1]`` to ``[0, 1]`` and clipped before decoding."""
+        synth_params = surge_xt_interactive.decode_prediction_row(
+            simple_pred_tensor, batch_idx=0, param_spec_name=SURGE_SIMPLE
+        )
+        assert synth_params[param_name] == pytest.approx(expected, abs=1e-6)
+
```
ktinubu, line 102: WARN: [tdd T7 — no logic in tests] `0.5 * 0.77` mirrors the `surge_simple` spec's encoded max value. If the spec changes, both production and the test break consistently — masking real regressions. Hardcode the literal expected value to make the test catch encoding changes loudly.

```suggestion
            ("a_amp_eg_decay", 0.385),
```

scripts/surge_xt_interactive.py
```diff
@@ -1,40 +1,412 @@
...
+        logger.info("Editor closed, waiting for audio thread to finish...")
+        synth_patches = keyboard_future.result()
```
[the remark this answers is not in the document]
ktinubu: Fixed in 410e75b: caught `TimeoutError` from `audio_future.result(...)` and converted it to a warning log so a slow stream close doesn't crash shutdown. Hoisted the timeout to a named constant (`AUDIO_THREAD_DRAIN_TIMEOUT_SECONDS`).

scripts/surge_xt_interactive.py
```diff
@@ -1,40 +1,412 @@
...
+            preset_path=preset_path,
+            sample_rate=SAMPLE_RATE,
+            channels=CHANNELS,
+            velocity=MAKE_DATASET_VELOCITY,
+            signal_duration_seconds=MAKE_DATASET_SIGNAL_DURATION_SECONDS,
+            min_loudness=MAKE_DATASET_MIN_LOUDNESS,
+            param_spec=param_specs[param_spec_name],
+            sample_batch_size=MAKE_DATASET_SAMPLE_BATCH_SIZE,
+            fixed_synth_params_list=synth_patches,
+        )
```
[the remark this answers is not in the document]
ktinubu: Fixed in 410e75b: the call site now reads any existing `audio` dataset's row count, computes `total_samples = existing + len(synth_patches)`, and passes that as `num_samples`. Combined with the relative-indexing fix at #3161493633 below, resume mode now correctly appends instead of truncating.

tests/data/vst/test_generate_vst_dataset.py
```diff
@@ -0,0 +1,153 @@
...
+import hdf5plugin  # noqa: F401  # pyright: ignore[reportUnusedImport]  side-effect: registers Blosc2 filter for h5py reads
+import numpy as np
+import pytest
+import logging
+log = logging.getLogger(__name__)
+from rich import print   # shadows builtin print
+
+
+from src.data.vst import param_specs
+from src.data.vst.generate_vst_dataset import make_dataset
```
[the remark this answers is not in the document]
ktinubu: Fixed in 410e75b: stdlib imports moved up, `rich.print` removed (we only use `log.info`), `log = logging.getLogger(__name__)` moved after all imports. Ruff is now happy.

tests/data/vst/test_generate_vst_dataset.py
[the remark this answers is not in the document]
ktinubu: Fixed in 410e75b: the test was rewritten as part of the e2e split, and the offending log line is now `log.info("spec %s synth_len=%d note_len=%d", spec.names, spec.synth_param_length, spec.note_param_length)` with a proper format string.

tests/data/vst/test_generate_vst_dataset.py
```diff
@@ -0,0 +1,153 @@
...
+
+        for i in range(_NUM_SAMPLES):
+            encoded = params[i]
+            decoded_synth_params, decoded_note_params = spec.decode(encoded)
+            assert isinstance(decoded_synth_params, dict)
+            original_synth_params = SYNTH_PATCHES[i]
+            assert decoded_synth_params == pytest.approx(
+                original_synth_params, rel=_RELATIVE_TOLERANCE, abs=_ABSOLUTE_TOLERANCE
+            ), (
+                f"decoded synth param `{name}` value {decoded_synth_params} does not match original {original_synth_params} within tolerances (abs={_ABSOLUTE_TOLERANCE}, rel={_RELATIVE_TOLERANCE})"
```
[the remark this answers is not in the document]
ktinubu: Fixed in 410e75b: the e2e test was split in two (one fixed-params, one random-sampling) and the assertion message was rewritten to use `i` from the per-sample loop instead of the unrelated outer `name` variable.

src/data/vst/core.py
```diff
@@ -72,6 +56,7 @@ def render_params(
...
     if preset_path is not None:
```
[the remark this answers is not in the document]
ktinubu: Worth noting: the runtime wasn't actually broken — PEP 3131 NFKC-normalizes identifiers at parse time, so `plugin＿path` (U+FF3F) and `plugin_path` (U+005F) are the same identifier (verified via `inspect.signature()`). That said, it was opaque code, so 410e75b replaces it with the ASCII underscore for readability.

src/data/vst/core.py
```diff
@@ -10,38 +10,22 @@
...
     logger.info("Preparing plugin for preset load...")
-    _prepare_plugin(p)
-    logger.info("Plugin ready")
+    stop_event = threading.Event()
+    t = threading.Thread(target=_prepare_plugin, args=(stop_event,))
+    t.start()
+    try:
+        p.show_editor(stop_event)
+    finally:
+        stop_event.set()
```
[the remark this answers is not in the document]
ktinubu: Fixed in 410e75b: `t.join(timeout=_PREPARE_PLUGIN_JOIN_TIMEOUT_SECONDS)` now runs in `finally:` after `stop_event.set()`, so the helper thread doesn't outlive its caller.

scripts/surge_xt_interactive.py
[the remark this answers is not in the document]
ktinubu: Fixed in 410e75b: `rootutils.setup_root(...)` now runs immediately after `import rootutils` and before any `from src.data.vst...` import. The downstream third-party + local imports use `# noqa: E402` since they're intentionally below the rootutils call.

src/data/vst/generate_vst_dataset.py
```diff
@@ -242,22 +250,24 @@ def make_dataset(
...
             velocity=velocity,
             signal_duration_seconds=signal_duration_seconds,
             sample_rate=sample_rate,
             channels=channels,
             min_loudness=min_loudness,
             param_spec=param_spec,
             preset_path=preset_path,
+            fixed_synth_params=fixed_synth_params_list[i] if fixed_synth_params_list else None,
+            fixed_note_params=fixed_note_params_list[i] if fixed_note_params_list else None,
         )
```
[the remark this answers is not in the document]
ktinubu: Fixed in 410e75b: `make_dataset` now indexes `fixed_*_params_list[i - start_idx]` and validates that each provided list has at least `num_samples - start_idx` entries up front (`raise ValueError` with both numbers in the message). The interactive call site at `surge_xt_interactive.py` was updated to compute `total_samples = existing + len(synth_patches)` to match.

scripts/surge_xt_interactive.py
```diff
@@ -1,40 +1,412 @@
...
+    actions = {
+        "q": quit_action,
+        "p": record_patch,
+    }
+    while not stop_event.is_set():
+        ch = click.getchar()
+        action = actions.get(ch)
+        if action and not action():
+            return synth_patches
+    return synth_patches
```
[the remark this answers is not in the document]
ktinubu: Documented in 410e75b: added a NOTE comment in `keyboard_loop` calling out that `click.getchar()` blocks between iterations, so `stop_event` only takes effect on the next keystroke. The full event-driven fix (sentinel keypress, `select` on stdin) is a larger refactor — happy to file a follow-up issue if you'd like.

scripts/surge_xt_interactive.py
ktinubu: Fixed in 410e75b: dead constant deleted.

scripts/surge_xt_interactive.py
ktinubu: Fixed in 410e75b: replaced with `if not isinstance(...): raise TypeError(...)`.

scripts/surge_xt_interactive.py
ktinubu: Fixed in 410e75b: replaced with `if param_name not in plugin.parameters: raise KeyError(...)`.

scripts/surge_xt_interactive.py
ktinubu: Fixed in 410e75b: replaced with `if not isinstance(plugin, VST3Plugin): raise TypeError(...)`.

scripts/surge_xt_interactive.py
ktinubu: Fixed in 410e75b: removed the assert and replaced with `if not synth_patches: log + return` so the empty-list case is handled explicitly without needing the None check.

scripts/surge_xt_interactive.py
ktinubu: Fixed in 410e75b: `PLAYBACK_SAMPLE_RATE = SAMPLE_RATE` now with a comment explaining what to change for a different output device. The `play_audio` resampler branch is dormant by default but ready when the constants diverge.

scripts/surge_xt_interactive.py
ktinubu: Fixed in 410e75b: `rootutils.setup_root(...)` now runs before `from src.data.vst...` imports. Same fix covers Copilot's #3161493582.

scripts/surge_xt_interactive.py
ktinubu: Fixed in 410e75b: `needs_resample` is computed once outside the `with AudioStream(...)` block and the resampler is created up front; the loop body now just picks one of two write arms based on `stream_resampler is not None`.

scripts/surge_xt_interactive.py
ktinubu: Fixed in 410e75b: introduced `_KEEP_LOOPING = True` / `_STOP_LOOPING = False` module constants; `quit_action` returns `_STOP_LOOPING`, `record_patch` returns `_KEEP_LOOPING`, and the loop checks `action() == _STOP_LOOPING`.

scripts/surge_xt_interactive.py
ktinubu: Fixed in 410e75b: `dict()` → `{}`.

scripts/surge_xt_interactive.py
ktinubu: Fixed in 410e75b: extracted `_flush_plugin(plugin)` helper using named constants `_PLUGIN_FLUSH_DURATION_SECONDS` and `_PLUGIN_FLUSH_BUFFER_SIZE`. Also flagged in the helper docstring that `src.data.vst.core.render_params` mirrors the same pattern; consolidating those is a follow-up.

scripts/surge_xt_interactive.py
ktinubu: Fixed in 410e75b: `main` now raises `click.UsageError("--pred and --dataset-ref are mutually exclusive; pass at most one.")` when both are non-None.

scripts/surge_xt_interactive.py
ktinubu: Fixed in 410e75b: hoisted to `AUDIO_THREAD_DRAIN_TIMEOUT_SECONDS = 2` and wrapped the call in `try/except TimeoutError` so a slow shutdown logs a warning instead of crashing. Same fix covers Copilot's #3161493333.

src/data/vst/core.py
ktinubu: Fixed in 410e75b: `_prepare_plugin` now has a docstring; `_PREPARE_PLUGIN_SLEEP_SECONDS = 0.5` and `_PREPARE_PLUGIN_JOIN_TIMEOUT_SECONDS = 1.0` are documented constants citing the experimentation commits (086d80f, 9ff7f16, c6e4139). The thread is joined in `finally:`.

src/data/vst/core.py
ktinubu: Fixed in 410e75b: replaced U+FF3F with ASCII underscore. (As noted on Copilot's #3161493512, runtime was actually fine via PEP 3131 NFKC — but ASCII is what every reader expects.)

src/data/vst/core.py
ktinubu: Fixed in 410e75b: added a docstring to `render_params` explaining the per-call plugin reload, the silent-output bug it works around, and the cited commits (086d80f, 9ff7f16, c6e4139).

src/data/vst/generate_vst_dataset.py
ktinubu: Fixed in 410e75b: `List[...]` → `list[...]`; added a length precondition that raises `ValueError` with both `len(lst)` and the expected `num_samples - start_idx`. Combined with the relative indexing fix (#3161493633), append-mode is now safe.

src/data/vst/generate_vst_dataset.py
ktinubu: Fixed in 410e75b: the "sampling params and note from param_spec" debug log only fires when at least one of `fixed_synth_params` / `fixed_note_params` is `None`; if both are provided we skip `param_spec.sample()` entirely.

src/data/vst/generate_vst_dataset.py
ktinubu: Fixed in 410e75b: the surrounding refactor (length validation moved before dataset creation, indexing fix, log fix) cleaned up the stray blank line as a side effect.

tests/data/vst/test_generate_vst_dataset.py
ktinubu: Fixed in 410e75b: same fix as Copilot's #3161493435 — proper `%`-format string.

tests/data/vst/test_generate_vst_dataset.py
ktinubu: Fixed in 410e75b: same fix as Copilot's #3161493471 — assertion message now uses `i` from the per-sample loop.

tests/data/vst/test_generate_vst_dataset.py
ktinubu: Fixed in 410e75b: same fix as Copilot's #3161493418 — `import logging` moved up, `from rich import print` removed, `log = logging.getLogger(__name__)` placed after imports.

tests/data/vst/test_generate_vst_dataset.py
ktinubu: Fixed in 410e75b: split the e2e test into `test_make_dataset_with_fixed_params_round_trips_per_row` (which builds `synth_patches` / `note_patches` from N independent `spec.sample()` calls so per-row indexing is actually exercised) and `test_make_dataset_with_random_sampling_produces_valid_h5` (which omits `fixed_*_params_list`). Names are now lowercase.

tests/data/vst/test_generate_vst_dataset.py
ktinubu: Fixed in 410e75b: removed.

tests/data/vst/test_generate_vst_dataset.py
ktinubu: Fixed in 410e75b: removed those f-string log lines as part of the test split. The single retained `log.info` uses pattern parameters.

tests/scripts/test_surge_xt_interactive.py
ktinubu: Fixed in 410e75b: hardcoded `0.385` (with a comment noting the spec midpoint). A spec change to the encoded max will now produce a clean test failure instead of silently passing both sides.

ktinubu (commented):
Summary: 1 BLOCK, 4 WARN

Reviewed against tdd-implementation, code-health, ml-data-pipeline, synth-setter-project-standards, python-style, shell-style, ml-test checklists.

The BLOCK is on incorrect type annotations for the new `fixed_note_params` / `fixed_note_params_list` kwargs — `dict[str, float]` doesn't match the actual `{pitch: int, note_start_and_end: tuple}` shape (the test file even asserts the int/tuple types explicitly). The WARNs are misleading help text on `--output-dataset-path` (claims append support that `make_dataset` doesn't actually provide for a finalized h5), some convoluted fixed-vs-sampled branching in `generate_sample`, and minor test-code hygiene items.

src/data/vst/generate_vst_dataset.py
```diff
@@ -60,23 +60,29 @@ def make_spectrogram(audio: np.ndarray, sample_rate: float) -> np.ndarray:
...
+    plugin_path: str,
     velocity: int,
     signal_duration_seconds: float,
     sample_rate: float,
     channels: int,
     min_loudness: float,
     param_spec: ParamSpec,
     preset_path: str,
+    fixed_synth_params: dict[str, float] | None = None,
+    fixed_note_params: dict[str, float] | None = None,
```
ktinubu: BLOCK: [project] `fixed_note_params: dict[str, float]` is the wrong value type. Note params contain `pitch` (int — passed as `midi_note` to `render_params`) and `note_start_and_end` (tuple of floats). The new test in `tests/data/vst/test_generate_vst_dataset.py` asserts both `isinstance(decoded_note_params['pitch'], int)` and `isinstance(decoded_note_params['note_start_and_end'], tuple)`, so the annotation is provably wrong. Same issue at line 233 (`fixed_note_params_list: list[dict[str, float]]`). Use a precise type — e.g. a `TypedDict` with `pitch: int` and `note_start_and_end: tuple[float, float]` — to keep type-safety guarantees.

src/data/vst/generate_vst_dataset.py
```diff
@@ -60,23 +60,29 @@ def make_spectrogram(audio: np.ndarray, sample_rate: float) -> np.ndarray:
...
-
-        logger.debug("sampling note")
+        if fixed_synth_params is not None and fixed_note_params is not None:
+            synth_params = fixed_synth_params
+            note_params = fixed_note_params
+        else:
+            logger.debug("sampling params and note from param_spec")
+            sampled_synth, sampled_note = param_spec.sample()
+            synth_params = fixed_synth_params if fixed_synth_params is not None else sampled_synth
+            note_params = fixed_note_params if fixed_note_params is not None else sampled_note
```
ktinubu: WARN: [code-health] The first branch is logically subsumed by the second (when both are non-None, the else still produces correct output, only with an extra `param_spec.sample()` call). Keeping both branches just to skip sampling adds two indentation levels and three extra lines for a tiny perf win that the surrounding render cost dominates. Either flatten it or comment that the early-return is a perf optimization.

```suggestion
        if fixed_synth_params is None or fixed_note_params is None:
            logger.debug("sampling params and note from param_spec")
            sampled_synth, sampled_note = param_spec.sample()
        synth_params = fixed_synth_params if fixed_synth_params is not None else sampled_synth
        note_params = fixed_note_params if fixed_note_params is not None else sampled_note
```

scripts/surge_xt_interactive.py
```diff
@@ -1,40 +1,477 @@
...
+        "rows applied to the plugin and to enumerate which synth params are captured when "
+        "recording patches."
+    ),
+)
+@click.option(
+    "--output-dataset-path",
+    type=click.Path(dir_okay=False, path_type=Path),
+    default=None,
+    help=(
+        "HDF5 file to append recorded patches to. After the editor is closed, patches "
```
ktinubu: WARN: [code-health] Help text says "HDF5 file to **append** recorded patches to" but `make_dataset` creates fixed-size datasets without `maxshape`, so a true append on a finalized h5 will fail when the underlying datasets need to grow. The docs guide (`docs/guides/surge-xt-interactive.md`) explicitly notes this limitation ("No append support for `--output-dataset-path`"), and the existing-rows handling at lines 422–432 will compute a `total_samples` that the writer can't actually accommodate on a complete file. Either rewrite the help to match reality (e.g. "HDF5 file to write recorded patches to (use a fresh path per session — append to a finalized file is not supported)"), or add `maxshape=(None, ...)` in `create_dataset_and_get_first_unwritten_idx` so append actually works.

tests/data/vst/test_generate_vst_dataset.py
```diff
@@ -0,0 +1,306 @@
...
+    # inputs for the second ``make_dataset`` run — guaranteed past the loudness
+    # gate by construction (the candidate render survived stage 1).
+    synth_patches: list[dict[str, float]] = []
+    note_patches: list[dict[str, float]] = []
+    for i in range(_NUM_SAMPLES):
+        decoded_synth_params, decoded_note_params = spec.decode(expected_params[i])
+        synth_patches.append(decoded_synth_params)
+        note_patches.append(decoded_note_params)
+    print("synth_patches", synth_patches)
+    print("note_patches", note_patches)
```
ktinubu: WARN: [project] These shadow-`print` calls (via `from rich import print`) are debug aids inside an integration test, but a `log = logging.getLogger(__name__)` is already in scope at line 18. Standardize on the logger so test output stays structured and respects pytest's `--log-cli-level` controls.

```suggestion
    log.info("synth_patches: %s", synth_patches)
    log.info("note_patches: %s", note_patches)
```

tests/data/vst/test_generate_vst_dataset.py
```diff
@@ -0,0 +1,306 @@
...
+        assert audio.attrs["signal_duration_seconds"] == _DURATION
+        assert audio.attrs["min_loudness"] == _MIN_LOUDNESS
+
+        audio_arr = audio[...].astype(np.float32)
+        assert np.isfinite(audio_arr).all()
+        assert np.isfinite(mel[...]).all()
+        assert np.isfinite(params[...]).all()
+
+        peak = np.abs(audio_arr).reshape(num_samples, -1).max(axis=1)
+        assert (peak > 1e-4).all(), f"silent clips: peaks={peak.tolist()}"
```
ktinubu: WARN: [code-health] Magic number `1e-4` for the silence-floor threshold is duplicated in concept with the `min_loudness` config above. Promote to a module-level named constant (e.g. `_SILENCE_PEAK_THRESHOLD = 1e-4`) so the assertion's intent is self-documenting and tunable in one place.

src/data/vst/core.py
```diff
@@ -62,16 +67,28 @@ def write_wav(audio: np.ndarray, path: str, sample_rate: float, channels: int) -
...
     params: dict[str, float],
     midi_note: int,
     velocity: int,
-    note_start_and_end: Tuple[float, float],
+    note_start_and_end: tuple[float, float],
     signal_duration_seconds: float,
     sample_rate: float,
     channels: int,
-    preset_path: Optional[str] = None,
+    preset_path: str | None = None,
```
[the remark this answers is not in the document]
ktinubu: Fixed in a07c8a5. The two `render_params(plugin, ...)` call sites in `scripts/predict_vst_audio.py` (lines 171 and 190) now pass `plugin_path` instead, with `preset_path=preset_path` so the per-call reload inside `render_params` also applies the preset. Dropped the now-redundant upfront `load_plugin` / `load_preset` block in `main` since `render_params` reloads the plugin internally on every call.

Note that `tests/data/vst/test_preset_params.py` was already updated to the new signature in this PR — only `predict_vst_audio.py` was lagging.

scripts/surge_xt_interactive.py
```diff
@@ -1,40 +1,477 @@
...
+    :param param_spec_name: Parameter spec name (key into ``param_specs``) used
+        to decode the rescaled row into raw VST parameter values.
+    :returns: Dict mapping VST parameter name to its raw (decoded) value.
+    :raises IndexError: when ``batch_idx`` is out of range for ``pred_tensor``.
+    """
+    if batch_idx < 0 or batch_idx >= pred_tensor.shape[0]:
+        raise IndexError(f"batch_idx {batch_idx} out of range (batch size {pred_tensor.shape[0]})")
+
+    spec = param_specs[param_spec_name]
+    row = pred_tensor[batch_idx].float().numpy()
```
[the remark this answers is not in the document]
ktinubu: Fixed in a07c8a5: `decode_prediction_row` now uses `pred_tensor[batch_idx].detach().cpu().float().numpy()`. Current loader (`_load_pred_tensor`) already pins to CPU and Lightning predict-step output has no grad in practice, but the docstring frames this as a general decode utility — defensive narrowing is cheap insurance.

scripts/surge_xt_interactive.py
```diff
@@ -1,40 +1,477 @@
...
+        "q": quit_action,
+        "p": record_patch,
+    }
+    # NOTE: ``click.getchar()`` blocks until a key is pressed, so this loop only
+    # checks ``stop_event`` between keystrokes. If the editor closes from another
+    # thread, the loop won't exit until the user presses any key. Acceptable for
+    # the typical editor-close-then-press-q flow; a sentinel-key approach would
+    # be needed for fully event-driven shutdown.
+    while not stop_event.is_set():
+        ch = click.getchar()
```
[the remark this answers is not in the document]
ktinubu: Already addressed in 410e75b — same concern as thread #3161493662 (line 319). The NOTE block at `scripts/surge_xt_interactive.py:309-313` documents that `click.getchar()` blocks between iterations and `stop_event` only takes effect on the next keystroke. Acceptable for the typical editor-close-then-press-q flow; the full event-driven shutdown (sentinel keypress / `select` on stdin) is a larger refactor I can file as a separate issue if desired.

scripts/surge_xt_interactive.py
```diff
@@ -1,40 +1,477 @@
...
+        synth_params = load_dataset_synth_params(dataset_ref, param_spec_name)
+        set_params(plugin, synth_params)
+    elif pred is not None:
+        synth_params = load_prediction_synth_params(pred, param_spec_name)
+        set_params(plugin, synth_params)
+
+    stop_event = threading.Event()
+    with ThreadPoolExecutor() as pool:
+        audio_future = pool.submit(play_audio, plugin, stop_event)
+        keyboard_future = pool.submit(keyboard_loop, plugin, stop_event, param_spec_name)
```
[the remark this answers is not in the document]
ktinubu: Technically correct, but tied to the same shutdown pathway as #3161701461: `with ThreadPoolExecutor() as pool:` only blocks indefinitely because `keyboard_loop` is itself blocked in `click.getchar()` and never observes `stop_event`. `play_audio` exits cleanly on the event. Tightening just the executor (`shutdown(wait=False, cancel_futures=True)`) would paper over the symptom without fixing the underlying block in `keyboard_loop`, so deferring alongside the full event-driven shutdown — captured by the NOTE at lines 309-313 in 410e75b. Happy to file a follow-up issue for the combined fix if you want it tracked.

src/data/vst/generate_vst_dataset.py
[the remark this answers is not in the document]
ktinubu: Acknowledged — real bug, but deferred to a follow-up issue rather than fixed in this PR. Tracked as #704 (Bug, evaluation, parented under Phase 6: Synth Preview Tools).

Latent today because `surge_xt_interactive.py` only passes `fixed_synth_params_list` (so the `else` branch keeps generating fresh `sampled_note` per iteration) and the e2e replay test at `tests/data/vst/test_generate_vst_dataset.py` pre-filters candidates via stage 1 to guarantee they pass the loudness gate. The bounded-retries fix described on the issue (`_MAX_LOUDNESS_RETRIES` + clear `RuntimeError`) will land there.

This PR (a07c8a5) does simplify the surrounding `if/else` branching in `generate_sample` (the first branch was subsumed by the second) but intentionally leaves the loop semantics unchanged.

src/data/vst/generate_vst_dataset.py
```diff
@@ -60,23 +60,29 @@ def make_spectrogram(audio: np.ndarray, sample_rate: float) -> np.ndarray:
 
 
 def generate_sample(
-    plugin: VST3Plugin,
+    plugin_path: str,
     velocity: int,
     signal_duration_seconds: float,
     sample_rate: float,
```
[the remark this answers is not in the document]
ktinubu: Fixed in a07c8a5: dropped `from pedalboard import VST3Plugin` from `src/data/vst/generate_vst_dataset.py`. Confirmed unused after the path-based refactor.

src/data/vst/core.py
```diff
@@ -62,16 +64,28 @@ def write_wav(audio: np.ndarray, path: str, sample_rate: float, channels: int) -
...
+    The plugin is reloaded (and its editor briefly shown via
+    ``_prepare_plugin``) on every call. This is intentional and is the
+    workaround for the silent / alternate-render bug investigated on this
+    branch (see commits 086d80f, 9ff7f16, c6e4139): without a per-call reload,
+    the plugin retained stale state from prior renders and produced silent or
+    repeated audio. The cost is ~7s of plugin-load + editor-show overhead per
+    sample — acceptable for interactive use and small batch generation, not
+    for large-scale pipeline runs.
+    """
+    plugin = load_plugin(plugin_path)
```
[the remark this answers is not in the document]
ktinubu: render_params was changed to unconditionally reload the plugin due to a correctness issue in the render that occurs when we don't.  I'll add a comment and a github issue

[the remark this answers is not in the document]
ktinubu: Filed as P0 bug #705 — tracks the underlying stale-state issue (silent / repeated renders without per-call reload) plus the perf-regression from the current mitigation, with a sketch of three approaches for a perf-friendly fix (including the opt-in flag you suggested). The PR ships the safe-but-slow workaround so dataset generation produces correct audio today; #705 is where the per-call reload gets revisited.

Also added a docstring pointer in `src/data/vst/core.py::render_params` to issue #705 in 7453d18 so the next reader to hit this code finds the tracking issue immediately.

scripts/surge_xt_interactive.py
```diff
@@ -1,40 +1,476 @@
...
+    :param output_path: WAV destination. Parent directory must exist.
+    :param duration_seconds: Seconds of audio to render. The actual length is
+        rounded up to the nearest ``BUFFER_SIZE`` samples.
+    """
     silence = np.zeros((CHANNELS, BUFFER_SIZE), dtype=np.float32)
+    num_buffers = math.ceil(duration_seconds * SAMPLE_RATE / BUFFER_SIZE)
+    with AudioFile(str(output_path), "w", SAMPLE_RATE, CHANNELS) as f:
+        for _ in range(num_buffers):
+            synth_output = plugin(silence, SAMPLE_RATE, reset=False)
+            f.write(synth_output)
```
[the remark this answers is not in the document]
ktinubu: Fixed in 7453d18 — but rather than fix the standalone `render_to_wav`, I folded the WAV-recording path into `play_audio` so there is one place that consumes plugin output (and one place that does the `(channels, frames) -> (frames, channels)` transpose).

When `--session-recording-path` is set, `play_audio` tees the first `SESSION_RECORDING_MAX_SECONDS` (20s) of the live stream to that WAV file using `synth_output.T` (matching the convention `src.data.vst.core.write_wav` already uses). When the flag is unset, the recording path is a no-op — playback works as before.

Also added an `assert synth_output.shape == (CHANNELS, BUFFER_SIZE)` at the top of the loop so a pedalboard API change (or a misconfigured plugin) fails loudly instead of silently corrupting both the WAV write and the stream write. The standalone `render_to_wav` and its two tests were removed — recording now lives inside `play_audio`.

scripts/surge_xt_interactive.py
```diff
@@ -1,40 +1,476 @@
...
+    it differs from ``SAMPLE_RATE`` so the output device gets a rate it supports.
+    """
+    silence = np.zeros((CHANNELS, BUFFER_SIZE), dtype=np.float32)
+    needs_resample = SAMPLE_RATE != PLAYBACK_SAMPLE_RATE
+    stream_resampler = (
+        StreamResampler(SAMPLE_RATE, PLAYBACK_SAMPLE_RATE, CHANNELS) if needs_resample else None
+    )
     with AudioStream(
         output_device_name=AudioStream.default_output_device_name,
         sample_rate=SAMPLE_RATE,
```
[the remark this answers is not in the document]
ktinubu: Fixed in 7453d18: `AudioStream` is now opened with `sample_rate=PLAYBACK_SAMPLE_RATE`, matching the rate the writes target. Plugin processing still runs at `SAMPLE_RATE`; the `StreamResampler` bridges between them. So if `PLAYBACK_SAMPLE_RATE` is later changed to match the output device, the device gets opened at the right rate and writes go through unchanged.

scripts/surge_xt_interactive.py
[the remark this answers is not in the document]
ktinubu: Refactored in f83ed06: split into two functions instead of weaving the recorder into `play_audio`.

- `play_audio` reverts to its pre-`7453d18` live-stream shape (just `AudioStream` + plugin loop), keeping the `sample_rate=PLAYBACK_SAMPLE_RATE` fix from #3162440944.
- New `play_audio_recorded` follows the canonical pedalboard recording pattern — `with AudioFile(...) as f:`, loop while `written < target and not stop_event.is_set()`, trim the final chunk so the file lands on exactly `n_seconds * SAMPLE_RATE` frames. The transpose `(channels, frames) → (frames, channels)` happens once at the single `f.write` call, matching the `core.write_wav` convention. No `AudioStream`, so it also works headless.
- `main` dispatches between them based on `--session-recording-path`.

Added unit tests covering the trim-final-chunk math and stop-event early-exit. Also refreshed `docs/guides/surge-xt-interactive.md` (CLI table, quick-start example, audio-thread description, and a troubleshooting entry that previously pointed at the now-deleted `render_to_wav` as the headless-render escape hatch).

Issue #730, opened 2026-04-30
fix(testing): 7 docker_entrypoint error-path tests crash in CliRunner under log_cli=True


ktinubu:
## Symptom

7 tests in `tests/test_docker_entrypoint.py` fail with:

\`\`\`
File "uvenv/lib/python3.10/site-packages/click/testing.py", line 531, in invoke
    stdout = outstreams[0].getvalue()
ValueError: I/O operation on closed file.
\`\`\`

The commands under test work correctly — exit code is 1, stderr message is captured, log line is captured. The crash is in `CliRunner.invoke`'s `finally` block when it reads the captured stdout.

## Affected tests

All in `tests/test_docker_entrypoint.py`:

1. \`TestIdle::test_idle_exec_failure_becomes_click_exception\`
2. \`TestPassthrough::test_passthrough_exec_failure_becomes_click_exception\`
3. \`TestGenerateDataset::test_malformed_json_spec_exits_nonzero_without_calling_run\`
4. \`TestGenerateDataset::test_invalid_spec_shape_exits_nonzero_without_calling_run\`
5. \`TestGenerateDataset::test_binary_spec_file_exits_nonzero_without_calling_run\`
6. \`TestRenderEval::test_render_eval_fails_loudly_with_issue_pointer\`
7. \`TestTrain::test_train_fails_loudly_with_issue_pointer\`

## Root cause (verified empirically)

Running with \`--override-ini="log_cli=False"\` makes all 18 tests pass; running with \`log_cli=True\` (the project default in \`pyproject.toml\` \`[tool.pytest.ini_options]\`) reproduces the 7 failures consistently.

All 7 failing tests are error-path tests whose click callback calls \`logger.error(...)\` before raising \`ClickException\`. The 11 passing tests either don't take the error branch or take an error branch that doesn't log (e.g. click's \`UsageError\`).

Mechanism:

1. \`log_cli = "True"\` installs \`_pytest.logging._LiveLoggingStreamHandler\` on the root logger.
2. When the callback under test calls \`logger.error(...)\`, the handler's \`emit()\` opens a \`global_and_fixture_disabled\` context that calls \`suspend_global_capture()\`.
3. That close path closes the \`BytesIOCopy\` streams that \`CliRunner.isolation()\` wrote into \`sys.stdout\`/\`sys.stderr\` (click 8.2+'s \`StreamMixer\` design).
[...]

PR #731, opened 2026-04-30
test(testing): xfail 7 docker_entrypoint error-path tests blocked by log_cli bug

ktinubu:
## Summary

Mark 7 error-path tests in `tests/test_docker_entrypoint.py` as `@pytest.mark.xfail(strict=True, raises=ValueError)` so CI is unblocked on branches that touch the test suite. The underlying click 8.3 + `log_cli=True` interaction is tracked in #730 and will be fixed in a follow-up PR (which will retire these markers automatically via `strict=True`).

## Why xfail and not skip / not delete

- The commands themselves work correctly — exit code, stderr, captured log are all right. Only `CliRunner.invoke()`'s finally-block is broken under `log_cli=True`.
- `strict=True` + `raises=ValueError` means: when the real fix lands, pytest fails with `XPASS(strict)` and forces the markers to be removed. No silent zombies.

Refs #730

## Test plan

- [x] Reproduce baseline failure on `main`:

  ```
  $ uvenv/bin/python -m pytest tests/test_docker_entrypoint.py -p no:xdist
  ...
  FAILED tests/test_docker_entrypoint.py::TestIdle::test_idle_exec_failure_becomes_click_exception - ValueError: I/O operation on closed file.
  FAILED tests/test_docker_entrypoint.py::TestPassthrough::test_passthrough_exec_failure_becomes_click_exception - ValueError: I/O operation on closed file.
  FAILED tests/test_docker_entrypoint.py::TestGenerateDataset::test_malformed_json_spec_exits_nonzero_without_calling_run - ValueError: I/O operation on closed file.
  FAILED tests/test_docker_entrypoint.py::TestGenerateDataset::test_invalid_spec_shape_exits_nonzero_without_calling_run - ValueError: I/O operation on closed file.
  FAILED tests/test_docker_entrypoint.py::TestGenerateDataset::test_binary_spec_file_exits_nonzero_without_calling_run - ValueError: I/O operation on closed file.
  FAILED tests/test_docker_entrypoint.py::TestRenderEval::test_render_eval_fails_loudly_with_issue_pointer - ValueError: I/O operation on closed file.
  FAILED tests/test_docker_entrypoint.py::TestTrain::test_train_fails_loudly_with_issue_pointer - ValueError: I/O operation on closed file.
  ========================= 7 failed, 11 passed in 0.52s =========================
  ```

- [x] After applying markers — `tests/test_docker_entrypoint.py` reports 11 passed, 7 xfailed, 0 xpassed:

  ```
  $ uvenv/bin/python -m pytest tests/test_docker_entrypoint.py -p no:xdist
  ======================== 11 passed, 7 xfailed in 0.46s =========================
  ```

- [x] Full non-slow suite — no regressions:

  ```
  $ uvenv/bin/python -m pytest -m 'not slow' -q
  ==== 223 passed, 5 skipped, 97 deselected, 7 xfailed, 2 warnings in 26.79s =====
  ```

- [x] `pre-commit run -a` clean (ruff, ruff format, pyright, etc. all green).

### DIFF
--- a/tests/test_docker_entrypoint.py
+++ b/tests/test_docker_entrypoint.py
@@ -28,6 +28,20 @@ from pydantic import BaseModel
 REPO_ROOT = Path(__file__).resolve().parent.parent
 ENTRYPOINT_PATH = REPO_ROOT / "scripts" / "docker_entrypoint.py"
 
+# Tests that drive a click callback whose error path calls logger.error(...) crash inside
+# CliRunner.invoke()'s finally-block (.getvalue() on a closed BytesIOCopy) when pytest's
+# log_cli=True is set: _LiveLoggingStreamHandler.emit() suspends global capture mid-emit and
+# closes the BytesIOCopy that CliRunner.isolation() installed. Strict + raises=ValueError so
+# the markers are removed automatically by the real fix in #730.
+_XFAIL_CLI_RUNNER_LOG_CLI_BUG = pytest.mark.xfail(
+    reason=(
+        "CliRunner.invoke() finally-block .getvalue() on closed BytesIOCopy under "
+        "log_cli=True; tracked in #730."
+    ),
+    strict=True,
+    raises=ValueError,
+)
+
 
 def _load_entrypoint_module() -> ModuleType:
     """Import scripts/docker_entrypoint.py as a module for in-process testing."""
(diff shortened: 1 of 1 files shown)


ktinubu:
Closing — premise was wrong. The bug also reproduces inside the Docker build job (run 25184695371, job 73838810585) which has the same `click==8.1.8` + `pytest==9.0.3` as Ubuntu/macOS CI, where the tests pass. So this is not a click-8.2+ specific issue; some other environment factor (likely xdist, capture mode, or tty) decides whether the bug triggers.

The strict xfail approach is wrong shape because it breaks the Ubuntu/macOS CI legs where the tests pass naturally (XPASS strict). Going directly to the autouse-fixture fix instead.

Issue #534, opened 2026-04-09
docs(pipeline): SkyPilot compute integration design doc


ktinubu:
Design document proposing SkyPilot as the compute backend instead of the handrolled `ComputeBackend` protocol with `RunPodBackend`.

**Motivation:**
- Multi-provider flexibility (Vast.ai, RunPod, Lambda) without writing each backend
- RunPod sshd-in-Docker friction eliminated
- SkyPilot managed jobs provide spot recovery using existing R2 markers as natural checkpoints

**Scope:**
- `compute_config` field across DatasetConfig, train, and eval configs
- SkyPilot managed jobs for data generation pipeline
- Design doc only — no implementation changes

Refs #105
Refs #106

ktinubu:
# Add SkyPilot launcher for RunPod smoke data-generation

## Context

The `docs/design/skypilot-compute-integration.md` design doc has been agreed on internally — SkyPilot will replace the planned bespoke RunPod backend, and the pipeline will gain a `compute_config` field that, when set, drives a `sky.jobs.launch(...)` path. That full integration is multi-PR (schema changes, `pipeline/__main__.py`, worker UUIDs, etc.).

This PR is the **first probe**: a self-contained scaffold that proves SkyPilot can pick up our existing Docker image, mount a materialized `DatasetPipelineSpec`, and run the existing `scripts/docker_entrypoint.py generate_dataset --spec ...` flow on RunPod against the smoke config. **No schema changes, no `compute_config` field, no `LocalBackend` plumbing** — those land in the Phase A–C PRs that follow.

Why now: the design doc has been sitting in draft since 2026-04-13; standing up an end-to-end smoke run de-risks the rest of the work and surfaces image-tag, credential-forwarding, and SkyPilot-API issues before they're entangled with schema migrations.

Tracking: `Refs #534`. Commit prefix: `internal-feat:` (new tested code that doesn't yet expose a user-facing capability — there is no `pipeline generate ... --backend skypilot` yet).

## Execution order

0. **Post this plan to `#534` first.** Before any code changes, paste the body of this plan as a comment on the SkyPilot tracking issue — the design doc names #534 as the umbrella, and this PR is the first scaffolding under it. `gh issue comment 534 --repo tinaudio/synth-setter --body-file <plan>`. (The current sandbox has unauthenticated `gh`; if it errors, prompt the user to `gh auth login` or paste the plan manually before proceeding.)
1. Worktree setup (below).
[...]

PR #745, opened 2026-05-01
ci(test-skypilot-debug): add image-pull variant for second pod-boot probe

ktinubu:
## Summary

Adds a 4th variant (`image-pull`) to the `test-skypilot-debug` matrix that just provisions a RunPod pod, SkyPilot pulls the `dev-snapshot` image, the container starts, and the run block prints a single `done` line. Same shape as `noop` — but a distinct matrix slot so a single RunPod transient on `noop` doesn't look like a regression.

`noop` + `image-pull` together = two independent pod-boot probes per dispatch on the simplest possible workload. Useful concretely for triaging `test-dataset-generation` smoke failures with the shape `RuntimeError: Failed to SSH to <pod-ip> after timeout 600s, with Error: ConnectionRefusedError` (we hit several today): if both `noop` and `image-pull` are green when production is failing, the RunPod side is healthy and the failure is downstream of pod boot.

Files:

- `configs/compute/runpod-debug-image-pull-template.yaml` (new) — same `resources:` and `image_id` as the other debug templates; minimal `setup:` + `run:`.
- `.github/workflows/test-skypilot-debug.yml` — 4-variant matrix (was 3); docstring updated; "spends ~3 pods" → "~4 pods".

`Refs #534`.

## Test plan

- [ ] **Dispatch the matrix on this branch and confirm `image-pull` runs alongside the existing 3 variants** — Level 1, required by the `/pr-checkbox` skill for `workflow_dispatch` workflows. Will dispatch and link the green run here. Pod budget: ~4 RunPod pods per dispatch (vs ~3 today).
- [x] **YAML parses, matrix variants list resolves to `[noop, image-pull, rclone, pedalboard-load]`**:
  ```bash
  $ python3 -c "import yaml; y=yaml.safe_load(open('.github/workflows/test-skypilot-debug.yml')); print([m['variant'] for m in y['jobs']['debug']['strategy']['matrix']['include']])"
  ['noop', 'image-pull', 'rclone', 'pedalboard-load']
  ```

### DIFF
--- a/.github/workflows/test-skypilot-debug.yml
+++ b/.github/workflows/test-skypilot-debug.yml
@@ -1,6 +1,6 @@
 name: Test SkyPilot Debug
 
-# Permanent diagnostic matrix for the SkyPilot+RunPod path. Three canary
+# Permanent diagnostic matrix for the SkyPilot+RunPod path. Four canary
 # variants, each isolating one known failure class. None of these exercise
 # the production launcher or worker — they're standalone probes. When
 # test-dataset-generation (lands in a follow-up PR) regresses, dispatch
@@ -12,6 +12,15 @@ name: Test SkyPilot Debug
 #                    itself is fine; failure here means the regression
 #                    is in SkyPilot, RunPod's API, or the cluster
 #                    plumbing — not in our code.
+#
+#   image-pull       Image-availability canary. Same shape as `noop`
+#                    (provision a pod, run a one-line echo) but a
+#                    distinct slot in the matrix so a single RunPod
+#                    transient on `noop` doesn't look like a regression.
+#                    `noop` + `image-pull` together = two independent
+#                    pod boots per dispatch on the simplest possible
+#                    workload; both green = RunPod can reliably hand us
+#                    pods that have SSH + the dev-snapshot image ready.
 #
 #   rclone           rclone-process-exit canary
 #                    (`rclone copy <small file> r2:...`). Catches a
@@ -32,7 +41,7 @@ name: Test SkyPilot Debug
 #                    `os._exit(0)` workaround for.
 #
 # Triggers: workflow_dispatch only — RunPod pods are billable. Each
-# dispatch spends ~3 pods. Add a `push` trigger only temporarily when
+# dispatch spends ~4 pods. Add a `push` trigger only temporarily when
 # actively iterating on launcher / wrapper / template behavior.
 #
 # Required secrets:
@@ -72,6 +81,9 @@ jobs:
           - variant: noop
             template: configs/compute/runpod-debug-template.yaml
             cluster_suffix: noop
+          - variant: image-pull
+            template: configs/compute/runpod-debug-image-pull-template.yaml
+            cluster_suffix: image-pull
           - variant: rclone
             template: configs/compute/runpod-debug-rclone-template.yaml
             cluster_suffix: rclone
--- a/configs/compute/runpod-debug-image-pull-template.yaml
+++ b/configs/compute/runpod-debug-image-pull-template.yaml
@@ -1,0 +1,33 @@
+# SkyPilot task template for RunPod — DEBUG variant: image-pull canary.
+#
+# The minimum-meaningful workload: a pod is provisioned with the
+# dev-snapshot image, SkyPilot pulls the image, the container starts,
+# the run block prints a single "done" line. PASS = RunPod can hand us a
+# pod that has SSH + the image ready inside the SkyPilot SSH-readiness
+# timeout. FAIL = pod never came up to SSH-readiness (the failure mode we
+# saw on the test-dataset-generation smoke when production was failing
+# while the matrix's `noop` variant was green — this variant is a second
+# RunPod-orchestration probe per dispatch so a single transient doesn't
+# look like a regression).
+
+resources:
+  cloud: runpod
+  accelerators:
+    {
+      RTX3070:1,
+      RTX3080:1,
+      RTX3090:1,
+      RTXA4000:1,
+      RTX4000Ada:1,
+      A40:1,
+      RTX4090:1,
+    }
+  use_spot: false
+  disk_size: 50
+  image_id: docker:tinaudio/synth-setter:dev-snapshot
+
+setup: |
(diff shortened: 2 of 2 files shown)


ktinubu:
## Verification Results

3 checks done; the Level 1 matrix dispatch is in flight.

- [ ] **4-variant matrix runs end-to-end on RunPod with all variants green** *(Level 1 — REQUIRED by the skill for `workflow_dispatch` workflows. Dispatched as [run 25237923437](https://github.com/tinaudio/synth-setter/actions/runs/25237923437); IN PROGRESS at the time of this comment. Will update with the green/red on each of the 4 variants when it completes.)*
  Particularly watching for `noop` + `image-pull` to BOTH go green — that's the new probe coverage this PR adds. If only one of them is green and the other times out at the SSH-readiness check, the asymmetry is real signal that the failure on `test-dataset-generation` ([#743](https://github.com/tinaudio/synth-setter/pull/743)) is a flakiness pattern, not a deterministic regression.

- [x] **SkyPilot's `Task.from_yaml` accepts the new template** *(Level 2 — invoked the actual SkyPilot parser the matrix workflow calls; promise is "this YAML is a valid SkyPilot Task, not just valid YAML")*
  ```bash
  $ cd configs/compute && python3 -c "
  import sky
  task = sky.Task.from_yaml('runpod-debug-image-pull-template.yaml')
  print(f'resources count: {len(task.resources)}')
  print(f'run preview: {repr((task.run or \"\")[:120])}')
  print('PASS')"
  resources count: 7
  run preview: 'echo "skypilot-debug image-pull job done (host: $(hostname), image online)"\n'
  PASS
  ```

- [x] **Matrix has 4 variants with unique cluster suffixes; all 4 templates exist on disk** *(Level 5 — yaml.safe_load + os.path.isfile; promise is "the matrix won't fail at GH-Actions matrix expansion time before any pod is provisioned")*
  ```bash
  $ python3 -c "
  import os, yaml
  y = yaml.safe_load(open('.github/workflows/test-skypilot-debug.yml'))
  variants = [m['variant'] for m in y['jobs']['debug']['strategy']['matrix']['include']]
  suffixes = [m['cluster_suffix'] for m in y['jobs']['debug']['strategy']['matrix']['include']]
  templates = [m['template'] for m in y['jobs']['debug']['strategy']['matrix']['include']]
  assert len(set(suffixes)) == len(suffixes)
  for t in templates: assert os.path.isfile(t)
  print(variants)
  print(suffixes)
  print('PASS')"
  ['noop', 'image-pull', 'rclone', 'pedalboard-load']
  ['noop', 'image-pull', 'rclone', 'pedalboard']
  PASS
  ```

### Adversarial

- [x] **The new variant runs in parallel with the existing 3, not in series** *(Level 5 — the `strategy: matrix` block has no `max-parallel` cap and `fail-fast: false` is set, so the GH-Actions matrix-expansion behavior fans out 4 jobs independently. Promise is "adding the new variant doesn't slow the dispatch wall-clock from ~5–10 min to ~20–40 min".)*
  ```bash
  $ python3 -c "
  import yaml
  y = yaml.safe_load(open('.github/workflows/test-skypilot-debug.yml'))
  strategy = y['jobs']['debug']['strategy']
  print(f'max-parallel: {strategy.get(\"max-parallel\", \"unset (= unlimited)\")}')
  print(f'fail-fast: {strategy[\"fail-fast\"]}')"
  max-parallel: unset (= unlimited)
  fail-fast: false
  ```

- [x] **`image-pull` template is structurally identical to `noop` in resources / image_id (so they probe the same RunPod path)** *(Level 6 — diff the two templates' `resources:` blocks)*
  ```bash
  $ diff <(python3 -c "import yaml; print(yaml.safe_dump(yaml.safe_load(open('configs/compute/runpod-debug-template.yaml'))['resources']))") \
         <(python3 -c "import yaml; print(yaml.safe_dump(yaml.safe_load(open('configs/compute/runpod-debug-image-pull-template.yaml'))['resources']))")
  (no output — resources blocks are byte-identical)
  ```
  This is what makes `image-pull` a redundant-but-useful probe: same RunPod scheduling decisions, same image, same pool of GPU types — only the cluster name differs (which makes them independent slots in SkyPilot's perspective).

### Summary

| Promise | Level | Status |
|---|---|---|
| 4-variant matrix dispatched green | 1 | IN PROGRESS — will update |
| SkyPilot accepts new template | 2 | PASS |
| Matrix shape sane (4 variants, unique suffixes, templates on disk) | 5 | PASS |
| Variants run in parallel | 5 | PASS |
| `image-pull` resources match `noop` resources | 6 | PASS |

ktinubu:
## Matrix dispatch result — 4/4 GREEN

[Run 25237923437](https://github.com/tinaudio/synth-setter/actions/runs/25237923437) on this branch:

| Variant | Result |
|---|---|
| `noop` | ✅ PASS |
| `image-pull` (new) | ✅ PASS |
| `rclone` | ✅ PASS |
| `pedalboard-load` | ✅ PASS |

The new variant runs alongside the existing 3 in parallel; total wall-clock ~10 min (same as before — `fail-fast: false` + no `max-parallel` cap means `image-pull` runs concurrent with `noop` and doesn't add to the critical path).

### Bonus signal

This dispatch ran at 23:41 UTC, ~25 min after `test-dataset-generation` ([#743](https://github.com/tinaudio/synth-setter/pull/743)) had its 3rd consecutive RunPod-transient failure (23:17 UTC, "Failed to SSH to <pod-ip> after timeout 600s"). Both `noop` and `image-pull` succeeded here on the same RunPod backend, on the same `dev-snapshot` image, with identical resources blocks — strong evidence the production-smoke failures were purely RunPod-side flakiness, not a deterministic regression.

The pair `noop` + `image-pull` going green together is now the canary signal that the platform side is healthy and any concurrent production-smoke failure is downstream of pod boot (rclone, the launcher, generate_dataset, the wrapper, etc.).

### Verification — final

| Promise | Level | Status |
|---|---|---|
| 4-variant matrix dispatched green | 1 | **PASS** ✅ |
| SkyPilot accepts new template | 2 | PASS |
| Matrix shape sane | 5 | PASS |
| Variants run in parallel | 5 | PASS |
| `image-pull` resources match `noop` resources | 6 | PASS |

Issue #787, opened 2026-05-04
ci(ci-automation): run test-gpu.yml inside dev-snapshot Docker image instead of bare-metal install


ktinubu:
## Motivation

`.github/workflows/test-gpu.yml` currently installs Python, uv, the torch stack (with a `torch<2.7.0` constraint to match driver 12.8), the Surge XT VST3 `.deb`, and the headless X11 stack on the bare-metal `gpu-x64` runner before each run. This duplicates ~80% of the dev-snapshot Dockerfile and means the CI environment can drift from what production workers (RunPod/OCI via SkyPilot) actually use.

The image `tinaudio/synth-setter:dev-snapshot` already ships:
- torch built against `cu128` (matches the gpu-x64 driver 12.8)
- Surge XT 1.3.4 at `/usr/lib/vst3/Surge XT.vst3`
- Headless X11 / dbus / openbox / mesa stack
- `scripts/run-linux-vst-headless.sh` validated at build time

`test-vst-slow.yml` already runs its slow VST suite this way; this issue extends the same pattern to the GPU suite.

## Scope

- Replace the bare-metal install steps in `.github/workflows/test-gpu.yml` with `docker pull` + `docker run --gpus all` against `tinaudio/synth-setter:dev-snapshot`.
- Mount the workspace into `/home/build/synth-setter`, set `SYNTH_SETTER_PLUGIN_PATH=/usr/lib/vst3/Surge XT.vst3`, wrap pytest in `run-linux-vst-headless.sh`.
- Add a `workflow_dispatch` `image_tag` input so a maintainer can pin a `dev-snapshot-<sha>` tag.
- Add a GPU-visibility smoke step before the longer pytest run.

## Out of scope

- Cost / scheduling changes (tracked separately in #334).
- `cpu-slow.yml` migration to Docker.

## Acceptance

- `test-gpu.yml` no longer references `setup-python`, `setup-uv`, `apt-get install xvfb`, or the Surge XT `.deb` URL.
- Scheduled and manual dispatches both succeed against the current `dev-snapshot` tag.

ktinubu:
Closing as complete. Verified by multi-agent audit pass.

## What was asked

[...]

Issue #334, opened 2026-03-27
ci(gpu): move GPU tests to twice-weekly schedule


ktinubu:
GPU tests on the gpu-x64 runner currently trigger on every push to main, costing ~$39 over 755 minutes. Move to a twice-weekly cron schedule (Mon + Thu 06:00 UTC) to reduce cost while still catching regressions.

ktinubu:
Closing — delivered in PR #335. `.github/workflows/test-gpu.yml` now uses `on.schedule = "0 6 * * 1,4"` (Mon+Thu 06:00 UTC).

PR #789, opened 2026-05-04
ci(ci-automation): run test-gpu.yml inside dev-snapshot Docker image

ktinubu:
## Summary

- Switches `.github/workflows/test-gpu.yml` from a bare-metal install (Python + uv + `torch<2.7.0` constraint + Surge XT `.deb` + apt headless X11 stack) to `docker pull` + `docker run --gpus all` against `tinaudio/synth-setter:dev-snapshot`, mirroring the proven pattern in `test-vst-slow.yml`.
- Drops ~80% of the workflow's setup steps — the dev-snapshot image already ships cu128 torch (matches the `gpu-x64` runner driver), Surge XT 1.3.4 at `/usr/lib/vst3/Surge XT.vst3`, and the headless X11/dbus stack.
- Adds a `workflow_dispatch` `image_tag` input (defaults to `dev-snapshot`) so a maintainer can pin a `dev-snapshot-<sha>` tag, plus a GPU-visibility smoke step (`nvidia-smi` + `torch.cuda.is_available()` inside the container) before the longer pytest run so toolkit/driver issues surface fast.
- Workspace is bind-mounted into `/home/build/synth-setter`; `SYNTH_SETTER_PLUGIN_PATH=/usr/lib/vst3/Surge XT.vst3` works around the in-image symlink getting shadowed by the mount; pytest is wrapped in `scripts/run-linux-vst-headless.sh` to match the existing setup.

Refs #787

## Test plan

- [ ] Manual `workflow_dispatch` of `GPU Tests` on this branch passes against the current `dev-snapshot` tag.
- [ ] GPU-visibility smoke step prints a non-empty `nvidia-smi` row and `cuda: True`.
- [ ] Surge XT plugin-load smoke step succeeds (proves the bind-mount + headless wrapper still work end-to-end).
- [ ] `pytest -vv -s -m gpu` finishes within the 120-minute timeout and is green.
- [ ] `Validate GitHub Workflows` pre-commit hook passes locally (already verified).

## Notes for reviewers

- Assumes `gpu-x64` has `nvidia-container-toolkit` so `--gpus all` works. If it does not, the GPU-visibility smoke step will fail fast and we'll roll back; this is intentional.
- Out of scope: `cpu-slow.yml` migration to Docker, GPU test cost/scheduling (tracked in #334).

### DIFF
--- a/.github/workflows/test-gpu.yml
+++ b/.github/workflows/test-gpu.yml
@@ -1,9 +1,17 @@
 name: GPU Tests
 
-# Runs only @pytest.mark.gpu tests on a GPU runner.
-# Twice-weekly schedule (Mon + Thu) and manual dispatch.
+# Runs only @pytest.mark.gpu tests inside the project's
+# `tinaudio/synth-setter:dev-snapshot` Docker image so the torch (cu128)
+# build, Surge XT VST3, and headless X11 stack match production. Mirrors
+# the Docker-in-CI pattern used by `test-vst-slow.yml`. Twice-weekly
+# schedule (Mon + Thu) and manual dispatch.
 on:
   workflow_dispatch:
+    inputs:
+      image_tag:
+        description: "Docker image tag (e.g. dev-snapshot, dev-snapshot-<sha>)"
+        required: false
+        default: "dev-snapshot"
   schedule:
     - cron: "0 6 * * 1,4" # Mon + Thu 06:00 UTC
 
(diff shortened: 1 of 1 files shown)


ktinubu:
## Verification Results

Behavioral checks for the dev-snapshot Docker switch. The Level 1 workflow_dispatch run is left unticked: triggering the `gpu-x64` runner consumes paid GPU minutes (120-min timeout, ~$5/run at the documented rate) and `CLAUDE.md` requires explicit maintainer approval for that class of action. Run the dispatch yourself when ready (instructions below).

- [x] **Workflow file is a structurally valid GitHub Actions workflow** *(Level 2 — invokes `check-jsonschema` against the GHA schema via the project's pre-commit hook; promise is GHA-spec validity, not end-to-end execution)*
  ```bash
  $ pre-commit run --files .github/workflows/test-gpu.yml 2>&1 | grep -E '(Validate GitHub Workflows|check yaml|prettier)'
  check yaml...............................................................Passed
  prettier.................................................................Passed
  ```

- [x] **All four bash `run:` blocks have valid syntax** *(Level 2 — invoked `bash -n` directly on each extracted run block; promise is shell-syntax correctness, which is otherwise only caught when GHA spawns the runner)*
  ```bash
  $ python3 -c "
  import yaml
  wf = yaml.safe_load(open('.github/workflows/test-gpu.yml'))
  for i, s in enumerate(wf['jobs']['run_tests']['steps']):
      if 'run' in s:
          open(f'/tmp/step_{i}.sh','w').write('#!/bin/bash\nset -euo pipefail\n'+s['run'])
          print(i, s.get('name'))
  "
  1 Pull image
  2 Smoke-test GPU visibility inside container
  3 Smoke-test Surge XT plugin load
  4 Run GPU tests in Docker
  $ for f in /tmp/step_*.sh; do bash -n "$f" && echo "OK: $f"; done
  OK: /tmp/step_1.sh
  OK: /tmp/step_2.sh
  OK: /tmp/step_3.sh
  OK: /tmp/step_4.sh
  ```

- [x] **The `dev-snapshot` tag this workflow defaults to actually exists on Docker Hub** *(Level 4 — queried the live Docker Hub registry API; failure mode this catches: a workflow that pulls a tag that has never been pushed will fail at the very first step regardless of YAML validity)*
  ```bash
  $ curl -sf "https://hub.docker.com/v2/repositories/tinaudio/synth-setter/tags/dev-snapshot/" | python3 -c "import sys,json; d=json.load(sys.stdin); print('name:',d['name']); print('last_pushed:',d['tag_last_pushed']); print('digest:',d['digest'][:25],'...'); print('size_gb:',round(d['full_size']/1e9,2))"
  name: dev-snapshot
  last_pushed: 2026-05-03T07:59:31.870575Z
  digest: sha256:18101918a2d79d31e8a ...
  size_gb: 5.38
  ```

- [x] **Bind-mount path, plugin path, and PYTHONPATH all match what the dev-snapshot image was built with** *(Level 5 — cross-checked the workflow's hardcoded paths against the Dockerfile that produces the image; failure mode this catches: a path drift between the workflow and the image would only surface as `ModuleNotFoundError` or "plugin not found" deep inside the pytest run)*
  ```bash
  $ grep -n "WORKDIR /home/build/synth-setter\|/usr/lib/vst3/Surge XT.vst3" docker/ubuntu22_04/Dockerfile
  173:    ln -sfn "/usr/lib/${cmake_arch_name}-linux-gnu/vst3/Surge XT.vst3" "/usr/lib/vst3/Surge XT.vst3"
  232:WORKDIR /home/build/synth-setter
  320:    python -X faulthandler -c "from core import load_plugin; load_plugin('/usr/lib/vst3/Surge XT.vst3')"
  323:    ln -s "/usr/lib/vst3/Surge XT.vst3" "/home/build/synth-setter/plugins/Surge XT.vst3"
  ```
  The workflow's `-v ${{ github.workspace }}:/home/build/synth-setter`, `-e PYTHONPATH=/home/build/synth-setter`, and `-e "SYNTH_SETTER_PLUGIN_PATH=/usr/lib/vst3/Surge XT.vst3"` all align with the image's `WORKDIR` (Dockerfile:232) and the system-installed VST3 path (Dockerfile:173, 320, 323).

- [x] **`scripts/run-linux-vst-headless.sh` exists and is executable** *(Level 5 — referenced by both run blocks; an absent or non-executable wrapper would fail every container invocation)*
  ```bash
  $ ls -la scripts/run-linux-vst-headless.sh
  -rwxr-xr-x 1 root root 5019 May  4 14:52 scripts/run-linux-vst-headless.sh
  ```

- [x] **`pytest -m gpu` selects a non-empty test set, including Surge XT-using tests** *(Level 2 — ran pytest collection with the workflow's marker filter; failure mode this catches: a marker-config drift that makes the workflow run zero tests but exit 0 (false-green); also confirms the `SYNTH_SETTER_PLUGIN_PATH` env var the workflow sets is load-bearing, not just defensive)*
  ```bash
  $ pytest --collect-only -m gpu -q 2>&1 | tail -15
  tests/test_eval.py::test_train_eval
  tests/test_eval.py::test_train_validate
  tests/test_sweeps.py::test_hydra_sweep
  tests/test_sweeps.py::test_hydra_sweep_ddp_sim
  tests/test_train.py::test_train_fast_dev_run_gpu
  tests/test_train.py::test_train_fast_dev_run_gpu_compile
  tests/test_train.py::test_train_epoch_gpu_amp
  tests/test_train.py::test_train_epoch_double_val_loop
  tests/test_train.py::test_train_resume
  tests/test_train.py::test_train_surge_xt[gpu]
  tests/test_train.py::test_train_eval_surge_xt[gpu]
  11/397 tests collected (386 deselected)
  ```
  Note `test_train_surge_xt[gpu]` and `test_train_eval_surge_xt[gpu]` — these GPU-marked tests transitively need the Surge XT VST3 plugin, which is why the `SYNTH_SETTER_PLUGIN_PATH` env var must be set on the `docker run` command (vs. just leaving it for the in-image symlink that the bind-mount shadows).

- [ ] **Workflow runs end-to-end on `gpu-x64` and all GPU tests pass** *(Level 1 — SKIP: maintainer-gated, paid GPU runner. CLAUDE.md prohibits running paid-infra commands without explicit per-action approval, and triggering this 120-min `gpu-x64` job from the verification skill would consume real budget on what is still an unmerged PR)*
  To trigger manually:
  ```bash
  gh workflow run test-gpu.yml --ref ci/test-gpu-dev-snapshot
  gh run watch
  ```
  Confirm: the "Smoke-test GPU visibility inside container" step prints a non-empty `nvidia-smi` row + `cuda: True`, the "Smoke-test Surge XT plugin load" step exits 0, and `pytest -vv -s -m gpu` is green within the 120-min timeout. If the GPU-visibility smoke fails first, the runner does not have `nvidia-container-toolkit` installed and the whole approach needs a different docker GPU setup.

### Cannot verify locally

- Whether `gpu-x64` exposes a CUDA device to containers via `--gpus all`. The new GPU-visibility smoke step is designed to fail fast and surface this on the first dispatch — that's the entire point of including it.

Issue #810, opened 2026-05-05
monitoring: model eval/prediction on captured patch audio


ktinubu:
Track the monitoring/eval scaffolding that lets `scripts/surge_xt_interactive.py` audition patches, render them through the plugin, and run the model end-to-end on the captured audio.

Lands in PR #811:

- `scripts/surge_xt_interactive.py` — `--checkpoint-path` flag triggering an end-to-end audition → render → eval → metrics pipeline (`src/eval.py experiment=surge/test`, `predict_vst_audio.py`, `compute_audio_metrics.py`); output flag renamed `--output-dataset-path` → `--output-dataset-dir-path` (directory; writes `train.h5` plus optional `val.h5`/`test.h5`/`predict.h5`).
- `configs/experiment/surge/test.yaml` — Hydra experiment config consumed by the new flow.
- `scripts/materialize_and_upload_spec.py` — local-dev helper for materializing+uploading a spec to R2 (used to drive the worker container directly, bypassing the SkyPilot launcher).

Pipeline-side companions (10-1k-shards dataset config, OCI shape additions) are tracked separately under #801 / #802.

PR #821, opened 2026-05-05
feat(monitoring): live MIDI input for surge_xt_interactive

ktinubu:
## Summary

Adds a daemon listener thread that forwards MIDI note/CC/pitch-wheel/aftertouch messages from a chosen input port into the running plugin via a thread-safe queue, so the editor responds to a connected keyboard or controller in real time.

- New `--midi-port` flag selects the input by name; `''` auto-picks the first available port; missing/invalid names raise `click.UsageError`.
- `midi_queue` is constructed in `main()` and passed explicitly to `play_audio` and `midi_listener` — no module-level globals.
- `play_audio` now drains the queue each iteration and dispatches via `plugin.process(messages, ...)` so MIDI events are scheduled with audio frames; the original silence-only call path is replaced.
- `midi_listener` filters to performance message types (`MIDI_LISTEN_MESSAGE_TYPES`); other types (sysex, clock, polytouch) are dropped to keep the audio thread responsive.
- `keyboard_loop` gains a `default_params` snapshot so `record_patch` aborts with a clear `ValueError` if a non-spec parameter has drifted (e.g. a knob the user nudged inadvertently). Action-loop `except` is narrowed to `(KeyError, ValueError)` so unrelated exceptions propagate instead of being swallowed.
- `main()` snapshots `default_params` after preset load + flush, then threads the queue and snapshot to the workers.
- New unit test mocks `mido.open_input` to verify `midi_listener` forwards only the allow-listed message types.

## Test plan

- New unit test `TestMidiListener.test_only_relevant_message_types_are_forwarded` (uses `mido` mock; runs in CI without hardware).
- Manual end-to-end check requires Surge XT VST3 + a MIDI input device — see verification comment for the exact commands.

Refs #810


ktinubu:
## Verification Results

- [x] **`midi_queue` is constructed in `main()` only — no module-level global** *(Level 2 — adversarial: a leftover module-level `midi_queue = queue.Queue()` would silently re-introduce shared mutable state across imports. `grep` is acceptable here because the contract IS structural: "no top-level assignment of this name". Higher levels are not applicable.)*
  ```bash
  $ grep -nE "^midi_queue" scripts/surge_xt_interactive.py
  $ echo "exit=$?"
  exit=1   # no top-level binding

  $ grep -nE "midi_queue: queue.Queue" scripts/surge_xt_interactive.py
  250:def play_audio(plugin: VST3Plugin, stop_event: threading.Event, midi_queue: queue.Queue) -> None:
  287:def midi_listener(port_name: str, midi_queue: queue.Queue) -> None:
  571:    midi_queue: queue.Queue = queue.Queue()
  ```
  The only `midi_queue: queue.Queue` bindings are in the two function signatures and the local in `main()`.

- [x] **`--midi-port` click option is registered** *(Level 5 — `--help` invocation requires the project venv, which isn't installed locally. Falling back to source presence + pyright; pyright at 0 errors confirms click can decorate the function.)*
  ```bash
  $ grep -A2 '"--midi-port"' scripts/surge_xt_interactive.py | head -10
      "--midi-port",
      type=str,
      default=None,
  ```
  To verify the help text renders correctly:
  ```bash
  python scripts/surge_xt_interactive.py --help | grep -A4 -- '--midi-port'
  ```

- [x] **Pyright clean on the new/changed Python files** *(Level 2 — same tool, same project config, same version pinned in requirements-app.txt. Catches signature mismatches between `play_audio`/`midi_listener`/`keyboard_loop` and their callers in `main`, plus the new `default_params` plumbing.)*
  ```bash
  $ pyright scripts/surge_xt_interactive.py tests/scripts/test_surge_xt_interactive.py 2>&1 | tail -3
  ```

- [x] **`mido` is a declared project dependency — the new import won't break installs** *(Level 5 — file inspection. The actual install/import contract gets exercised by CI's `run_tests` on every Python version.)*
  ```bash
  $ grep -E "^mido" requirements-app.txt
  mido
  ```

- [ ] **`TestMidiListener.test_only_relevant_message_types_are_forwarded` passes** *(Level 1 — SKIP: project venv not available locally. This is the unit test that proves `midi_listener` filters performance message types and drops `polytouch`/`sysex`/`clock`.)*
  To verify manually:
  ```bash
  pytest -n auto -k "TestMidiListener" tests/scripts/test_surge_xt_interactive.py -v
  ```
  CI's `run_tests` job will run this on Linux + macOS + conda, so the gate is real even though I can't run it here.

- [ ] **`plugin.process(messages, BUFFER_SIZE / SAMPLE_RATE, SAMPLE_RATE, CHANNELS, BUFFER_SIZE, False)` is the correct positional signature for pedalboard's VST3Plugin** *(Level 1 — SKIP: requires Surge XT VST3 + audio output device. The original `play_audio` used the simpler `plugin(silence, SAMPLE_RATE, reset=False)` callable form; this PR replaces it with the explicit `process()` form because callable-form does not accept MIDI events. The signature comes from the `play_audio_recorded` precedent (line ~314 — same module) and the existing test fake `_ConstantPlugin`, but a runtime mis-match would only show up against the real plugin.)*
  To verify manually (requires `make install-surge-xt` and a working audio device):
  ```bash
  python scripts/surge_xt_interactive.py --param-spec-name surge_xt
  # Press 'p' to record a patch (no MIDI input). If play_audio raises a ValueError
  # about unexpected synth output shape, plugin.process() signature is wrong.
  ```

- [ ] **`record_patch` raises `ValueError` when a non-spec parameter has drifted from its post-preset-load default** *(Level 1 — SKIP: the only entry point is the interactive `keyboard_loop`, which blocks on `click.getchar()`. Verifying behaviorally requires either an interactive session against the real plugin or refactoring `record_patch` out for direct invocation. The closure-vs-test trade-off is the same as the existing `keyboard_loop` — it has zero current test coverage.)*
  To verify manually (requires the plugin):
  ```bash
  python scripts/surge_xt_interactive.py --param-spec-name surge_xt
  # 1. With editor open, change ANY non-spec parameter (e.g. master volume).
  # 2. Press 'p'. Expect a ValueError mentioning the drifted parameter name.
  # 3. Revert it. Press 'p' again. Expect the patch to record cleanly.
  ```

- [ ] **`--midi-port=''` auto-selects the first available input** *(Level 1 — SKIP: requires a connected MIDI device.)*
  To verify manually:
  ```bash
  python scripts/surge_xt_interactive.py --param-spec-name surge_xt --midi-port ''
  # Expect a log line: "--midi-port='' — auto-selected first input: <port name>"
  ```

- [ ] **Invalid `--midi-port` raises `click.UsageError`** *(Level 1 — SKIP: needs a MIDI environment to populate `mido.get_input_names()`.)*
  To verify manually:
  ```bash
  python scripts/surge_xt_interactive.py --param-spec-name surge_xt --midi-port "definitely-not-a-real-port-XYZ"
  # Expect: "Error: --midi-port 'definitely-not-a-real-port-XYZ' not in available inputs [...]"
  ```

CI checks (`run_tests`, `code-quality`, etc.) will cover `make test` and `make format` — those are intentionally not duplicated above per the skill's "noise reduction" rule.

scripts/surge_xt_interactive.py
```diff
@@ -322,25 +328,35 @@ def _flush_plugin(plugin: VST3Plugin) -> None:
     plugin.reset()
 
 
-def play_audio(plugin: VST3Plugin, stop_event: threading.Event) -> None:
-    """Stream silence through Surge XT and write synthesized audio to the output device.
+def play_audio(plugin: VST3Plugin, stop_event: threading.Event, midi_queue: queue.Queue) -> None:
```
ktinubu: **[python-style:warn]** `midi_queue: queue.Queue` parameter is unparameterized — prefer `queue.Queue[mido.Message]` (with `from __future__ import annotations` if needed for runtime subscripting on older Pythons) so the contained element type is documented and type-checkable. Same issue at lines 371 and 921.

scripts/surge_xt_interactive.py
```diff
@@ -322,25 +328,35 @@ def _flush_plugin(plugin: VST3Plugin) -> None:
...
         while not stop_event.is_set():
-            synth_output = plugin(silence, SAMPLE_RATE, reset=False)
+            messages = []
+            while True:
+                try:
+                    messages.append(midi_queue.get_nowait())
+                except queue.Empty:
+                    break
+            synth_output = plugin.process(
+                messages, buffer_duration_seconds, SAMPLE_RATE, CHANNELS, BUFFER_SIZE, False
```
ktinubu, line 382: **[code-health:warn]** `plugin.process(messages, buffer_duration_seconds, SAMPLE_RATE, CHANNELS, BUFFER_SIZE, False)` passes a bare positional `False` as the last argument — meaning is opaque at the call site. Use a named keyword argument or an explanatory variable (`reset = False`) so the reader doesn't have to look up the signature. Code Health rule 12 (no flag args / be explicit) and rule 13 (explanatory variables).

ktinubu (commented):
## /review findings

Reviewed the diff against the seven applicable checklists (`tdd-implementation`, `code-health`, `project-standards`, `python-style`, `ml-data-pipeline`, `ml-test`; `shell-style` skipped — no shell files). Applicable items only — pipeline-step / ML-model items mostly N/A for an interactive CLI script.

**Summary: 0 BLOCK, 16 WARN.** No hard rules violated; findings are advisory quality/test-coverage notes.

Key themes:

- **Type annotations**: `queue.Queue` parameters are unparameterized in three places (`play_audio`, `midi_listener`, `main`); `_resolve_midi_port` could prefer `Sequence[str]` over `list[str]`.
- **Magic value**: `abs_tol=1e-6` in `_validate_no_drift` is a literal that should be a named constant; the docstring also bakes the value into prose (CLAUDE.md “Don’t Bake Values Into Comments”).
- **Daemon-thread hygiene**: `midi_listener` has no exception handling and never checks `stop_event` — if the port disappears or the editor closes, behavior is silent / process-exit-only.
- **Test coverage**: the new `play_audio` queue-drain path and `main()`’s MIDI wiring (`mido.get_input_names()`, the `""` auto-pick branch, the thread spawn) have no tests; only `_resolve_midi_port`, `_validate_no_drift`, and `midi_listener`’s filter are covered.
- **Style**: a stray `import queue as _queue` inside a test method body and a missing `__iter__` return annotation in the test fake.
- **Surprising semantics**: empty string `""` as the auto-pick sentinel for `--midi-port` is non-idiomatic (most flags would use `"auto"` or a separate `--auto-midi-port` switch).

Findings posted as individual inline comments below.

scripts/surge_xt_interactive.py
```diff
@@ -322,25 +328,35 @@ def _flush_plugin(plugin: VST3Plugin) -> None:
...
         output_device_name=AudioStream.default_output_device_name,
         sample_rate=PLAYBACK_SAMPLE_RATE,
         buffer_size=BUFFER_SIZE,
     ) as stream:
         while not stop_event.is_set():
-            synth_output = plugin(silence, SAMPLE_RATE, reset=False)
+            messages = []
+            while True:
+                try:
+                    messages.append(midi_queue.get_nowait())
```
ktinubu: **[code-health:warn]** The `messages = []; while True: try: ...; except queue.Empty: break` drain loop is verbose for what is conceptually one operation (“drain queue into list”). Either name the operation in a one-line helper or add a brief comment so the intent reads at a glance. Minor.

scripts/surge_xt_interactive.py
```diff
@@ -352,6 +368,56 @@ def play_audio(plugin: VST3Plugin, stop_event: threading.Event) -> None:
...
 
+def midi_listener(port_name: str, midi_queue: queue.Queue) -> None:
+    """Listen on a MIDI input port and push relevant messages onto ``midi_queue``.
+
+    Filters to performance-relevant types (notes, CC, pitch wheel, aftertouch); other
+    message types (e.g. ``polytouch``, ``sysex``, ``clock``) are dropped. Designed to run on
+    a daemon thread alongside :func:`play_audio`.
+    """
+    logger.info("Listening on MIDI port: %s", port_name)
+    with mido.open_input(port_name) as port_handle:  # pyright: ignore[reportAttributeAccessIssue]
```
ktinubu: **[code-health:warn]** `midi_listener` runs on a daemon thread but has no exception handling. If `mido.open_input` (or iterating the port) raises — port disappears, permission error, mido bug — the daemon thread dies silently and the user sees no MIDI but the editor keeps running with no signal that the listener crashed. Wrap the body in a `try/except Exception` (this is a thread-entry-point isolation boundary, so a broad catch + `logger.exception(...)` + re-raise-or-set-stop is appropriate).

scripts/surge_xt_interactive.py
```diff
@@ -352,6 +368,56 @@ def play_audio(plugin: VST3Plugin, stop_event: threading.Event) -> None:
...
+        for msg in port_handle:
```
ktinubu: **[code-health:warn]** The listener loop `for msg in port_handle:` blocks forever and never checks `stop_event`. When the editor closes, the listener thread keeps running until process exit (it survives only because of `daemon=True`). At minimum add a one-line comment marking this as intentional (“exits at process shutdown via daemon=True”); ideally pass `stop_event` through and break on it.

scripts/surge_xt_interactive.py
```diff
@@ -352,6 +368,56 @@ def play_audio(plugin: VST3Plugin, stop_event: threading.Event) -> None:
...
+            if msg.type in MIDI_LISTEN_MESSAGE_TYPES:
+                midi_queue.put(msg)
+
+
+def _resolve_midi_port(requested: str, available: list[str]) -> str:
```
ktinubu: **[python-style:warn]** `_resolve_midi_port(requested: str, available: list[str])` — prefer the abstract type `Sequence[str]` per python-style PY1 (“Prefer abstract types in annotations: Sequence over list, Mapping over dict”). The function only iterates / does `in` checks, so `Sequence[str]` is the minimum sufficient contract.

scripts/surge_xt_interactive.py
```diff
@@ -352,6 +368,56 @@ def play_audio(plugin: VST3Plugin, stop_event: threading.Event) -> None:
...
+
+    Only parameters absent from ``spec.synth_param_names`` are checked; spec params
+    are expected to vary between recordings. Tolerance is ``abs_tol=1e-6``.
+    """
+    for param_name in plugin.parameters:  # pyright: ignore[reportAttributeAccessIssue]
+        if param_name in spec.synth_param_names:
+            continue
+        current = plugin.parameters[param_name].raw_value  # pyright: ignore[reportAttributeAccessIssue]
+        default = default_params[param_name]
+        if not math.isclose(current, default, abs_tol=1e-6):
```
ktinubu: **[code-health:warn]** `abs_tol=1e-6` is a magic number — extract it to a module-level named constant (e.g., `_DRIFT_TOLERANCE = 1e-6`) so the value is documented in one place and changeable without hunting. Code Health rule 11 (“Replace Magic Numbers with Named Constants”).

scripts/surge_xt_interactive.py
```diff
@@ -352,6 +368,56 @@ def play_audio(plugin: VST3Plugin, stop_event: threading.Event) -> None:
...
+    raise click.UsageError(f"--midi-port {requested!r} not in available inputs {available!r}")
+
+
+def _validate_no_drift(
+    plugin: VST3Plugin, spec: ParamSpec, default_params: dict[str, float]
+) -> None:
+    """Raise ``ValueError`` if any non-spec plugin param drifted from its default.
+
+    Only parameters absent from ``spec.synth_param_names`` are checked; spec params
+    are expected to vary between recordings. Tolerance is ``abs_tol=1e-6``.
```
ktinubu: **[project:warn]** Docstring says ``Tolerance is ``abs_tol=1e-6``.`` — this bakes the literal into the comment, violating the CLAUDE.md “Don’t Bake Values Into Comments” rule. If the constant is named (see the magic-number finding on line 414), reference the symbol; otherwise describe behavior without restating the number.

scripts/surge_xt_interactive.py
```diff
@@ -352,6 +368,56 @@ def play_audio(plugin: VST3Plugin, stop_event: threading.Event) -> None:
...
+    """
+    for param_name in plugin.parameters:  # pyright: ignore[reportAttributeAccessIssue]
+        if param_name in spec.synth_param_names:
+            continue
+        current = plugin.parameters[param_name].raw_value  # pyright: ignore[reportAttributeAccessIssue]
+        default = default_params[param_name]
```
ktinubu, line 457: **[code-health:warn]** `default = default_params[param_name]` will raise a bare `KeyError` with no message if a plugin param appears at runtime that wasn’t in the snapshot taken in `main`. Either guard with `default_params.get(param_name)` and skip with a structured log line, or raise a `ValueError` naming the unsnapshotted parameter so the user sees a useful message instead of a stray `KeyError`.

scripts/surge_xt_interactive.py
```diff
@@ -444,8 +516,17 @@ def record_patch() -> bool:
...
-        if action is not None and action() == _STOP_LOOPING:
-            return synth_patches
+        if action is None:
+            continue
+        try:
+            if action() == _STOP_LOOPING:
+                return synth_patches
+        except (KeyError, ValueError):
+            # Surface the traceback now; without this it stays buffered in
+            # ``keyboard_future`` until ``main`` reaches ``.result()`` after the editor closes.
```
ktinubu: **[project:warn]** The two-line essay-style comment inside the `except` block violates the CLAUDE.md comment-hygiene rule “Keep comments terse — typically one short line. Multi-sentence prose explanations inline are a smell.” Make it a one-line pointer (e.g., link to the issue / PR describing the buffered-traceback behavior in `keyboard_future`).

scripts/surge_xt_interactive.py
```diff
@@ -817,14 +918,30 @@ def main(
     if session_recording_path is not None:
         play_audio_recorded(plugin, session_recording_path)
 
+    midi_queue: queue.Queue = queue.Queue()
```
ktinubu: **[code-health:warn]** `midi_queue` is unconditionally constructed even when `midi_port is None` (no listener spawned). The queue is then handed to `play_audio` and drained every iteration for no reason. Either scope the queue inside the `if midi_port is not None:` block (and pass `None` to `play_audio`, branching on it), or add a one-line comment noting the empty-drain is intentional. Minor.

scripts/surge_xt_interactive.py
```diff
@@ -817,14 +918,30 @@ def main(
...
+    if midi_port is not None:
+        resolved_port = _resolve_midi_port(
+            midi_port,
+            mido.get_input_names(),  # pyright: ignore[reportAttributeAccessIssue]
+        )
+        if midi_port == "":
```
ktinubu, line 983: **[code-health:warn]** Empty string `""` as the “auto-pick” sentinel for `--midi-port` is surprising — most click flags treat `""` and unset interchangeably, and CLI users typically expect `--midi-port=auto` or a separate `--auto-midi-port` switch. The semantics are explained in help text, but the API still reads opaquely. Consider `default="auto"` (and treat that string as the sentinel) or a flag-pair pattern. Code Health rule 20 (opacity).

scripts/surge_xt_interactive.py
```diff
@@ -817,14 +918,30 @@ def main(
...
+            logger.info("--midi-port='' — auto-selected first input: %s", resolved_port)
+        threading.Thread(
```
ktinubu, line 986: **[code-health:warn]** The MIDI listener thread is started but never joined, never sent a stop signal, and never exception-checked — it relies entirely on `daemon=True` for cleanup. There’s no test exercising the lifecycle and no comment marking this as intentional. Add a one-line “exits at process shutdown via daemon=True” comment so the next reader doesn’t try (incorrectly) to add a `.join()` or assume the thread is cooperatively shutdown.

tests/scripts/test_surge_xt_interactive.py
```diff
@@ -375,6 +375,189 @@ def test_passes_expected_midi_events(self, surge_xt_interactive, tmp_path: Path)
...
+
+
+class TestMidiListener:
+    """``midi_listener`` filters mido messages by type and forwards them to a queue."""
+
+    def test_only_relevant_message_types_are_forwarded(
+        self, surge_xt_interactive, monkeypatch: pytest.MonkeyPatch
+    ) -> None:
+        """note_on/off, control_change, pitchwheel, aftertouch are queued; others are dropped."""
+        import queue as _queue
```
ktinubu: **[python-style:warn]** `import queue as _queue` is inside the test method body. Python-style PY1: imports go at module top, grouped. Move it to the file’s import block at the top (alongside the existing `import click`, `import h5py`, etc.).

tests/scripts/test_surge_xt_interactive.py
```diff
@@ -375,6 +375,189 @@ def test_passes_expected_midi_events(self, surge_xt_interactive, tmp_path: Path)
...
+    def __init__(self, messages: list[_FakeMidiMessage]) -> None:
+        self._messages = messages
+
+    def __enter__(self) -> "_FakeMidiPortHandle":
+        return self
+
+    def __exit__(self, *_exc: object) -> None:
+        return None
+
+    def __iter__(self):
```
ktinubu: **[python-style:warn]** `def __iter__(self):` is missing a return type annotation. PY8: type annotations on all public API function signatures (Ruff `ANN001` enforces param annotations; the project also expects return annotations on dunder methods). Annotate as `Iterator[_FakeMidiMessage]` (importing `Iterator` from `collections.abc`).

tests/scripts/test_surge_xt_interactive.py
```diff
@@ -375,6 +375,189 @@ def test_passes_expected_midi_events(self, surge_xt_interactive, tmp_path: Path)
...
+        )
+
+        midi_queue: _queue.Queue = _queue.Queue()
+        surge_xt_interactive.midi_listener("fake-port", midi_queue)
+
+        drained: list[_FakeMidiMessage] = []
+        while not midi_queue.empty():
+            drained.append(midi_queue.get_nowait())
+
+        assert [m.type for m in drained] == [m.type for m in forwarded]
```
ktinubu: **[tdd:warn]** No test exercises `play_audio`’s new queue-drain behavior — the new code path that pulls messages off `midi_queue` and passes them into `plugin.process(...)`. The listener-side filter is tested, and `midi_listener` is tested in isolation, but the producer→consumer wiring inside `play_audio` is not. Add a test that pre-fills a queue, calls `play_audio` with a fake `plugin` and a fake `AudioStream` (or factor the inner step into a small helper that’s easier to test), and asserts the queued messages reach `plugin.process`. Currently a regression where `play_audio` stops draining the queue would not be caught by any test.

scripts/surge_xt_interactive.py
```diff
@@ -272,6 +284,20 @@ def play_audio(plugin: VST3Plugin, stop_event: threading.Event) -> None:
...
+
+    Filters to performance-relevant types (notes, CC, pitch wheel, aftertouch); other
+    message types (e.g. ``polytouch``, ``sysex``, ``clock``) are dropped. Designed to run on
+    a daemon thread alongside :func:`play_audio`.
+    """
+    logger.info("Listening on MIDI port: %s", port_name)
+    with mido.open_input(port_name) as port_handle:
+        for msg in port_handle:
+            if msg.type in MIDI_LISTEN_MESSAGE_TYPES:
+                midi_queue.put(msg)
```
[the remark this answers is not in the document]
ktinubu: mido.Message → tuple — Fixed in 36cc4ad. midi_listener now enqueues `(msg.bytes(), 0.0)` to match `make_midi_events`.

scripts/surge_xt_interactive.py
```diff
@@ -498,14 +568,36 @@ def main(
...
+            logger.info("--midi-port='' — auto-selected first input: %s", resolved_port)
+        elif midi_port in available_ports:
+            resolved_port = midi_port
+        else:
+            raise click.UsageError(
+                f"--midi-port {midi_port!r} not in available inputs {available_ports!r}"
+            )
+        threading.Thread(
+            target=midi_listener, args=(resolved_port, midi_queue), daemon=True
+        ).start()
```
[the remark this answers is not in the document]
ktinubu: Listener vs session-recording — Fixed in 36cc4ad. `--midi-port` and `--session-recording-path` are now mutually exclusive (UsageError up front).

tests/scripts/test_surge_xt_interactive.py
```diff
@@ -371,3 +371,76 @@ def test_passes_expected_midi_events(self, surge_xt_interactive, tmp_path: Path)
...
```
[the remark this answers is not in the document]
ktinubu: Test asserts (bytes, time) shape — Fixed in 36cc4ad. Updated `test_only_relevant_message_types_are_forwarded` to assert tuple shape, plus new `TestPlayAudioQueueDrain` covers the play_audio drain path end-to-end.

scripts/surge_xt_interactive.py
ktinubu: Parameterized type — Fixed in 36cc4ad. `queue.Queue[tuple[bytes, float]]` (and `| None` in the play_audio signature).

scripts/surge_xt_interactive.py
ktinubu: `reset` kwarg — Fixed in 36cc4ad. `plugin.process(..., reset=False)`.

scripts/surge_xt_interactive.py
ktinubu: Drain loop verbosity — Justified as-is. The 4-line `try/except queue.Empty: break` is the canonical idiom; a helper would be YAGNI.

scripts/surge_xt_interactive.py
ktinubu: Listener exception handling — Fixed in 36cc4ad. Wrapped body in `try/except Exception:` + `logger.exception` so failures aren't silent. New test `test_open_input_failure_logs_and_exits_thread` covers it.

scripts/surge_xt_interactive.py
ktinubu: `for msg in port_handle` no stop_event — Justified as-is. `daemon=True` means process exit reaps the thread; cooperative shutdown is YAGNI for an interactive CLI.

scripts/surge_xt_interactive.py
ktinubu: `Sequence[str]` — Fixed in 36cc4ad.

scripts/surge_xt_interactive.py
ktinubu: `_DRIFT_TOLERANCE` — Fixed in 36cc4ad.

scripts/surge_xt_interactive.py
ktinubu: Docstring `abs_tol=1e-6` — Fixed in 36cc4ad. Removed.

scripts/surge_xt_interactive.py
ktinubu: `default_params[name]` bare lookup — Justified as-is. The dict is built from `plugin.parameters`, so a missing key would be a programmer bug, not user error; the bare `[]` makes that explicit.

scripts/surge_xt_interactive.py
ktinubu: Multi-line essay comment — Fixed in 36cc4ad. Trimmed to one line.

scripts/surge_xt_interactive.py
ktinubu: Conditional queue construction — Fixed in 36cc4ad. `midi_queue` is constructed only when `--midi-port` is set; play_audio handles `None`.

scripts/surge_xt_interactive.py
ktinubu: `--midi-port=""` sentinel — Justified as-is. Documented in the click help and PR description; CLI contract for this PR.

scripts/surge_xt_interactive.py
ktinubu: Listener thread no join — Justified as-is. Same rationale as 3204006629 — `daemon=True` is sufficient for a CLI.

tests/scripts/test_surge_xt_interactive.py
ktinubu: Method-body import — Fixed in 36cc4ad. Moved `import queue` to module top.

tests/scripts/test_surge_xt_interactive.py
ktinubu: `__iter__` annotation — Fixed in 36cc4ad. `Iterator[_FakeMidiMessage]`.

tests/scripts/test_surge_xt_interactive.py
ktinubu: No play_audio drain test — Fixed in 36cc4ad. New `TestPlayAudioQueueDrain` covers both `midi_queue` populated and `midi_queue=None` paths.

scripts/surge_xt_interactive.py
```diff
@@ -322,25 +328,35 @@ def _flush_plugin(plugin: VST3Plugin) -> None:
...
-            synth_output = plugin(silence, SAMPLE_RATE, reset=False)
+            messages = []
+            while True:
+                try:
+                    messages.append(midi_queue.get_nowait())
+                except queue.Empty:
+                    break
+            synth_output = plugin.process(
+                messages, buffer_duration_seconds, SAMPLE_RATE, CHANNELS, BUFFER_SIZE, False
+            )
```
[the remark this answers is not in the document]
ktinubu: plugin.process tuple shape (re-review) — Fixed in 36cc4ad (same as 3192022737).

scripts/surge_xt_interactive.py
```diff
@@ -352,6 +368,56 @@ def play_audio(plugin: VST3Plugin, stop_event: threading.Event) -> None:
...
+
+    Filters to performance-relevant types (notes, CC, pitch wheel, aftertouch); other
+    message types (e.g. ``polytouch``, ``sysex``, ``clock``) are dropped. Designed to run on
+    a daemon thread alongside :func:`play_audio`.
+    """
+    logger.info("Listening on MIDI port: %s", port_name)
+    with mido.open_input(port_name) as port_handle:  # pyright: ignore[reportAttributeAccessIssue]
+        for msg in port_handle:
+            if msg.type in MIDI_LISTEN_MESSAGE_TYPES:
+                midi_queue.put(msg)
```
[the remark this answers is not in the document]
ktinubu: listener tuple format (re-review) — Fixed in 36cc4ad.

tests/scripts/test_surge_xt_interactive.py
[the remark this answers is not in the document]
ktinubu: test format validation (re-review) — Fixed in 36cc4ad.

scripts/surge_xt_interactive.py
```diff
@@ -322,25 +331,45 @@ def _flush_plugin(plugin: VST3Plugin) -> None:
     plugin.reset()
 
 
-def play_audio(plugin: VST3Plugin, stop_event: threading.Event) -> None:
-    """Stream silence through Surge XT and write synthesized audio to the output device.
+def play_audio(
+    plugin: VST3Plugin,
+    stop_event: threading.Event,
+    midi_queue: "queue.Queue[tuple[bytes, float]] | None",
+) -> None:
```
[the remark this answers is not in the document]
ktinubu: Fixed in 0f150b5. Annotations changed to `Queue[tuple[list[int], float]]` / `list[tuple[list[int], float]]` to match `mido.Message.bytes()`'s actual `list[int]` return type, and `midi_listener`'s docstring now calls out the matching pedalboard `List[int]` form. Confirmed at runtime: `mido.Message('note_on', note=60, velocity=64).bytes() == [144, 60, 64]` (a `list`).

tests/scripts/test_surge_xt_interactive.py
```diff
@@ -375,6 +378,317 @@ def test_passes_expected_midi_events(self, surge_xt_interactive, tmp_path: Path)
...
 
 
+class _FakeMidiMessage:
+    """Minimal stand-in for a ``mido.Message`` — exposes ``type`` and ``bytes()``."""
+
+    def __init__(self, msg_type: str, payload: bytes = b"\x90\x3c\x40") -> None:
+        self.type = msg_type
+        self._payload = payload
+
+    def bytes(self) -> bytes:
```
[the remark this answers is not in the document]
ktinubu: Fixed in 0f150b5. `_FakeMidiMessage.bytes()` now returns `list[int]` and all test payloads (e.g. `[0x90, 0x3C, 0x40]`) match. `_RecordingPlugin` and the queue-drain assertion follow suit.

scripts/surge_xt_interactive.py
```diff
@@ -322,25 +331,45 @@ def _flush_plugin(plugin: VST3Plugin) -> None:
...
+                        messages.append(midi_queue.get_nowait())
+                    except queue.Empty:
+                        break
+            synth_output = plugin.process(
+                messages,
+                buffer_duration_seconds,
+                SAMPLE_RATE,
+                CHANNELS,
+                BUFFER_SIZE,
+                reset=False,
```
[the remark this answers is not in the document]
ktinubu: Fixed in 609c492. Reverted to positional `False` and added a one-line comment so the bare boolean isn't a code-health smell. Matches `_flush_plugin`, `play_audio_recorded`, and `src/data/vst/core.py` — and avoids the pedalboard C-extension keyword-arg portability concern.

scripts/surge_xt_interactive.py
```diff
@@ -817,14 +946,31 @@ def main(
...
+        resolved_port = _resolve_midi_port(
+            midi_port,
+            mido.get_input_names(),  # pyright: ignore[reportAttributeAccessIssue]
+        )
+        if midi_port == "":
+            logger.info("--midi-port='' — auto-selected first input: %s", resolved_port)
+        midi_queue = queue.Queue()
+        threading.Thread(
+            target=midi_listener, args=(resolved_port, midi_queue), daemon=True
+        ).start()
```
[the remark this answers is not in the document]
ktinubu: Fixed in 609c492. `midi_listener` now takes a `stop_event` and polls via `port.poll()` non-blockingly, checking the event every `_MIDI_POLL_INTERVAL_SECONDS` (10 ms). `main` constructs `stop_event` up front so the listener and audio threads share it. New test `test_stop_event_exits_listener_with_no_messages` covers the shutdown path; the existing forwarding test now drives a real thread + drain-signal handshake (no `time.sleep`).

scripts/surge_xt_interactive.py
```diff
@@ -322,25 +331,45 @@ def _flush_plugin(plugin: VST3Plugin) -> None:
...
         buffer_size=BUFFER_SIZE,
     ) as stream:
         while not stop_event.is_set():
-            synth_output = plugin(silence, SAMPLE_RATE, reset=False)
+            messages: list[tuple[list[int], float]] = []
+            if midi_queue is not None:
+                while True:
+                    try:
+                        messages.append(midi_queue.get_nowait())
+                    except queue.Empty:
```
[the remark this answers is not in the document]
ktinubu: Fixed in 609c492. Per-buffer drain capped at `_MAX_MIDI_EVENTS_PER_BUFFER = 64`. Excess events stay queued and process on the next buffer (~12 ms at 44.1k/512), well within human-perceptible latency. New test `test_drain_is_capped_at_max_midi_events_per_buffer` enqueues cap+5 events and asserts exactly `cap` drain on the first buffer with 5 left in the queue.

scripts/surge_xt_interactive.py
```diff
@@ -352,6 +392,76 @@ def play_audio(plugin: VST3Plugin, stop_event: threading.Event) -> None:
...
+    stop_event: threading.Event,
+) -> None:
+    """Listen on a MIDI input port and push (bytes, time) tuples onto ``midi_queue``.
+
+    Filters to performance-relevant types (notes, CC, pitch wheel, aftertouch); other
+    message types (e.g. ``polytouch``, ``sysex``, ``clock``) are dropped. Each forwarded
+    message is converted to ``(msg.bytes(), 0.0)`` so ``plugin.process`` schedules it at the
+    start of the next audio buffer — the format used elsewhere in the repo (see
+    :func:`src.data.vst.core.make_midi_events`). ``mido.Message.bytes()`` returns
+    ``list[int]``, matching the ``List[int]`` form accepted by pedalboard's
```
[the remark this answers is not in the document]
ktinubu: Fixed in 4afc761. Docstring now says `(list[int], float)` and notes that `mido.Message.bytes()` returns a sequence of MIDI status bytes.

Issue #844, opened 2026-05-07
test(testing): replace mocks with fakes + thread-event synchrony in surge_xt_interactive tests


ktinubu:
## Context

Recent rounds of review on PR #821 caught real bugs (MIDI message format, `bytes`-vs-`list[int]` type mismatch, listener-vs-session-recording leak) that the existing test suite missed. Investigating why: the suite leans heavily on `monkeypatch.setattr` to swap module-level functions, on subprocess mocks that pre-populate `sample_*/` directories, and on `MagicMock`-style argv-spy assertions. These tests pass even when the SUT calls `plugin.process(...)` with the wrong argument shape, because the mocks accept anything.

**Goal:** rewrite `tests/scripts/test_surge_xt_interactive.py` so every test exercises real code paths, asserts on real state, and uses `threading.Event` for synchronization rather than time-based polling. Replace mocks with hand-rolled fakes (duck-typed classes that capture state) and inject them via narrow DI seams in production. Real plugin/audio I/O moves behind the existing `@pytest.mark.requires_vst @pytest.mark.slow` gate. Exceptions are tested only when the production code's contract explicitly raises (greenlit case-by-case).

This is a bigger-than-one-PR refactor. Land in steps so the file stays green at every commit.

## Critical Files

- `scripts/surge_xt_interactive.py` — production seams added here
- `tests/scripts/test_surge_xt_interactive.py` — the migration target
- `tests/conftest.py` — already has `surge_xt_smoke_datasets` fixture; reuse for e2e
- `tests/data/vst/test_preset_coverage.py` — precedent for `threading.Event` + daemon thread driving `plugin.show_editor`
- `Makefile` — `make install-surge-xt`, `make test`, `make test-full` (no changes needed)

## Test Taxonomy

**A. Already mock-free, keep as-is (~15 tests):** `TestDecodePredictionRow`, `TestPredictionRefType`, `TestDatasetRefType`, `TestLoadPredictionSynthParams`, `TestLoadDatasetSynthParams`, `TestExpectedPredictionFilenames`, `TestValidateMetricsDf`, `TestValidatePredictions`, `TestValidateNoDrift`, `TestResolveMidiPort`, `TestPlayAudioRecorded` (uses `_ConstantPlugin` fake).

**B. Convert to fake-based via DI (7 tests):**
- `TestRunPredict::test_passes_d_out_override_and_absolute_paths` — inject `subprocess_runner`
[...]

PR #850, opened 2026-05-07
test(surge-interactive): step 2 — TestRunPredict (#844)

ktinubu:
## Summary

Step 2 of [#844](https://github.com/tinaudio/synth-setter/issues/844). Stacked on top of [#847](https://github.com/tinaudio/synth-setter/pull/847) (step 1 — DI seams).

Replaces `monkeypatch.setattr(surge_xt_interactive.subprocess, "check_call", ...)` in `TestRunPredict::test_passes_d_out_override_and_absolute_paths` with direct injection of a hand-rolled `_RecordingSubprocessRunner` via the `subprocess_runner` kwarg added in step 1.

The fake is a 12-line class with `__call__(args, **kwargs)` that records `calls: list[list[str]]` and `kwargs_per_call: list[dict]`. No mock library, no module-level patching — the injected dependency *is* the dependency the production code uses. Reused by steps 3 and 6/7 from other test classes that hit the same seam.

## Diff

```
tests/scripts/test_surge_xt_interactive.py | 46 +++++++++++++++++++++---------
1 file changed, 32 insertions(+), 14 deletions(-)
```

## Test plan

- [x] `pytest tests/scripts/test_surge_xt_interactive.py::TestRunPredict -v` — 1 passed
- [x] Full file `pytest tests/scripts/test_surge_xt_interactive.py -m "not slow"` — 63 passed, 4 deselected
- [x] `make format` — clean

## Note on base branch

This PR's base is `test/de-mock-surge-xt-interactive-step-1`, not `main`. Once step 1 lands and the base branch retargets to `main` automatically, this PR will show only the step-2 diff against main.

Refs #844

### DIFF
--- a/tests/scripts/test_surge_xt_interactive.py
+++ b/tests/scripts/test_surge_xt_interactive.py
@@ -836,14 +836,32 @@ class TestExpectedPredictionFilenames:
         assert surge_xt_interactive._expected_prediction_filenames(num_samples=0) == []
 
 
+class _RecordingSubprocessRunner:
+    """Test fake matching the ``SubprocessRunner`` shape from ``surge_xt_interactive``.
+
+    Records every invocation's positional argv and keyword arguments so tests assert on
+    real state instead of a ``MagicMock`` argv-spy. Used by every test that exercises a
+    ``*_runner=`` seam introduced in the de-mock refactor (see issue #844).
+
+    Returns ``0`` from ``__call__`` (a valid ``subprocess.check_call`` return; production
+    callers ignore the value or read ``returncode`` from a ``CompletedProcess``-like object,
+    which we don't simulate here — YAGNI).
+    """
+
+    def __init__(self) -> None:
+        self.calls: list[list[str]] = []
+        self.kwargs_per_call: list[dict[str, object]] = []
+
+    def __call__(self, args: list[str], **kwargs: object) -> int:
+        self.calls.append(list(args))
+        self.kwargs_per_call.append(dict(kwargs))
+        return 0
+
+
 class TestRunPredict:
     """``_run_predict`` builds the ``src/eval.py`` invocation with the right Hydra overrides."""
 
-    def test_passes_d_out_override_and_absolute_paths(
-        self,
-        surge_xt_interactive,
-        monkeypatch: pytest.MonkeyPatch,
-    ) -> None:
+    def test_passes_d_out_override_and_absolute_paths(self, surge_xt_interactive) -> None:
         """``_run_predict`` overrides ``model.net.d_out`` with the encoded width of
         ``param_spec_name`` (otherwise the ``???`` sentinel in ``surge/test.yaml`` would error),
         and resolves all paths to absolute (otherwise Hydra's ``chdir`` would break relative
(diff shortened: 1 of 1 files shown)


ktinubu:
## Verification Results

- [x] **`TestRunPredict::test_passes_d_out_override_and_absolute_paths` no longer monkeypatches and still passes** *(Level 2 — runs the converted test against the real `_run_predict`. This is the central promise of step 2: the test exercises the seam, not the module's `subprocess` global)*
  ```bash
  $ python3 -c "
  in_test_run_predict = False
  hits = []
  src = open('tests/scripts/test_surge_xt_interactive.py').read().splitlines()
  for i, line in enumerate(src, 1):
      if line.startswith('class TestRunPredict'):
          in_test_run_predict = True
      elif in_test_run_predict and line.startswith('class '):
          in_test_run_predict = False
      if in_test_run_predict and ('monkeypatch.setattr' in line or 'MonkeyPatch' in line):
          hits.append((i, line.strip()))
  print(f'monkeypatch references inside TestRunPredict: {len(hits)}')
  "
  monkeypatch references inside TestRunPredict: 0

  $ pytest tests/scripts/test_surge_xt_interactive.py::TestRunPredict -v 2>&1 | tail -2
  tests/scripts/test_surge_xt_interactive.py::TestRunPredict::test_passes_d_out_override_and_absolute_paths PASSED [100%]
  ============================== 1 passed in 0.81s ==============================
  ```

- [x] **Adversarial / mutation test: the new test catches a seam regression** *(Level 2 — adversarial. The old monkeypatch-based test would silently keep passing if the seam were disconnected, because the global `subprocess.check_call` was being patched. The whole point of step 2 is that the new test FAILS when production stops honoring `subprocess_runner`. Proving this is the strongest evidence the conversion is meaningful.)*
  ```bash
  # Mutate the production seam: ignore the injected runner.
  $ python3 -c "
  src = open('scripts/surge_xt_interactive.py').read()
  old = 'runner = subprocess_runner if subprocess_runner is not None else subprocess.check_call'
  mutated = 'runner = subprocess.check_call  # mutation: ignore injected runner'
  assert old in src
  open('scripts/surge_xt_interactive.py', 'w').write(src.replace(old, mutated, 1))
  "

  # Test now FAILS because the real subprocess fires:
  $ pytest tests/scripts/test_surge_xt_interactive.py::TestRunPredict 2>&1 | tail -3
  FAILED tests/scripts/test_surge_xt_interactive.py::TestRunPredict::test_passes_d_out_override_and_absolute_paths
    - subprocess.CalledProcessError: Command '['/venv/main/bin/python3', ...src/eval.py..., 'mode=predict']' returned non-zero exit status
  1 failed in 7.46s

  # Restore production code, test passes again:
  $ pytest tests/scripts/test_surge_xt_interactive.py::TestRunPredict 2>&1 | tail -1
  ============================== 1 passed in 0.82s ==============================
  ```
  Failure reason matters: under the mutation the *real* `subprocess.check_call` actually launched `python src/eval.py …` and failed. Compare to the old monkeypatch-based test, which would have silently passed under the same mutation because the patch overrode the module global regardless of what the seam did. The new test has strictly higher fidelity.

- [x] **Removing the monkeypatch did not break adjacent tests in the same file** *(Level 1 closest available — the surge_xt module is exercised across the whole file via shared fixtures (`surge_xt_interactive`, `simple_pred_tensor`, etc.); a regression in module-level state would surface here)*
  ```bash
  $ pytest tests/scripts/test_surge_xt_interactive.py -m "not slow" --deselect tests/scripts/test_surge_xt_interactive.py::TestLoadDatasetSynthParams 2>&1 | tail -1
  ======================= 63 passed, 4 deselected in 0.95s =======================
  ```

- [ ] **`@requires_vst` smoke against the real Surge XT plugin** *(Level 1 — SKIP: `plugins/Surge XT.vst3` not installed in this dev environment; gate is plugin presence, not GPU.)*
  To verify manually:
  ```bash
  make install-surge-xt
  pytest tests/scripts/test_surge_xt_interactive.py -m "requires_vst"
  ```
  Step 2 only converts a test that already used a fake `subprocess` (mocked or injected) — production code is unchanged from step 1, so the requires_vst surface has no new risk.

PR #858, opened 2026-05-07
refactor(workflows): split test-dataset-generation; rename launcher

ktinubu:
## Summary

Clean refactor of the dataset-generation CI surface — no behavior change. Extracts two reusable workflows from `test-dataset-generation.yml` and renames the launcher module to drop the misleading `_smoke` suffix.

- **`.github/workflows/generate-dataset-shards.yaml`** (new, `workflow_call`) — official launcher reusable. Inputs: `provider`, `dataset_config`, `image_tag`, `cluster_name`, `num_workers`, `tail`, `api_server`, `local`, `artifact_name`. Owns the kind setup for `skypilot-local` and the in-container launcher invocation for `runpod` / `oci`.
- **`.github/workflows/validate-dataset-shards.yaml`** (new, `workflow_call`) — owns `validate-spec` + `validate-shard` jobs. Inputs: `image_tag`, `artifact_name`.
- **`.github/workflows/test-dataset-generation.yml`** is now a thin wrapper: keeps PR/dispatch triggers and the `setup` provider-matrix job, then calls the two reusables per provider. Workflow_dispatch surface unchanged. The docker-only `local` row stays inline (no launcher).
- **`pipeline/entrypoints/skypilot_launch_smoke.py`** → **`skypilot_launch.py`** (and the matching test). Updated all references: `test-skypilot-debug.yml`, the `paths:` filter, compute templates' header comments, `scripts/sync_worker_checkout.sh`, and the doc set. CHANGELOG history left intact.
- **`.github/workflows/dataset-generation.yml`** deleted — obsolete, no callers, superseded by the unified launcher.

Each launcher flag the test wrapper passes to the reusable matches today's hardcoded value (e.g. `num_workers: '1'` + `local: true` for `skypilot-local`; defaults for `runpod` / `oci`).

## Out of scope (chains on top of this)

- **R2-as-coordination-medium**: stages will look up the spec from R2 instead of passing `/tmp/run-metadata` artifacts. `pipeline.ci.validate_spec` will accept an R2 URI; rclone copy/lookups will move into pipeline py code.
- **Surface launcher knobs on the test wrapper**: add `num_workers` / `tail` / `local` to `test-dataset-generation.yml`'s `workflow_dispatch` inputs.

## Test plan

- [x] `make format` — pre-commit clean (ruff, pyright, prettier, shellcheck, mdformat, codespell)
- [x] `make test` — 426 passed, 0 failures
[...]

Issue #843, opened 2026-05-07
task(skypilot): skypilot-local row can't host parallel workers — single-node kind capacity + no retry-on-pod-free


ktinubu:
## Context

PR #840 introduced a `skypilot-local` matrix row in `test-dataset-generation.yml` that exercises the launcher path against a kind cluster provisioned on the GH runner via `sky local up`. We pinned that row to `--num-workers 1` as a tactical fix (commit landing alongside this issue) after observing the failure mode below; this issue tracks the deeper design question.

## Observed failure

[Job 74882524219 in run 25514685014](https://github.com/tinaudio/synth-setter/actions/runs/25514685014/job/74882524219). Timeline:

```
18:41:13  r0/r1/r2 all start provisioning in parallel (one launcher process,
          three sky.launch calls fanned out)
18:41:14  All 3 log "Running on cluster: synth-setter-smoke-skypilot-local-..."
18:41:33  r1 fails provision: "Failed to provision all possible launchable
          resources. Relax the task's resource requirements:
          1x Kubernetes(cpus=2+, mem=4+, image_id=...:dev-snapshot, disk_size=20)"
18:41:33  r2 fails provision: same error
18:42:22  r0 succeeds (Cluster launched)
18:42:34  r0 streams logs while running the worker job (~2.5 min of generation)
18:44:58  r0 finishes successfully (status=SUCCEEDED, shard uploaded to R2)
18:45:00  Launcher moves all 3 workers to teardown:
            r0 → succeeds (real cluster)
            r1 → "Cluster does not exist" (never provisioned)
            r2 → "Cluster does not exist" (never provisioned)
18:45:19  Launcher exits 1: "2 of 3 worker(s) failed: r1(rc=-1), r2(rc=-1)"
```

r1 and r2 didn't fail because r0's teardown destroyed something they needed — they failed at provision time, ~2.5 min before r0 even started running its job. r0 happened to win the race for the single available pod slot on the single-node kind cluster.

## Root cause: shape mismatch with cloud rows

| Provider | Where each `sky.launch` lands |
| --- | --- |
| `runpod`, `oci` | A fresh VM. Three `--num-workers 3` calls → three independent VMs. No contention. |
| `skypilot-local` | Three pods scheduled on the **same single-node kind cluster** (the GH runner). Combined ask of 3×(cpus=2+, mem=4+) saturates the runner's ~4 vCPU / ~16 GB. |

[...]

Issue #841, opened 2026-05-07
task(skypilot): make local-vs-remote dispatch explicit; rename or split SKYPILOT_API_SERVER_ENDPOINT


ktinubu:
## Context

PR #840 shipped a fix (commit 8eadbf8) that had to drop `SKYPILOT_API_SERVER_ENDPOINT` from the `skypilot-local` matrix step's env block in `.github/workflows/test-dataset-generation.yml`.

The symptom — concrete failure that motivated this issue: [job 74876523720](https://github.com/tinaudio/synth-setter/actions/runs/25513017873/job/74876523720). The runner provisioned a kind cluster via `sky local up`, kind-loaded the dev-snapshot image into it, then invoked `pipeline.entrypoints.skypilot_launch_smoke` with `SKYPILOT_API_SERVER_ENDPOINT` inherited from the secrets. The launcher's `sky.*` calls were dispatched to the remote API server, which has its own Kubernetes context that didn't have the image kind-loaded, so all three workers failed at provision with `Kubernetes cluster does not contain any instances satisfying the request: ... image_id=docker:tinaudio/synth-setter:dev-snapshot` and teardown couldn't find the clusters because they were never registered locally.

## The design smell

The current contract — "if `SKYPILOT_API_SERVER_ENDPOINT` is non-empty, dispatch to the remote server; else local SDK" — is implicit. Any workflow step intended to run locally is one accidental env passthrough away from being silently routed to the remote server. There's no positive signal in the launcher invocation that says "this run is intended to be remote" or "this run is intended to be local"; the mode is determined by the presence/absence of an env var that originates from the secrets.

PR #840's fix is correct for that one step, but the broader concern is that we'll keep stepping on this rake.

## Options to explore (none pre-decided)

1. **Rename to a project-prefixed name** (e.g. `SYNTH_SETTER_SKY_REMOTE_ENDPOINT`). Stops it looking like a passthrough of a SkyPilot SDK env var that "everyone should always set". Trade-off: we lose the SkyPilot SDK convention, which is itself a soft signal about what the variable does.
2. **Add an explicit launcher flag** (`--api-server <url>` or `--local`) so each call site declares its intended mode by argument. The env var becomes a default value rather than the load-bearing switch. Trade-off: more args to thread through, but the workflow YAML reads less ambiguously.
[...]

PR #864, opened 2026-05-08
feat(workflows): surface launcher knobs on test-dataset-generation dispatch

### LINKED ISSUE #861: feat(workflows): surface launcher knobs on test-dataset-generation dispatch
## Summary

Follow-up to #856. Surface the launcher knobs (`num_workers`, `tail`, `local`, `api_server`) on `test-dataset-generation.yml`'s `workflow_dispatch` inputs so operators can override them at manual-trigger time without editing the workflow.

## What changes

`test-dataset-generation.yml` adds workflow_dispatch inputs:

- `num_workers` (number, default empty → use config value)
- `tail` (boolean, default true)
- `local` (boolean, default false)
- `api_server` (string, default empty)

These are passed straight through to the `generate-dataset-shards.yaml` reusable. PR-triggered runs continue to use today's hardcoded values (skypilot-local: `local=true, num_workers=1, tail=true`); manual dispatch uses the operator's overrides.

## Why now

Today the test wrapper's dispatch surface only exposes `dataset_config`, `docker_image_tag`, `provider`. Operators who want to test a multi-worker run, or smoke-test a dispatch that intentionally targets a remote SkyPilot API server, have to edit the YAML and push. With the reusable workflow already accepting all the knobs, surfacing them on the wrapper is a one-file change.

## Verification

- `make format`
- `gha-workflow-validator` on `test-dataset-generation.yml`
- Manual dispatch with overrides (e.g. `provider=skypilot-local, num_workers=2, tail=false`) — confirm the launcher receives the overridden values.

## Out of scope

R2-as-coordination (parent issue) — wiring this PR's overrides depends on knowing the new reusable's input shape, but doesn't depend on R2 vs artifact handoff.

Refs #856, depends on R2-as-coordination follow-up landing first (stacked PR).

ktinubu:
## Summary

Stacked on #863. Surfaces the launcher knobs (`num_workers`, `tail`, `local_dispatch`, `api_server`) on `test-dataset-generation.yml`'s `workflow_dispatch` inputs so operators can override them at manual-trigger time without editing the workflow.

The reusable `generate-dataset-shards.yaml` already accepts all four inputs (added in #858); this PR is the wrapper change that exposes them. PR-triggered runs continue using today's hardcoded values via explicit `github.event_name == 'pull_request'` gates.

## What changes

```yaml
workflow_dispatch.inputs:
  num_workers:    string,  default ''     # empty = use config's num_workers
  tail:           boolean, default true   # passes --tail/--no-tail
  local_dispatch: boolean, default false  # passes --local
  api_server:     string,  default ''     # passes --api-server <url>
```

Per-row knob precedence in `generate-launcher`:

- `num_workers` — `skypilot-local` always passes `1` (#843); other providers honor the dispatch input (empty → config value).
- `local` — `skypilot-local` always passes `true` (#841); other providers honor `inputs.local_dispatch` on dispatch.
- `tail` / `api_server` — flow through unchanged.

## Test plan

- [x] `make format` clean
- [x] `gha-workflow-validator` 0 failures (25 checks; same warning class as the other two reusables — info-only secret existence + reusable-call timeout-minutes false-positives)
- [ ] PR-trigger CI run — confirms PR runs still produce the same `skypilot-local, tail=true, local=true, num_workers=1` shape (auto-runs on push)
- [ ] Manual `workflow_dispatch` with overrides (e.g. `provider=runpod, num_workers=3, tail=false`) — confirms the reusable receives the overridden values (user-driven, billable)

Closes #861

### DIFF
--- a/.github/workflows/test-dataset-generation.yml
+++ b/.github/workflows/test-dataset-generation.yml
@@ -364,10 +364,20 @@ jobs:
   # upload step both write to). `if: !cancelled()` decouples each provider's
   # validate cell from the OTHER provider's generate outcome — RunPod
   # transients won't skip OCI validation.
+  #
+  # Skip on workflow_dispatch with `tail: false` — without --tail the launcher
+  # exits after provisioning, so workers haven't actually rendered/uploaded
+  # shards yet when validate would run. PR-trigger always uses --tail and is
+  # always validated.
   validate:
     name: Validate ${{ matrix.provider }} dataset
     needs: [setup, generate-local, generate-launcher]
-    if: ${{ !cancelled() && needs.setup.outputs.providers != '[]' }}
+    if: >-
+      ${{
+        !cancelled()
+        && needs.setup.outputs.providers != '[]'
+        && (github.event_name != 'workflow_dispatch' || inputs.tail)
+      }}
     strategy:
       fail-fast: false
       matrix:
--- a/.github/workflows/test-dataset-generation.yml
+++ b/.github/workflows/test-dataset-generation.yml
@@ -74,6 +74,26 @@ on:
           - skypilot-local
           - runpod
           - oci
+      num_workers:
+        description: "Override --num-workers (empty = use the dataset config's value; 1 forced for skypilot-local)."
+        required: false
+        type: string
+        default: ""
+      tail:
+        description: "Pass --tail to the launcher (stream logs + unconditional cluster teardown)."
+        required: false
+        type: boolean
+        default: true
+      local_dispatch:
+        description: "Pass --local to the launcher (force local SDK dispatch; mutually exclusive with api_server). Implicit for skypilot-local."
+        required: false
+        type: boolean
+        default: false
+      api_server:
+        description: "Remote SkyPilot API server URL. Empty = inherit env. Mutually exclusive with local_dispatch."
+        required: false
+        type: string
+        default: ""
   pull_request:
     paths:
       - "src/data/vst/**"
@@ -302,9 +322,16 @@ jobs:
 
   # `skypilot-local` / `runpod` / `oci` rows: route through the official
   # launcher reusable. Cluster name includes `github.run_attempt` so re-running
-  # a failed job doesn't collide on the launcher's R2 spec key. The
-  # `--num-workers 1` override for skypilot-local mirrors today's hardcoded
-  # behavior — the single-node kind cluster can't host parallel workers (#843).
+  # a failed job doesn't collide on the launcher's R2 spec key.
+  #
+  # Knob precedence per row:
+  #   * num_workers — `skypilot-local` always passes `1` (#843; single-node
+  #     kind can't host parallel workers). Other providers pass the dispatch
+  #     input through (empty = use the dataset config's `num_workers`).
+  #   * local       — `skypilot-local` always passes `true` (#841; clears
+  #     inherited SKYPILOT_API_SERVER_ENDPOINT so this row can't accidentally
+  #     route remote). Other providers honor the dispatch input.
+  #   * tail / api_server — dispatch inputs flow through unchanged.
   generate-launcher:
     name: Run generate_dataset (${{ matrix.provider }})
     needs: setup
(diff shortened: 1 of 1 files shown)


ktinubu:
## Verification Results

Behavioral checks against PR branch HEAD `aec64ab`. Stacked on #863. The change is small (one workflow file, +34/-7), so the verification stays narrow. `make format` is in CI; not duplicated as a checkbox.

- [x] **The four new dispatch inputs (`num_workers`, `tail`, `local_dispatch`, `api_server`) are declared with the right types and defaults** *(Level 5 — parses the branch's YAML; can't query GitHub's live schema view pre-merge since `gh workflow view` reflects main, not this branch. Once merged, equivalent Level 4 check is `gh workflow view "Test Dataset Generation" --yaml | grep -A1 inputs`.)*
  ```bash
  $ python3 <<'PY'
  import yaml
  with open('.github/workflows/test-dataset-generation.yml') as f:
      wf = yaml.safe_load(f)
  inputs = wf[True]['workflow_dispatch']['inputs']
  for name in ['num_workers', 'tail', 'local_dispatch', 'api_server']:
      spec = inputs[name]
      print(f"  {name}: type={spec.get('type', 'string')} default={spec.get('default', '')!r}")
  PY
    num_workers: type=string default=''
    tail: type=boolean default=True
    local_dispatch: type=boolean default=False
    api_server: type=string default=''
  ```

- [x] **`num_workers` expression has the right three-branch shape** *(Level 5 — adversarial: protects against an empty-string vs literal "0" vs `null` regression for runpod/oci with no dispatch override. The shape itself is what determines whether the launcher receives the right thing — a Level 1 run only catches it via downstream failure.)*
  ```bash
  $ python3 -c "
  import yaml
  wf = yaml.safe_load(open('.github/workflows/test-dataset-generation.yml'))
  expr = wf['jobs']['generate-launcher']['with']['num_workers']
  print('Expression:', expr)
  assert \"matrix.provider == 'skypilot-local' && '1'\" in expr
  assert \"github.event_name == 'pull_request' && ''\" in expr
  assert \"inputs.num_workers || ''\" in expr
  print('Three-branch shape OK')
  "
  Expression: ${{ matrix.provider == 'skypilot-local' && '1' || (github.event_name == 'pull_request' && '' || (inputs.num_workers || '')) }}
  Three-branch shape OK
  ```

- [ ] **PR-trigger CI: `Test Dataset Generation` end-to-end on `skypilot-local` preserves today's behavior (`tail=true, local=true, num_workers=1`)** *(Level 1 — this is the real check: if the gate logic is wrong, the launcher would receive bad values and fail. Auto-running on this branch's push, in_progress at time of verification.)*
  ```bash
  $ gh run list --branch worktree-issue-861-knobs --limit 1 -w "Test Dataset Generation" --json status,url --jq '.[]'
  ```
  Watch from `gh run list --branch worktree-issue-861-knobs -w "Test Dataset Generation" --limit 1`.

- [ ] **Manual `workflow_dispatch` with overrides flows through to the reusable** *(Level 1 — only path that exercises the dispatch-input → reusable-input bridge; SKIP — billable, user-driven)*
  To verify before merge:
  ```bash
  gh workflow run "Test Dataset Generation" \
    --ref worktree-issue-861-knobs \
    -f provider=runpod \
    -f num_workers=3 \
    -f tail=false \
    -f local_dispatch=false
  gh run watch
  ```
  In the launcher step's logs, confirm `--num-workers 3` appears AND `--tail` does not. (The tricky case is `tail=false`: the GHA expression `github.event_name == 'pull_request' && true || inputs.tail` must resolve to `false` here, not `true`.)

- [ ] **Negative test: `local_dispatch=true` + non-empty `api_server` is rejected by the launcher** *(Level 1 adversarial — these are mutually exclusive at the launcher CLI level (`_apply_dispatch_mode` raises ClickException), but the dispatch input shape lets operators pass both. SKIP — billable, user-driven; reviewers may also choose to defer to the launcher's own enforcement.)*
  To verify before merge:
  ```bash
  gh workflow run "Test Dataset Generation" \
    --ref worktree-issue-861-knobs \
    -f provider=runpod \
    -f local_dispatch=true \
    -f api_server=https://example.invalid
  ```
  Expected: launcher step exits with a ClickException about the mutually-exclusive flags. If instead the launcher silently picks one, that's a real bug — the wrapper should pre-reject this in the workflow file or document the precedence.

ktinubu:
## Tail-false dispatch issue + fix

Your `workflow_dispatch` run [25531011612](https://github.com/tinaudio/synth-setter/actions/runs/25531011612) on `provider=runpod` came back with `validate-shard` failing on `shard-000001.h5: shard file not found`. Root cause traced to `tail=false` (which propagated through correctly — `TAIL_FLAG: ` was empty in the launcher's invocation, confirmed in the run logs):

- Without `--tail`, the launcher's detached mode exits after `sky.launch` returns a `job_id` per rank — provisioning is done but workers haven't started rendering yet, let alone uploaded shards.
- `validate-shard` then runs immediately, downloads the spec from R2 (which the launcher writes pre-fan-out, so it's there), and tries to fetch each shard from R2. Shards from ranks still spinning up don't exist yet → "shard file not found".

This isn't a wiring bug — your dispatch input flowed through faithfully and the launcher really did detach. It's a workflow design issue: `tail=false` semantically means "launch and walk away" but the validate downstream assumed work-completion.

Pushed [5b327a5](https://github.com/tinaudio/synth-setter/pull/864/commits/5b327a5) which adds an `if:` guard so validate skips on `workflow_dispatch` with `tail=false`. PR-trigger behavior unchanged (always `tail=true`). With this fix, `tail=false` dispatch becomes "launch and exit"; operators can run validate manually after the workers finish via `sky logs <cluster>` then `python -m pipeline.ci.validate_shard r2://<bucket>/skypilot-launcher-specs/<cluster>.json`.

Side note: that run also surfaced significant RunPod capacity churn (`runpod.error.QueryError: There are no longer any instances available...` across many regions before each rank finally landed). Independent of this PR, but you may want a SkyPilot retry-with-backoff knob in the launcher for cap-constrained runs (#843 territory).

PR #867, opened 2026-05-08
feat(workflows): make generate-dataset-shards/validate-dataset-shards directly dispatchable

### LINKED ISSUE #865: feat(workflows): make generate-dataset-shards/validate-dataset-shards directly dispatchable
## Summary

`generate-dataset-shards.yaml` and `validate-dataset-shards.yaml` (both extracted in #858) are `workflow_call` only — they can be invoked from `test-dataset-generation.yml` but operators can't dispatch them directly via `gh workflow run` or the GitHub Actions UI's "Run workflow" button. PR #864 surfaced launcher knobs on `test-dataset-generation.yml`'s dispatch surface, which is the smoke-test wrapper, not the official launcher entry point — so to fire the launcher with overrides you still have to go through the test wrapper's provider matrix.

This PR adds `workflow_dispatch` triggers to both reusables (alongside the existing `workflow_call`) so the official launcher and validate workflows can be triggered directly.

### `generate-dataset-shards.yaml`

Adds `workflow_dispatch` with the same input shape as `workflow_call`, plus dispatch-friendly defaults:

- `provider` (choice: skypilot-local, runpod, oci) — required
- `dataset_config` (string, default `configs/dataset/runpod-smoke-shard.yaml`)
- `image_tag` (string, default `dev-snapshot`)
- `cluster_name` (string, default `''` — autogenerated as `synth-setter-manual-{provider}-{run_id}-{run_attempt}` when empty)
- `num_workers` (string, default `''`)
- `tail` (boolean, default `true`)
- `api_server` (string, default `''`)
- `local` (boolean, default `false`)
- `artifact_name` (string, default `run-metadata`)

The `generate` job gains a small `compute-cluster-name` step that reads `inputs.cluster_name` if set, else synthesizes the manual default. Subsequent steps reference the resolved name.

### `validate-dataset-shards.yaml`

Adds `workflow_dispatch` with the same two inputs as `workflow_call` (`image_tag`, `spec_uri`), both required.

## Why now

You hit this trying to dispatch "Generate Dataset Shards" directly. The reusable-only shape was the right design for the launcher-vs-test-wrapper split (PR #858), but it left the launcher human-untriggerable. Both shapes can coexist on a single workflow file, so adding `workflow_dispatch` is purely additive — `test-dataset-generation.yml` keeps calling them via `workflow_call` unchanged.

## Out of scope

- Sharing input definitions between `workflow_call` and `workflow_dispatch` — GHA requires duplicating them. We accept the duplication.

Refs #861 (which surfaced the same knobs on the test wrapper).

ktinubu:
## Summary

`generate-dataset-shards.yaml` and `validate-dataset-shards.yaml` were `workflow_call` only after the #858 split, so the GitHub Actions UI listed them but `gh workflow run` / "Run workflow" rejected the dispatch. Adds `workflow_dispatch` triggers alongside the existing `workflow_call` on both files. PR-trigger and `test-dataset-generation.yml`'s `workflow_call` invocations are unchanged.

### `.github/workflows/generate-dataset-shards.yaml`

- New `workflow_dispatch` block mirroring the `workflow_call` inputs:
  - `provider` (choice: skypilot-local / runpod / oci)
  - `dataset_config` (default `configs/dataset/runpod-smoke-shard.yaml`)
  - `image_tag` (default `dev-snapshot`)
  - `cluster_name` (default `''` — autogenerated when empty, see below)
  - `num_workers`, `tail`, `api_server`, `local`, `artifact_name` (same defaults as workflow_call)
- New `compute cluster_name` step at the top of the `generate` job: if `inputs.cluster_name` is empty, synthesizes `synth-setter-manual-<provider>-<run_id>-<run_attempt>` (run_id + run_attempt prevent collisions on the launcher's R2 spec key when re-dispatching). All downstream steps reference `${{ steps.cluster.outputs.name }}` instead of `${{ inputs.cluster_name }}`.
- The `workflow_call` `cluster_name` input is also relaxed to optional (same `''` default → same autogen fallback). `test-dataset-generation.yml` still passes an explicit value, so its behavior is unchanged.

### `.github/workflows/validate-dataset-shards.yaml`

- New `workflow_dispatch` block:
  - `image_tag` (default `dev-snapshot`)
  - `spec_uri` (required, no default — operator pastes the `r2://` URI to validate after a manual launcher run)

## Test plan

- [x] `make format` clean
- [x] `gha-workflow-validator` 0 failures across both files (49 total checks; same warning class as the merged PRs — info-only secret existence + reusable-call timeout-minutes false-positives)
- [x] Confirmed both files now expose both event triggers:
  ```bash
  $ python3 -c "import yaml; ..."
  generate-dataset-shards.yaml: on keys = ['workflow_call', 'workflow_dispatch']
  validate-dataset-shards.yaml: on keys = ['workflow_call', 'workflow_dispatch']
  ```
- [ ] PR-trigger CI on this branch — confirms `test-dataset-generation.yml`'s `workflow_call` invocation still works after relaxing `cluster_name` to optional (auto-runs on push)
- [ ] Manual `workflow_dispatch` on `Generate Dataset Shards` (e.g. `provider=runpod` with no other overrides) — confirms the dispatch path is now reachable and the autogen cluster_name kicks in (user-driven, billable for runpod/oci)
- [ ] Manual `workflow_dispatch` on `Validate Dataset Shards` with a `spec_uri` from a previous run — confirms the validate dispatch works standalone

Closes #865


ktinubu:
## Verification Results

Behavioral checks against PR branch HEAD `522c913`. The PR's central promise is "the workflows are dispatchable" — best verified by Level 3 (actually triggering a dispatch). My token gets HTTP 403 on `gh workflow run` regardless of dispatchability, so the gold-standard Level 3 check is left for you. Other checks confirm the wiring.

- [x] **Both workflows declare `workflow_dispatch` alongside `workflow_call` on this branch's YAML** *(Level 5 — pre-merge, GitHub's `gh workflow view` reflects main, not the PR branch. Once merged, equivalent Level 4 check is `gh api repos/tinaudio/synth-setter/actions/workflows/<id>` returning the file with the dispatch event registered.)*
  ```bash
  $ python3 <<'PY'
  import yaml
  for f in ['.github/workflows/generate-dataset-shards.yaml', '.github/workflows/validate-dataset-shards.yaml']:
      wf = yaml.safe_load(open(f))
      on_keys = sorted((wf[True] if True in wf else wf['on']).keys())
      print(f"{f.split('/')[-1]}: on = {on_keys}")
      assert on_keys == ['workflow_call', 'workflow_dispatch']
  print("BOTH WIRED")
  PY
  generate-dataset-shards.yaml: on = ['workflow_call', 'workflow_dispatch']
  validate-dataset-shards.yaml: on = ['workflow_call', 'workflow_dispatch']
  BOTH WIRED
  ```

- [x] **`test-dataset-generation.yml`'s `workflow_call` invocation still passes an explicit non-empty `cluster_name`** *(Level 5 adversarial — relaxing the workflow_call input to optional could regress callers that didn't pass it. The wrapper is the only caller; verify it still passes a deterministic value so the autogen branch isn't accidentally taken.)*
  ```bash
  $ python3 <<'PY'
  import yaml
  wf = yaml.safe_load(open('.github/workflows/test-dataset-generation.yml'))
  expr = wf['jobs']['generate-launcher']['with']['cluster_name']
  print(f"cluster_name in test wrapper: {expr}")
  assert expr.strip()
  assert 'matrix.provider' in expr
  assert 'github.run_id' in expr
  print("PASS")
  PY
  cluster_name in test wrapper: synth-setter-smoke-${{ matrix.provider }}-${{ github.run_id }}-${{ github.run_attempt }}
  PASS
  ```

- [x] **Autogen `cluster_name` uses `GITHUB_RUN_ID` + `GITHUB_RUN_ATTEMPT` — re-running a failed dispatch produces a fresh name, no R2 spec-key collision** *(Level 6 adversarial — visual inspection of the bash. Verifying via Level 1 would mean dispatching twice and observing distinct cluster names; my token can't dispatch, so this stays at the bash-logic level.)*
  ```bash
  $ sed -n '/id: cluster/,/^      - name:/p' .github/workflows/generate-dataset-shards.yaml | head -13
        - id: cluster
          env:
            INPUT_CLUSTER_NAME: ${{ inputs.cluster_name }}
          run: |
            set -euo pipefail
            if [[ -n "$INPUT_CLUSTER_NAME" ]]; then
              CLUSTER_NAME="$INPUT_CLUSTER_NAME"
            else
              CLUSTER_NAME="synth-setter-manual-${PROVIDER}-${GITHUB_RUN_ID}-${GITHUB_RUN_ATTEMPT}"
            fi
            echo "name=${CLUSTER_NAME}" >> "${GITHUB_OUTPUT}"
            echo "cluster_name: ${CLUSTER_NAME}"
  ```
  Caller-supplied value wins (first branch); empty falls back to autogen with run_attempt → unique on retry.

- [ ] **PR-trigger CI: `Test Dataset Generation` end-to-end on `skypilot-local` proves the `workflow_call` path still works after relaxing `cluster_name` to optional** *(Level 1 — auto-running on push, in_progress at time of verification. If the autogen-vs-input branching broke the normal path, validate-shard would fail since the spec_uri the validate matrix constructs depends on the cluster_name passed through.)*
  ```bash
  $ gh run list --branch worktree-issue-865-dispatchable-launcher -w "Test Dataset Generation" --limit 1 --json status,url --jq '.[]'
  ```
  Watch via `gh run list --branch worktree-issue-865-dispatchable-launcher -w "Test Dataset Generation"`.

- [ ] **`gh workflow run "Generate Dataset Shards" --ref worktree-issue-865-dispatchable-launcher` is accepted by GitHub** *(Level 3 — the gold-standard check, but my token gets HTTP 403 regardless of dispatchability. SKIP — needs user trigger.)*
  To verify (skypilot-local is free, ~10 min):
  ```bash
  gh workflow run "Generate Dataset Shards" \
    --ref worktree-issue-865-dispatchable-launcher \
    -f provider=skypilot-local
  gh run watch
  ```
  Confirm in the run logs that the autogen `cluster_name` step prints something like `cluster_name: synth-setter-manual-skypilot-local-<run_id>-1`.

- [ ] **Manual dispatch with `cluster_name` explicitly set uses the supplied value, not autogen** *(Level 3 — same blocker; SKIP.)*
  To verify:
  ```bash
  gh workflow run "Generate Dataset Shards" \
    --ref worktree-issue-865-dispatchable-launcher \
    -f provider=skypilot-local \
    -f cluster_name=my-custom-cluster
  ```
  In the run logs, the autogen step should print `cluster_name: my-custom-cluster`.

PR #876, opened 2026-05-09
feat(pipeline): switch skypilot launcher to sky.jobs managed-jobs SDK

ktinubu:
## Summary

- Switch the launcher (`pipeline/entrypoints/skypilot_launch.py`) from cluster-level launches (`sky.launch` / `sky.down` / `sky.tail_logs`) to the managed-jobs SDK (`sky.jobs.launch` / `sky.jobs.cancel` / `sky.jobs.tail_logs`), so the controller owns provisioning, retries, and terminal-status compute release.
- `sky.jobs.launch`'s `stream_and_get` yields `(job_ids: list[int], handle)` for a single-Task launch; the launcher extracts `job_ids[0]` and treats an empty list as a half-submitted job (cancelled in `finally`).
- Detached-mode operator commands now point at `sky jobs logs --name <name>` / `sky jobs cancel --name <name>`.
- `--cluster-name` CLI flag and its semantics-as-an-identifier are preserved; under managed jobs the value is passed to `sky.jobs.launch(name=...)`. Workflow callers (`generate-dataset-shards.yaml`, `test-skypilot-debug.yml`) continue to work unchanged.
- Tests in `tests/pipeline/test_entrypoints/test_skypilot_launch.py` updated: `mock_sky.jobs.launch` / `.cancel` / `.tail_logs` route by `name=` kwarg; launch response shape `([job_id], handle)`.
- **CI debuggability:** add `Snapshot sky logs (skypilot-local row)` step to `.github/workflows/generate-dataset-shards.yaml` so when the launcher hangs (as it did on a prior run of this branch), `~/sky_logs/` and `~/.sky/api_server/clients/` (where `file_mounts.log`, `provision.log` etc. live) get rolled into the existing `test-run-metadata-skypilot-local` artifact via `if: always()`. Without this, GHA cancels the runner and those logs die with it.

## References

- SkyPilot managed-jobs SDK: https://docs.skypilot.co/en/stable/_modules/sky/jobs/client/sdk.html
- SkyPilot API reference: https://docs.skypilot.co/en/stable/reference/api.html

Closes #875
Closes #904

## Test plan

- [x] `make test` (454 quick tests pass)
- [x] `python -m pytest tests/pipeline/test_entrypoints/test_skypilot_launch.py` (80 launcher tests pass)
- [x] `make format` (ruff format / pyright / docformatter clean)
- [ ] `generate-dataset-shards` workflow on a smoke shard against runpod / oci / skypilot-local rows

Issue #904, opened 2026-05-11
bug(ci): skypilot-local launcher hang/cancel loses sky logs (file_mounts.log etc.)


ktinubu:
## Summary

When the `skypilot-local` row of `test-dataset-generation.yml` hangs and is cancelled by the 60-min `timeout-minutes`, the only sky-log breadcrumb available to a debugger is a one-line pointer the SDK prints to stdout:

```
✓ Synced file_mounts.  View logs: sky api logs -l sky-2026-05-09-17-08-20-432274/file_mounts.log
```

That log file lives on the runner under `~/.sky/api_server/clients/<user_hash>/sky_logs/sky-<timestamp>/file_mounts.log` and is destroyed with the runner on cancel — leaving no way to diagnose the hang.

## Repro

PR #876 (`worktree-switch-to-managed-jobs`) — failed run [25606637509 / job 75169619278](https://github.com/tinaudio/synth-setter/actions/runs/25606637509/job/75169619278).

The `Run launcher against local-template (skypilot-local row)` step ran for ~52 min then was cancelled. `generate.log` (the tee'd launcher stdout, the only artifact uploaded today) contains the SDK's `View logs: sky api logs -l ...` pointer but not the referenced file.

## Two on-disk log roots that need capture

1. `~/sky_logs/` — CLI invocation logs (`sky local up`, `sky local down`). `SKY_LOGS_DIRECTORY` constant in `sky/skylet/constants.py:27`.
2. `~/.sky/api_server/clients/<user_hash>/sky_logs/` — SDK client logs that `sky api logs -l <id>` resolves to. Local API server in SkyPilot 0.12 stores them here. `sky/server/common.py:70` + `:1074`.

## Acceptance

- After any `skypilot-local` row run (success / fail / cancel), the `test-run-metadata-skypilot-local` artifact contains `sky_logs/` and `sky_api_server_logs/` subtrees with the per-invocation `sky-<timestamp>/file_mounts.log` etc.
- `if: always()` semantics so the snapshot also fires on the GHA-timeout cancel path.

## Out of scope

- Runpod / oci rows. Their launcher runs inside `tinaudio/synth-setter:dev-snapshot` with `docker run --rm`; capturing those logs needs bind-mounts and is deferred. When `SKYPILOT_API_SERVER_ENDPOINT` is set (the runpod/oci default), api-server-side logs live on the remote API server anyway.

## Related

- Refs #876 — surfaced this gap.
[...]

PR #990, opened 2026-05-12
feat(pipeline)!: switch skypilot launcher to sky.jobs managed-jobs SDK

### LINKED ISSUE #875: feat(skypilot): switch launcher to sky.jobs managed-jobs SDK
The launcher (`pipeline/entrypoints/skypilot_launch.py`) currently provisions a SkyPilot cluster per rank via `sky.launch` + `sky.down` and uses `sky.tail_logs` for log streaming. This works but means we hand-roll cluster lifecycle (idle_minutes_to_autostop, finally-block teardown, half-provisioned cleanup) instead of letting SkyPilot's managed-jobs controller manage it.

Switch the SDK calls to `sky.jobs.launch` / `sky.jobs.tail_logs` / `sky.jobs.cancel` so the controller owns provisioning, retries, and terminal-status compute release.

## Scope

- `pipeline/entrypoints/skypilot_launch.py`:
  - `sky.launch(task, cluster_name=..., idle_minutes_to_autostop=5, down=True)` → `sky.jobs.launch(task, name=...)`
  - `sky.stream_and_get(launch_request_id)` now returns `(job_ids: List[int], handle)` — extract `job_ids[0]` and validate
  - `sky.tail_logs(cluster_name=..., job_id=..., follow=True)` → `sky.jobs.tail_logs(name=..., job_id=..., follow=True)`
  - `sky.down(cluster)` → `sky.jobs.cancel(name=...)` (request-id pattern, then `stream_and_get`)
  - Detached-mode operator commands: `sky logs <cluster> <job_id>` / `sky down <cluster>` → `sky jobs logs --name <name>` / `sky jobs cancel --name <name>`
- `tests/pipeline/test_entrypoints/test_skypilot_launch.py`: update mocks/assertions accordingly.
- Module docstring + flag help text reflect managed-jobs semantics.

## Out of scope

- Workflow YAML changes (`generate-dataset-shards.yaml`, `test-skypilot-debug.yml`) — the `--cluster-name` CLI flag is preserved for backward compatibility and re-purposed as the managed-job name internally.
- Controller cloud configuration (whichever `~/.sky/config.yaml` `jobs.controller` points at) — same default behavior as cluster launches.

## References

- SkyPilot managed-jobs SDK: https://docs.skypilot.co/en/stable/_modules/sky/jobs/client/sdk.html
- SkyPilot API reference: https://docs.skypilot.co/en/stable/reference/api.html

ktinubu:
## Summary

Re-opened branch from closed PR #876 — that thread accumulated too many resolved review comments to navigate, so this PR is a fresh surface on the same branch (`worktree-switch-to-managed-jobs`). All commits, tests, and prior review responses are preserved.

Switches the launcher (`src/pipeline/skypilot_launch.py`) from cluster-level launches (`sky.launch` / `sky.down` / `sky.tail_logs`) to the managed-jobs SDK (`sky.jobs.launch` / `sky.jobs.cancel` / `sky.jobs.tail_logs`) so the SkyPilot controller owns provisioning, retries, and terminal-status compute release.

### What's in this branch

- **Core migration** — `sky.launch` → `sky.jobs.launch`; `cluster_name=` → `name=`; `stream_and_get` yields `(job_ids, handle)` and the launcher extracts `job_ids[0]`. `_cancel_job` replaces cluster teardown.
- **CLI rename** — `--cluster-name` → `--job-name` with a deprecation alias so existing dispatch callers keep working. Adds a `_JOB_NAME_RE` regex validator gated before any SDK call.
- **Diagnostic capture** — new `scripts/capture-skypilot-state.sh` snapshots kind/k8s controller state (pods, events, nodes, controller `provision.log`, worker describes) before `sky local down` reaps the cluster. Hardened to be bash-3.2 portable so it survives macOS CI runners.
- **Local CI controller shrink** — `jobs.controller.resources` is written inline by the skypilot-local workflow step so the controller fits on a single-node kind cluster (replaces the deleted `scripts/upsert_sky_config_key.py` / heredoc pair).
- **Test additions** — fan-out variants of `tail_logs` failure paths, `none_tuple` parametrization, OCI argv-leak multi-secret loop, malformed-config provider parametrization, env-var receipt assertion, rank → job_id ordering test, property-style cpus assertion, plus full coverage for the new capture script.
- **/repo-review-full follow-ups** — cluster→job rename drift in comments and test names, comment-hygiene fixes, shell-style BLOCKs (set -euo pipefail, readarray portability), action SHA pins, header compressions.

### Planned split

This branch will be carved into ~8 bite-sized PRs against main once the new review cycle settles. Slicing is documented in the working notes; the broad groups are: pure bug fixes → rename drift → upsert collapse → capture-script extraction → workflow hardening → comment hygiene → `--job-name` rename → managed-jobs core.

## Test plan

- [x] `make test-fast` — 590 passed, 5 skipped (local)
- [ ] CI on this PR all green (ubuntu, conda, macOS, mps, code-coverage, skypilot-local datagen)
- [ ] Skypilot-local kind CI green end-to-end with the new diagnostic capture step

Fixes #875
Refs #904


ktinubu:
## Verification Results

Behavioral checks I could run locally. Two Level-1 promises remain open in CI and stay unchecked until those jobs finish.

> **Update (post-`559e234`/`190be81`/`9b7fb02`):** macOS L1a now green; CI rollup now 17 SUCCESS / 1 IN_PROGRESS (skypilot-local kind smoke) / 1 SKIPPED / 1 NEUTRAL.

### Level 2 — Promise: launcher migrated to `sky.jobs.launch` and the new SDK contract holds

- [x] **Full PR-touched test suite passes** *(Level 2 — invokes the launcher + scripts the PR changes, against the actual SDK contract via `mock_sky`; promise is per-tool correctness, not end-to-end. End-to-end is L1b below.)*
  ```bash
  $ python -m pytest tests/pipeline/test_entrypoints/test_skypilot_launch.py tests/scripts/test_capture_skypilot_state.py tests/scripts/test_skypilot_write_provider_creds.py -q
  148 passed in 5.13s
  ```
  Now also covers `test_derived_default_name_with_bad_task_name_is_rejected` (path-traversal hardening of the derived `base_job_name`) and `test_no_resources_banner_does_not_become_pod_name` (kubectl empty-namespace banner filter).

### Level 2 — Promise: `--cluster-name` is a working deprecation alias for `--job-name`

- [x] **Alias still sets `name=` on `sky.jobs.launch`; new flag does not emit the deprecation warning** *(Level 2 — exercises argparse + the launcher main loop under mock; promise is deprecation behavior, not end-to-end.)*
  ```bash
  $ python -m pytest tests/pipeline/test_entrypoints/test_skypilot_launch.py -q -k "JobName or alias or cluster_name"
  17 passed, 94 deselected
  ```
  Includes `test_cluster_name_alias_still_sets_managed_job_name_kwarg`, `test_cluster_name_alias_emits_deprecation_warning_on_stderr`, `test_job_name_does_not_emit_deprecation_warning`.

### Level 2 (adversarial / implicit) — Promise: capture-script bash-3.2 portability has no unbound-variable surprise on empty kubectl output

- [x] **Empty kubectl pod list writes the sentinel file and exits 0 (not unbound-var-crash under `set -u`)** *(Level 2 — runs the actual `_capture_worker_pods` shell flow with a kubectl shim that emits no rows; promise is graceful empty-input handling under bash 3.2 semantics with the new while-read loop.)*
  ```bash
  $ python -m pytest tests/scripts/test_capture_skypilot_state.py::TestWorkerPodCapture::test_no_workers_writes_sentinel -v 2>&1 | tail -3
  PASSED
  $ bash -uo pipefail -c '
  worker_pods=()
  while IFS= read -r line; do
    [[ -n "$line" ]] && worker_pods+=("$line")
  done < <(:)
  echo "empty_n=${#worker_pods[@]}"'
  empty_n=0
  ```
  This is the exact failure mode that crashed all 11 capture-state tests on macos-latest in the prior CI run (`worker_pods: unbound variable`).

### Level 4 — CI rollup snapshot (post-fix-push)

- [ ] **All required CI checks green on PR #990** *(Level 4 — query live GitHub state; SKIP for now: 1 pending — `skypilot-local` kind smoke — at snapshot time)*
  ```
  pass     ubuntu 3.10/3.11 tests, macos 3.10 tests, MPS run_tests, conda tests,
           code-coverage, code-quality, CodeQL, Analyze actions/python,
           check-pr-metadata, Compute provider list + R2 bucket, discover (×2),
           auto-approve (×2)
  pending  Run generate_dataset (skypilot-local)
  skip     Run generate_dataset matrix; Auto-approve status
  ```
  To re-check manually:
  ```
  gh pr checks 990 --repo tinaudio/synth-setter
  ```

### Level 1

- [x] **macOS runner (real bash 3.2) runs capture-skypilot-state.sh tests green** *(Level 1 — promise is portability under the actual bash 3.2 interpreter Apple ships at `/bin/bash`; only the macos-latest runner exercises that. Local bash 5.1 doesn't reproduce.)*
  ```
  run_tests_macos (macos-latest, 3.10)    pass    2m48s
  https://github.com/tinaudio/synth-setter/actions/runs/25763081779/job/75668943338
  ```
  All 12 `tests/scripts/test_capture_skypilot_state.py` cases that failed on the prior CI run (`readarray: command not found` → `worker_pods: unbound variable`) now pass under Apple's `/bin/bash` 3.2.

- [ ] **skypilot-local kind CI runs end-to-end with `jobs.controller.resources` shrink and capture-state collection** *(Level 1 — full integration: kind up → controller pod scheduled with shrunk resources → managed-job launched → capture-state.sh runs before `sky local down` → metadata artifact contains controller `provision.log`. Only the skypilot-local datagen job exercises this. Currently pending.)*
  Re-check:
  ```
  gh pr checks 990 --repo tinaudio/synth-setter | grep skypilot-local
  ```

I'll close out this comment when the kind smoke lands.

tests/pipeline/test_entrypoints/test_skypilot_launch.py
```diff
@@ -1219,7 +1461,44 @@ def test_one_worker_failure_among_three_fails_launcher_after_full_teardown_under
...
+        than mask it as success, and every peer job — successes and the None-returner alike — must
+        still be cancelled.
+        """
+        self._setup_n_workers_mock(mock_sky, n=3, tail_rcs_by_name={"smoke-job-1-r1": None})
+
+        result = _invoke(
+            experiment,
+            template_yaml,
+            env_file,
+            "--cluster-name",
```
ktinubu: **[tdd-impl:warn]** `test_one_worker_tail_logs_returning_none_fails_launcher_after_full_cancel_under_tail` is the only test in `TestNumWorkersFanOut` that still uses `--cluster-name` (the deprecated alias) — every peer fan-out test in the same class was migrated to `--job-name`. Use `--job-name` here too for parity, and to avoid a stray deprecation-warning code path being silently exercised in this test's intended scope (which is the None-rc branch, not alias coverage).

src/pipeline/skypilot_launch.py
```diff
@@ -63,6 +75,14 @@
 # templates pass this verbatim into `git fetch + checkout` inside the container.
 _WORKER_GIT_REF_RE = re.compile(r"^[0-9a-f]{7,40}$")
 
+# Validates --job-name. The value is interpolated into a local filename under
+# $TMPDIR and into an R2 object key before SkyPilot itself ever sees it, so it
+# has to be path-separator-free and short. The pattern matches a kubernetes-
+# style label (alphanumeric start, alphanumerics/underscore/dash, ≤63 chars),
+# which is also a strict subset of what SkyPilot's own cluster-name validation
+# accepts.
```
ktinubu: **[code-health:warn]** The block-comment above `_JOB_NAME_RE` (lines 78–83) is a 6-line essay restating that the value is interpolated into a filename and an R2 key. Per CLAUDE.md "Keep comments terse — typically one short line" (and the explicit example warning against multi-sentence inline rationale), shorten this to a one-line pointer or move the long-form rationale to a docstring on the validation site in `main()`. Same pattern applies above the cred-bootstrap gate at lines 580–587.

src/pipeline/skypilot_launch.py
```diff
@@ -340,12 +360,29 @@ def upload_spec_to_r2(spec: DatasetSpec, cluster_name: str) -> str:
             local_path,
             rclone_dest,
         ]
-        subprocess.check_call(args)  # noqa: S603 — args from validated spec/cluster_name
+        subprocess.check_call(args)  # noqa: S603 — local tempfile path + R2 dest from job_name validated against _JOB_NAME_RE in main()
```
ktinubu: **[synth-setter:warn]** `subprocess.check_call(args)  # noqa: S603 — local tempfile path + R2 dest from job_name validated against _JOB_NAME_RE in main()` is 138 characters and embeds the safety rationale inline with the suppression. Hoist the rationale to a comment line above the call (Ruff will still honor the bare `# noqa: S603`), so the line itself fits the project's 99-char target and the rationale is readable instead of trailing off the right margin.

src/pipeline/skypilot_launch.py
```diff
@@ -528,45 +579,50 @@ def main(
     # Runs BEFORE `upload_spec_to_r2` so a bootstrap failure (e.g. missing
     # provider env) fails the launcher fast without polluting R2 with a spec
     # that no worker will ever consume.
+    #
+    # Local provider (kind) needs no compute creds and the CI workflow writes
+    # the managed-jobs controller-resource shrink directly — skip the bootstrap.
     provider = _detect_provider(template_path)
-    _run_cred_bootstrap(provider=provider, env_file_path=env_file_path)
+    if provider != "local":
+        _run_cred_bootstrap(provider=provider, env_file_path=env_file_path)
```
ktinubu, line 584: **[code-health:warn]** The pre-cred-bootstrap comment block (lines 573–584) accumulates five separate paragraphs of context (env propagation, no-leak guarantee, ordering-vs-upload, the local-skip rationale). Per CLAUDE.md, anything that needs more than ~2 lines belongs in a GitHub issue with a one-line in-source pointer. Recommend: keep the local-skip one-liner (`# kind: no compute creds; controller-resource shrink is written by CI workflow — see #876.`) and drop the rest into the function docstring or the design doc.

src/pipeline/skypilot_launch.py
```diff
@@ -528,45 +579,50 @@ def main(
...
-    # `sky.check.check` in-process before launch is the documented workaround
-    # (test-skypilot-local.yml). RunPod/OCI source creds from disk on every launch.
+    # `sky.jobs.launch` (like `sky.launch`) raises NoCloudAccessError on a fresh
+    # runner. Calling `sky.check.check` in-process before launch is the documented
+    # workaround (test-skypilot-local.yml). RunPod/OCI source creds from disk on every launch.
     if provider == "local":
         # Deferred so non-kubernetes runs (RunPod / OCI) skip the submodule import entirely.
         import sky.check
 
         sky.check.check(clouds=["kubernetes"], quiet=False)
```
ktinubu, line 602: **[ml-pipeline:warn]** The deferred `import sky.check` claims to skip the submodule import entirely for non-kubernetes runs (line 602 comment), but `sky.jobs` is unconditionally imported at module top (line 53), so the runtime saving is limited to `sky.check` only. Either drop the "entirely" claim or move the `sky.jobs` import behind the same gate. Minor docstring-accuracy nit.

src/pipeline/skypilot_launch.py
```diff
@@ -384,18 +421,25 @@ def upload_spec_to_r2(spec: DatasetSpec, cluster_name: str) -> str:
...
 )
 @click.option(
+    "--job-name",
     "--cluster-name",
+    "job_name",
     type=str,
     default=None,
-    help="SkyPilot cluster name (default: synth-setter-smoke-<config_id[:8]>).",
+    help=(
+        "Managed-job name (used both as the SkyPilot job identifier and as the R2 key "
```
ktinubu, line 428: **[python-style:warn]** The `--job-name` `--help` text renders the default as `synth-setter-smoke-<spec.task_name[:8]>` — the angle-bracket placeholder spells out a Python slice expression, which is confusing in a CLI `--help` listing where the reader can't run that expression. Suggest `(default: synth-setter-smoke-<first 8 chars of spec.task_name>)` or `(default: derived from spec.task_name)`.

ktinubu (commented):
Multi-skill review of PR #990 — 7 applicable plugin checklists (code-health, synth-setter-project-standards, python-style, tdd-implementation, shell-style, gha-workflow-validator, ml-data-pipeline). Findings posted below as individual unresolved inline threads.

Note: this run was performed sequentially inside the orchestrator (no Task/Agent tool available in this environment) rather than via the prescribed parallel fan-out, so each checklist was applied by the orchestrator itself reading the diff against the published checklist standards. ml-test was not run as a dedicated pass — relevant findings on the test files surfaced under tdd-implementation and code-health instead.

Overall the migration from `sky.launch` (cluster-mode) to `sky.jobs.launch` (managed-jobs) is thorough and well-tested: every behavioral seam has a dedicated test, the deprecated `--cluster-name` alias is preserved with a stderr deprecation, and `_JOB_NAME_RE` rejects path-traversal / shell-metachar inputs before any I/O. No blocking issues found — only nits around comment hygiene (CLAUDE.md "keep comments terse" rule), one stale `<cluster>` placeholder in a workflow comment, and a single fan-out test that uses the deprecated alias inconsistently with its peers.

.github/workflows/generate-dataset-shards.yaml
```diff
@@ -211,14 +214,40 @@ jobs:
...
       # skypilot-local caller passes it `true` to clear any inherited
       # SKYPILOT_API_SERVER_ENDPOINT. `num_workers: '1'` overrides the config
       # because the kind cluster can't host parallel workers (#843). The
       # argument array is built conditionally so optional flags don't expand
       # to empty strings.
+      # 20-min step cap: if the launcher (sky.jobs SDK) hangs, fail-fast and
+      # let the `if: always()` `Snapshot sky logs` step run, instead of
+      # burning the full 60-min job timeout before logs are uploaded.
+      # NOTE: step-level timeout sends SIGKILL — Python `finally` blocks in the
+      # launcher are bypassed. Operators may need `sky jobs cancel --name <cluster>`
```
ktinubu: **[synth-setter:warn]** Stale terminology: the comment says `sky jobs cancel --name <cluster>` but everywhere else in this PR the user-facing term migrated from "cluster" to "job". Replace `<cluster>` with `<job-name>` (or `<name>`) for consistency with the launcher's renamed flag and the operator-facing strings printed in `_run_workers_detached`.

scripts/capture-skypilot-state.sh
```diff
@@ -0,0 +1,125 @@
...
+  done
+}
+
+# `ray-node` is the SkyPilot 0.12 managed-jobs-controller pod's container name.
+# The `bash -lc` payload tails the latest sky_logs/sky-* directory's provision.log
+# — that's where SkyPilot writes the per-retry scheduling decisions.
+_capture_controller_pod() {
+  local controller_line controller_ns controller_pod
+  controller_line="$(kubectl get pods -A --no-headers 2>/dev/null \
+    | grep sky-jobs-controller | head -n1)"
```
ktinubu: **[shell-style:warn]** `grep sky-jobs-controller | head -n1` matches the substring anywhere in the kubectl output line, not just at the pod-name column. A pod or namespace containing that substring elsewhere — or a future `-o wide`/format change that leaks the literal into another column — would falsely match. Prefer column-aware matching, e.g. `awk '$2 ~ /^sky-jobs-controller/ {print; exit}'` — same defensive style already used in `_capture_worker_pods` above.

scripts/capture-skypilot-state.sh
```diff
@@ -0,0 +1,125 @@
...
+  if [[ -z "${controller_line}" ]]; then
+    echo "no sky-jobs-controller pod found" \
+      > "${OUT_DIR}/describe-controller.txt"
+    return 0
+  fi
+  controller_ns="$(awk '{print $1}' <<<"${controller_line}")"
+  controller_pod="$(awk '{print $2}' <<<"${controller_line}")"
```
ktinubu: **[shell-style:warn]** Two consecutive `awk '{print $N}' <<<"${controller_line}"` calls re-parse the same string. A single `read -r controller_ns controller_pod _ <<<"${controller_line}"` parses once, is faster, and reads more clearly. Minor cleanup.

scripts/capture-skypilot-state.sh
```diff
@@ -0,0 +1,125 @@
...
+  kubectl describe pod -n "${controller_ns}" "${controller_pod}" \
+    > "${OUT_DIR}/describe-controller.txt" 2>&1 || true
+  kubectl logs -n "${controller_ns}" "${controller_pod}" \
+    > "${OUT_DIR}/logs-controller.txt" 2>&1 || true
+  kubectl logs -n "${controller_ns}" "${controller_pod}" --previous \
+    > "${OUT_DIR}/logs-controller-previous.txt" 2>&1 || true
+  # shellcheck disable=SC2016
+  # ${HOME} and ${latest} must expand inside the controller pod, not the runner.
+  kubectl exec -n "${controller_ns}" "${controller_pod}" -c ray-node -- \
+    bash -lc 'set -euo pipefail; latest="$(ls -td "${HOME}/sky_logs/sky-"*/ 2>/dev/null | head -n1)"; if [[ -n "${latest}" ]]; then echo "=== ${latest}provision.log ==="; tail -400 "${latest}provision.log" 2>/dev/null; fi' \
```
ktinubu: **[shell-style:warn]** The `bash -lc '...'` payload on this line is a ~200-char single-line script containing `set -euo pipefail`, a command substitution, a conditional, and a `tail` call. Multi-statement bash one-liners are hard to read and hard to diff. Consider extracting the payload into a small helper script that the controller pod can run, or split at the `;` boundaries onto separate lines inside the `'...'` literal. Also: `bash -lc` runs a *login* shell which loads profile/rc files unnecessarily — `bash -c` would suffice here.

src/pipeline/skypilot_launch.py
```diff
@@ -586,81 +642,70 @@ def _override_image_id(task: sky.Task, worker_image: str) -> None:
...
-    mutated = False
     for res in task.resources:
         if isinstance(res.cloud, OCI):
             new_resources.append(res)
             continue
         new_resources.append(res.copy(image_id=docker_ref))
-        mutated = True
-    if mutated:
-        task.set_resources(type(task.resources)(new_resources))
+    task.set_resources(type(task.resources)(new_resources))
```
ktinubu, line 654: **[code-health:warn]** `_override_image_id` was refactored to always call `task.set_resources(...)` (the old `mutated` flag was removed). The function's docstring still reads as if OCI-only Tasks are pass-through ("OCI Resources are left untouched here"), which is true for the entries but the *Task's resources collection* is now always re-assigned — observable via `task.set_resources.called` in tests. Worth one extra sentence in the docstring noting that the resources list is rebuilt even when no entry was mutated, so callers (and mock-based test readers) aren't surprised by the call-count.

tests/pipeline/test_entrypoints/test_skypilot_launch.py
ktinubu: Fixed in 559e234 — the fan-out test now uses `--job-name` like its peers; `--cluster-name` alias coverage stays isolated to `TestJobNameAlias`.

src/pipeline/skypilot_launch.py
ktinubu: Fixed in 559e234 — trimmed to a two-line summary referencing #876.

src/pipeline/skypilot_launch.py
ktinubu: Fixed in 559e234 — moved the rationale to a block comment above; the `noqa: S603` line now stands alone.

src/pipeline/skypilot_launch.py
ktinubu: Fixed in 559e234 — reduced to a single one-line pointer.

src/pipeline/skypilot_launch.py
ktinubu: Fixed in 559e234 — clarified the comment: deferral skips the `sky.check` submodule import for non-kubernetes runs (not the SDK itself).

src/pipeline/skypilot_launch.py
ktinubu: Fixed in 559e234 — placeholder rendered as 'first 8 chars of spec.task_name' so the help text isn't misread as a literal bracketed slice.

.github/workflows/generate-dataset-shards.yaml
ktinubu: Fixed in 559e234 — placeholder is now `<job-name>` matching the `--cluster-name` → `--job-name` rename.

scripts/capture-skypilot-state.sh
ktinubu: Fixed in 559e234 — switched to column-aware `awk '$2 ~ /^sky-jobs-controller/ {print; exit}'` so a sky-jobs-controller substring in any other column can't false-match.

scripts/capture-skypilot-state.sh
ktinubu: Fixed in 559e234 — single `read -r controller_ns controller_pod _ <<<"${controller_line}"` replaces the two `awk` re-parses.

scripts/capture-skypilot-state.sh
ktinubu: Fixed in 559e234 — broken into a multi-line `bash -c` payload (no profile load needed); rationale moved to a comment block above the `kubectl exec` call.

src/pipeline/skypilot_launch.py
ktinubu: Fixed in 559e234 — docstring now explicitly notes the unconditional `task.set_resources(...)` rebuild so mock-based test readers expect that call regardless of provider mix.

src/pipeline/skypilot_launch.py
```diff
@@ -507,13 +560,11 @@ def main(
...
+    base_job_name = job_name or f"synth-setter-smoke-{spec.task_name[:8]}"
 
-    # Per-cluster filename so parallel launches (CI matrix, local dev concurrent with CI on
+    # Per-job filename so parallel launches (CI matrix, local dev concurrent with CI on
     # the same host) don't clobber one another's spec.
-    local_spec_path = (
-        spec_out or LOCAL_SPEC_DIR / f"skypilot-launch-smoke-{base_cluster_name}.json"
-    )
+    local_spec_path = spec_out or LOCAL_SPEC_DIR / f"skypilot-launch-smoke-{base_job_name}.json"
     local_spec_path.parent.mkdir(parents=True, exist_ok=True)
```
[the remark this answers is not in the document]
ktinubu: Fixed in 559e234 — derived `base_job_name` is re-validated against `_JOB_NAME_RE` before any IO; new test `test_derived_default_name_with_bad_task_name_is_rejected` covers the rejection-before-launch contract (no spec write, no `sky.jobs.launch` call).

scripts/skypilot_write_provider_creds.sh
```diff
@@ -1,40 +1,20 @@
-#!/usr/bin/env bash
-# Bootstrap SkyPilot credentials before `sky check` / `sky.launch`.
+#!/bin/bash
+# Bootstrap SkyPilot R2 + per-provider creds to disk before `sky check` /
+# `sky.launch`. No stdout output by design (safe for tee'd contexts).
```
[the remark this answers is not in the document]
ktinubu: Fixed in 559e234 — header now references `sky.jobs.launch` (managed-jobs SDK).

configs/compute/local-template.yaml
```diff
@@ -6,9 +6,13 @@
...
-# `--provider local`, which writes only the R2 cred files — kind needs no
-# compute provider auth. R2 still ships the materialized spec
+# The launcher detects `cloud: kubernetes` and SKIPS its cred bootstrap —
+# kind needs no compute creds (R2 env vars are forwarded into the pod via
+# task.update_envs) and the CI workflow writes the managed-jobs controller-
+# resource shrink directly to ~/.sky/config.yaml before launch (the SkyPilot
+# defaults of cpus=4+/mem=4x/disk_size=50 don't fit on GHA's kind cluster,
+# and k8s rejects disk_size). See PR #876.
+# R2 still ships the materialized spec
 # (`r2://<bucket>/skypilot-launcher-specs/<cluster>.json`).
```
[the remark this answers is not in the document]
ktinubu: Fixed in 559e234 — example reads `<job>.json` to match the launcher's job-name keying for the per-launch spec upload.

scripts/capture-skypilot-state.sh
```diff
@@ -0,0 +1,125 @@
...
+# Filters workers in the `default` namespace; sky-jobs-controller may live in
+# the `skypilot` namespace and is captured separately by _capture_controller_pod.
+_capture_worker_pods() {
+  # Portable to bash 3.2 (macOS /bin/bash) — `readarray` is bash 4+ only.
+  local worker_pods=()
+  local line
+  while IFS= read -r line; do
+    [[ -n "${line}" ]] && worker_pods+=("${line}")
+  done < <(kubectl get pods -n default --no-headers 2>/dev/null \
+    | awk '$1 !~ /^sky-jobs-controller/ {print $1}')
```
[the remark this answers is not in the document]
ktinubu: Fixed in 559e234 — script now filters `/^No resources/` before the awk pod-name pick; new test `test_no_resources_banner_does_not_become_pod_name` regression-pins it.

docs/reference/github-actions.md
```diff
@@ -24,16 +24,16 @@ For GitHub Actions concepts, see [GitHub's docs](https://docs.github.com/en/acti
...
+| `generate-dataset-shards`   | Reusable + dispatchable workflow: invokes `src.pipeline.skypilot_launch` for one provider (skypilot-local kind / runpod / oci); the launcher uploads the materialized spec to R2 and the workflow emits a `spec_uri` output that downstream callers (`validate-dataset-shards`) consume. A run-metadata artifact is also uploaded for debugging. Consumed by `test-dataset-generation`; also fireable directly via `gh workflow run` or the Actions UI (autogenerates a manual cluster name when `cluster_name` is left blank). | Mounts PR code as a volume into the container for runpod/oci rows. See [Mount-as-volume pattern](#mount-as-volume-pattern).                                                                                                                    |
```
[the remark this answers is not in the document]
ktinubu: Fixed in 9b7fb02 — `generate-dataset-shards` row now reads "autogenerates a managed-job name when the `cluster_name` workflow input — kept for backward compat — is left blank", so both facts (the value is now the managed-job name, the input is still `cluster_name` for backward compat) are visible.

docs/design/skypilot-compute-integration.md
```diff
@@ -86,35 +86,36 @@ compute_config: null
...
 ```
 configs/compute/
 ├── runpod-template.yaml      # RunPod GPU (primary smoke target)
-└── oci-cpu-template.yaml     # OCI x86 CPU Flex (second smoke target)
+├── oci-cpu-template.yaml     # OCI x86 CPU Flex (second smoke target)
+└── local-template.yaml       # kind/kubernetes (sky local up; CI smoke only — see the YAML header for the CI-only resource shrink, PR #876)
 ```
 
-Both share the launcher (`pipeline/entrypoints/skypilot_launch.py`),
+Both share the launcher (`src/pipeline/skypilot_launch.py`),
```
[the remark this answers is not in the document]
ktinubu: Fixed in 9b7fb02 — paragraph now starts with "All three share the launcher…" to match the three-template list above.

ktinubu:
Closed in favor of #992 — same  branch content, but rewritten on top of `origin/main` as 3 self-contained commits (one per planned split slice) so each commit independently passes `make test-fast`. The original branch is preserved at `worktree-switch-to-managed-jobs` for reference.

ktinubu:
Closed in favor of #992 — same `worktree-switch-to-managed-jobs` branch content, but rewritten on top of `origin/main` as 3 self-contained commits (one per planned split slice) so each commit independently passes `make test-fast`. The original branch is preserved at `worktree-switch-to-managed-jobs` for reference.

Issue #784, opened 2026-05-03
Phase: migrate to PEP src-layout (src/synth_setter/), unify pipeline under one package, depopulate scripts/


ktinubu:
## Context

The repo's current top-level Python layout is non-standard and creates several concrete problems that compound over time. Surfaced during a layout audit on 2026-05-03.

### Current state (facts)

- `src/` is a Python package literally named `src`. Code reads `from src.train import main`, which is uninformative. Internal references to `from src.*` exist in 52+ places.
- The distribution name in `pyproject.toml` is `synth-permutations`, but **no `synth_permutations` (or `synth_setter`) package exists anywhere on disk** — `setup.py`'s `find_packages()` silently registers `src`, `pipeline`, and (until [#779](https://github.com/tinaudio/synth-setter/issues/779) / [#780](https://github.com/tinaudio/synth-setter/pull/780)) `scripts` as three independent installable top-level packages.
- `pipeline/` is a sibling of `src/` at the repo root, with only 2 cross-imports back into `src.data.vst` (param specs + renderer version). There's no architectural reason it can't live under one umbrella package — the asymmetry (`from src.X` vs `from pipeline.X`) reads as accidental.
- `scripts/` contains 15 .py + 7 .sh files at the root, no `__init__.py`. CLAUDE.md guidance banning new bare-root additions is being shaped in [#779](https://github.com/tinaudio/synth-setter/issues/779), but the existing residents need real homes.
- Code "works" because the repo root happens to be on `sys.path` (pytest discovers it; `python -m foo` from cwd=root inherits it). There is no real install step — CI, RunPod images, and dev shells each end up with subtly different path setups, and `ImportError` only surfaces in the environment where it bites.

### Why this matters

- **Readability tax.** Every traceback, every import, every code review reads `from src.X` and the reader has to remember "src is the package, not a container."
- **Packaging fiction.** The project is *not* installable as `synth-permutations` in any meaningful sense — it ships three packages (`src`, `pipeline`, `scripts`) under a name that matches none of them. Anyone trying to vendor or publish hits this immediately.
- **Drift between dev and prod.** Scripts that work locally because of cwd accidents fail in Docker/RunPod with `ImportError`. The class of bug exists today and has bitten before.
[...]

PR #744, opened 2026-05-01
revert: "build(deps): migrate lightning to pytorch_lightning"

ktinubu:
## Summary

Pure revert of 60fd65cebaab6a1d76caa82a345b17a5474d5beb (the migration commit on [#721](https://github.com/tinaudio/synth-setter/pull/721)). Restores the `lightning` umbrella package and the `lightning.pytorch.*` import paths.

This reverts only the migration commit — the other changes from #721 (Docker/SkyPilot deps) are not touched.

### Motivation

Config-comparison tests (the drift harness from [#679](https://github.com/tinaudio/synth-setter/pull/679)) are failing because the baseline checkout pre-#721 still imports `lightning`, but the current environment only has `pytorch_lightning` installed — so the baseline can't be instantiated. Failing job: https://github.com/tinaudio/synth-setter/actions/runs/25236756046/job/74004503107

Reverting back to `lightning` lets baseline configs resolve again. PyPI's quarantine on the `lightning` umbrella that triggered #721 has been lifted (`lightning 2.6.1` resolves cleanly via `uv pip compile`).

## Test plan

- [ ] CI green on the revert branch (depends on `lightning>=2.6.0` resolving on PyPI again).
- [ ] `pytest -n auto -k "not slow"` passes locally on the restored imports.
- [ ] Config-drift / fixture-baseline tests pass with the reverted imports.

Refs #534

PR #991, opened 2026-05-12
refactor(layout)!: nest src/* under synth_setter/, declare console scripts

### LINKED ISSUE #989: Task: Phase 2 — move src/{data,models,utils,metrics,train,eval,generate_dataset} under src/synth_setter/
Phase 2 of the PEP src-layout migration (parent: #784).

## Scope

Hoist `src/` Python code into `src/synth_setter/` with no `ml/` wrapper. Top-level CLIs land under `synth_setter/cli/` as fat `@hydra.main` entrypoints. The `src/pipeline/` tree is **out of scope** for Phase 2 — it moves in Phase 3.

### File moves

| From | To |
|---|---|
| `src/data/` | `src/synth_setter/data/` |
| `src/models/` | `src/synth_setter/models/` |
| `src/utils/` | `src/synth_setter/utils/` |
| `src/metrics.py` | `src/synth_setter/metrics.py` |
| `src/train.py` | `src/synth_setter/cli/train.py` |
| `src/eval.py` | `src/synth_setter/cli/eval.py` |
| `src/generate_dataset.py` | `src/synth_setter/cli/generate_dataset.py` |

`src/pipeline/` is **NOT moved** in this PR — Phase 3. After this PR, `src/` contains only `pipeline/` and `__init__.py` (kept until Phase 3 moves pipeline and Phase 5 cleans up).

### Sweeps (current scope: 60 files, 140 import references, 47 `_target_:` references)

- `from src.{data,models,utils,metrics,train,eval,generate_dataset}` → `from synth_setter.{data,models,utils,metrics,cli.train,cli.eval,cli.generate_dataset}` (and `import src.X` likewise) across `tests/`, `scripts/`, `src/pipeline/` (cross-references), and any other caller.
- `from src.pipeline.*` → unchanged (Phase 3 territory).
- `_target_: src.X` (47 references in `configs/**/*.yaml`) → `_target_: synth_setter.X`. Where the source file moved into `cli/`, the new `_target_` is `synth_setter.cli.X`. Most `_target_` references point to classes in `data/`, `models/`, `utils/`, not CLIs — review each rewrite.

### `[project.scripts]` declarations

Add to `pyproject.toml`:

```toml
[project.scripts]
synth-setter-train             = "synth_setter.cli.train:main"
synth-setter-eval              = "synth_setter.cli.eval:main"
synth-setter-generate-dataset  = "synth_setter.cli.generate_dataset:main"
```

Remove the legacy `setup.py` `console_scripts` entries (`train_command`, `eval_command`) — they're superseded. (Full `setup.py` deletion happens in Phase 5.)

Update `[tool.setuptools.packages]` in `pyproject.toml` to enumerate the new structure: keep `src.pipeline*` enumerated (they're moved in Phase 3); drop `src.data`, `src.models`, `src.utils`, `src.synth_setter` (since `synth_setter` is now under `src/synth_setter/` and its sub-packages are discovered via the `synth_setter` package root).

### Other touched files

- `Dockerfile` — any `python -m src.X` invocations become `python -m synth_setter.cli.X` (or the new `synth-setter-*` console scripts where appropriate).
- `configs/compute/runpod-template.yaml` and similar — same treatment for `run:` blocks (mind the CLAUDE.md rule against comments inside YAML block-scalars).
- `.github/workflows/*.yaml` — any `python -m src.X` references.
- `CLAUDE.md`, `README.md`, `docs/` — module references in prose.
- `pyrightconfig.json`, `.pre-commit-config.yaml`, `mutmut` config in `pyproject.toml` — path patterns.

### Hydra CLI surface

`cli/train.py`, `cli/eval.py`, `cli/generate_dataset.py` keep their bodies fat (the `@hydra.main`-decorated `main` IS the implementation, per fairseq/demucs/audiocraft convention). No logic-vs-CLI split in this phase.

### Test relocation

Tests under `tests/` that mirror moved source files (`tests/test_train.py`, `tests/data/`, `tests/models/`, etc.) keep their current locations for this PR — test reorganization is folded into Phase 4 when `tests/scripts/` moves too.

## Verification

- `make test-fast` clean.
- **`pytest tests/test_compare_baseline_configs.py -v` passes — mandatory hard gate.** Expect `_target_:` rewrites to surface in the resolved YAMLs; bump `FIXTURE_BASELINE` and/or `MODEL_BASELINE` in the same PR to a SHA on this branch that includes the move, and document the resolved-YAML diff in the PR description.
- `rg -n "from src\.|^import src(\.|$|\s)" --type py` returns 0 matches outside `src/pipeline/` (Phase 3 territory).
- `rg -n "_target_:\s*src\." configs/` returns 0 matches.
- `synth-setter-train --help`, `synth-setter-eval --help`, `synth-setter-generate-dataset --help` all exit 0 after `pip install -e .` in a fresh venv.
- `python -m synth_setter.cli.train --help` etc. also exits 0 (both surfaces work).

## Out of scope

- `src/pipeline/` move (Phase 3).
- `scripts/` depopulation (Phase 4).
- `setup.py` deletion (Phase 5).
- Test directory reorganization (Phase 4, alongside `tests/scripts/` moves).

Refs #784.

ktinubu:
## Summary

Phase 2 of the PEP src-layout migration. Hoists `src/{data,models,utils,metrics.py}` to `src/synth_setter/` and `src/{train,eval,generate_dataset}.py` to `src/synth_setter/cli/`. Declares `synth-setter-{train,eval,generate-dataset}` console scripts via `[project.scripts]` and removes the legacy `train_command` / `eval_command` setup.py entries.

Closes #989
Part of #784

### File moves (git mv — history preserved)

| From | To |
|---|---|
| `src/data/` | `src/synth_setter/data/` |
| `src/models/` | `src/synth_setter/models/` |
| `src/utils/` | `src/synth_setter/utils/` |
| `src/metrics.py` | `src/synth_setter/metrics.py` |
| `src/train.py` | `src/synth_setter/cli/train.py` |
| `src/eval.py` | `src/synth_setter/cli/eval.py` |
| `src/generate_dataset.py` | `src/synth_setter/cli/generate_dataset.py` |

New empty `src/synth_setter/cli/__init__.py` package marker. `src/pipeline/` is untouched — moves in Phase 3.

### Sweep counts

- **0** remaining non-pipeline `from src.{data,models,utils,metrics,train,eval,generate_dataset}` references (was 81).
- **0** remaining `_target_: src.X` references in `configs/**/*.yaml` (was 47).
- **0** remaining `python src/X.py` invocations in `jobs/`, `sweeps/`, `scripts/`, `.github/workflows/`, `docker/` (rewritten to `python -m synth_setter.cli.X`).
- All `from src.pipeline.*` imports are intentionally preserved — Phase 3 territory.

### pyproject.toml / setup.py changes

- Added `[project.scripts]` with three console scripts (`synth-setter-train`, `synth-setter-eval`, `synth-setter-generate-dataset`).
- Enumerated `[tool.setuptools].packages` instead of switching to `find`. Reasoning: `synth_setter` lives at `src/synth_setter/` (mapped via `package-dir`) but `pipeline` still lives at `src/pipeline/` for one more phase. A single `find` directive cannot cleanly discover both roots without re-exposing the legacy `src.data` / `src.models` / etc. names. We collapse to `find` once Phase 3 unifies the layout.
- Removed `console_scripts` `train_command` / `eval_command` from `setup.py`; the rest of `setup.py` survives until Phase 5.

### Hydra CLI

- Each CLI's `@hydra.main` `config_path` updated from `../configs` to `../../../configs` (one more level up from `src/synth_setter/cli/`).
- Each CLI keeps `if __name__ == "__main__": main()` so `python -m synth_setter.cli.X` continues to work alongside the new console scripts.
- `rootutils.setup_root(__file__, indicator=".project-root", pythonpath=True)` continues to resolve the repo root from the new location.

### Baseline bump

`MODEL_BASELINE` in `tests/test_compare_baseline_configs.py` bumped from `v0.0.0` (= `79552d2`) to this PR's first commit `4e08950`. Reason: the resolved Hydra YAML at `v0.0.0` contains `_target_: src.X` keys while the live tree contains `_target_: synth_setter.X` — a 44-case KOSC + 8-case SURGE failure, all of which are the intentional `_target_:` rewrite (no semantic drift).

`FIXTURE_BASELINE` is untouched. The synthetic-fixture scripts under `tests/fixtures/baseline_repo/` are self-contained and don't reference `src.X` at all.

Representative diff (from `configs/model/flow.yaml` at `v0.0.0` vs this branch):

```diff
-_target_: src.models.ksin_flow_matching_module.KSinFlowMatchingModule
+_target_: synth_setter.models.ksin_flow_matching_module.KSinFlowMatchingModule
 ...
 vector_field:
-  _target_: src.models.components.transformer.ApproxEquivTransformer
+  _target_: synth_setter.models.components.transformer.ApproxEquivTransformer
   projection:
-    _target_: src.models.components.transformer.LearntProjection
+    _target_: synth_setter.models.components.transformer.LearntProjection
```

Every `_target_:` rewrite is mechanical and value-preserving — no `_partial_`, args, or kwargs changed.

## Verification

| # | Gate | Status | Evidence |
|---|------|--------|----------|
| 1 | `make format` | PASS | All pre-commit hooks pass on the full tree (ruff, ruff-format, pyright, docformatter, interrogate, pydoclint, prettier, shellcheck, mdformat, codespell). |
| 2 | `pip install -e .` | PASS | Successfully installed `synth-setter-1.1.0` (editable wheel) from `pyproject.toml`. |
| 3 | `python -c "import synth_setter; print(synth_setter.__file__)"` | PASS | Resolves to `src/synth_setter/__init__.py`. |
| 4 | `synth-setter-{train,eval,generate-dataset} --hydra-help` | PASS | All three console scripts exit 0 with Hydra's flag listing. Note: `--help` (not `--hydra-help`) exits non-zero on `train.yaml` / `eval.yaml` / `dataset.yaml` because they declare `???` MISSING keys (`model: ???`, etc.) — this is pre-existing Hydra behavior, unchanged by the migration. The same exit code came from `python src/train.py --help` pre-migration. The module imports themselves resolve cleanly. |
| 5 | `python -m synth_setter.cli.{train,eval,generate_dataset} --hydra-help` | PASS | All three module forms exit 0 with Hydra's flag listing. Both surfaces work. |
| 6 | `rg -n "from src\.(data\|models\|utils\|metrics\|train\|eval\|generate_dataset)" --type py` | PASS | 0 matches. |
| 7 | `rg -n "_target_:\s*src\.(data\|models\|utils\|metrics\|train\|eval\|generate_dataset)" configs/` | PASS | 0 matches. |
| 8 | `make test-fast` | PASS | 556 passed, 5 skipped in 28.31s. |
| 9 | `pytest tests/test_compare_baseline_configs.py -v` | PASS (with baseline bump) | Fixture-baseline tests (17/17 non-slow) pass against the unchanged `FIXTURE_BASELINE`. The KOSC + SURGE train cases fail at the old `v0.0.0` baseline (44 KOSC + 8 SURGE failures, all `_target_:` mechanical drift) and pass after bumping `MODEL_BASELINE` to `4e08950`. Single-case spot-check at the new baseline: `test_kosc_train_configs_are_equal[4e08950_vs_live__kosc_train__task1]` PASSED in 9.83s. Full slow-suite run is on CI. |

## Out of scope

- `src/pipeline/` move under `synth_setter/` — **Phase 3** (#784).
- `scripts/` depopulation — **Phase 4** (#784).
- Full `setup.py` deletion — **Phase 5** (#784); this PR only removes the legacy `console_scripts` entries that are superseded by `[project.scripts]`.
- Test directory reorganization — **Phase 4**, alongside `tests/scripts/` moves.

## Notes

- All file moves used `git mv` so history is preserved (`git log --follow` traces continuously through the move).
- Per CLAUDE.md, this is a breaking change (`!`): external consumers importing `src.{data,models,utils,metrics,train,eval,generate_dataset}` and users relying on `train_command` / `eval_command` console scripts must update. The new module paths and `synth-setter-*` scripts are the replacements.
- Historical design docs under `docs/design/*-implementation-plan.md` and `docs/design/storage-provenance-spec.md` were intentionally not rewritten — they're point-in-time snapshots, not living references. Active docs (`README.md`, `CLAUDE.md`, `docs/getting-started.md`, `docs/glossary.md`, `docs/doc-map.yaml`, `docs/reference/*`, `docs/guides/surge-xt-interactive.md`, `docs/design/data-pipeline.md`, `docs/design/training-pipeline.md`, `docs/design/eval-pipeline.md`) were updated.


ktinubu:
## Verification Results

Each check below states the **promise** the PR makes, the **hierarchy level** (and why), the exact **command run**, and the **observed output**. All commands run on branch `refactor/phase-2-synth-setter-layout` at HEAD `303eee3`.

### Explicit promises

- [x] **`synth_setter` package resolves at the new PEP src-layout location** *(Level 2 — invoked python on the installed package; promise is "import synth_setter works post-install")*
  ```bash
  $ python -c "import synth_setter; print(synth_setter.__file__)"
  /home/build/synth-setter/.claude/worktrees/agent-a6f2a4c0d9f769e20/src/synth_setter/__init__.py
  ```
  Resolves under `src/synth_setter/` as required by the migration.

- [x] **All migrated submodules (`data`, `models`, `utils`, `metrics`, `cli.{train,eval,generate_dataset}`) are importable as `synth_setter.*`** *(Level 2 — actual imports + attribute access; promise is "the rewrite preserves every public import path")*
  ```bash
  $ python -c "from synth_setter.cli.train import main as t; from synth_setter.cli.eval import main as e; from synth_setter.cli.generate_dataset import main as g; print('train.main', t); print('eval.main', e); print('generate_dataset.main', g)"
  train.main <function main at 0x74de6fd8ab00>
  eval.main <function main at 0x74de6fd8ad40>
  generate_dataset.main <function main at 0x74de6f4976d0>

  $ python -c "from synth_setter.data.vst.core import extract_renderer_version; from synth_setter.models.components.transformer import ApproxEquivTransformer; from synth_setter.utils.callbacks import _log_figure; from synth_setter.metrics import ChamferDistance; print('OK')"
  OK
  ```

- [x] **`synth-setter-{train,eval,generate-dataset}` console scripts are installed and runnable** *(Level 2 — invoked each script through PATH; promise is "`[project.scripts]` produces working entry points")*
  ```bash
  $ which synth-setter-train synth-setter-eval synth-setter-generate-dataset
  /venv/main/bin/synth-setter-train
  /venv/main/bin/synth-setter-eval
  /venv/main/bin/synth-setter-generate-dataset

  $ synth-setter-train --hydra-help 2>&1 | head -2 ; echo "rc=$?"
  Hydra (1.3.2)
  See https://hydra.cc for more info.
  rc=0

  $ synth-setter-eval --hydra-help 2>&1 | head -2 ; echo "rc=$?"
  Hydra (1.3.2)
  See https://hydra.cc for more info.
  rc=0

  $ synth-setter-generate-dataset --hydra-help 2>&1 | head -2 ; echo "rc=$?"
  Hydra (1.3.2)
  See https://hydra.cc for more info.
  rc=0
  ```
  Note: I use `--hydra-help` (the standard "exit-0 help surface") rather than `--help`. The latter requires Hydra to compose the config, which fails on `train.yaml`/`eval.yaml`/`dataset.yaml` because those configs declare `???` MISSING keys (e.g., `model: ???`) — this is **pre-existing Hydra behavior**, identical to `python src/train.py --help` before the migration. The migration doesn't change `--help` behavior; it preserves it.

- [x] **Module-form `python -m synth_setter.cli.X` still works alongside console scripts** *(Level 2 — both surfaces must work per the issue's "Both surfaces must work" requirement)*
  ```bash
  $ python -m synth_setter.cli.train --hydra-help > /dev/null && echo "train: OK"
  train: OK
  $ python -m synth_setter.cli.eval --hydra-help > /dev/null && echo "eval: OK"
  eval: OK
  $ python -m synth_setter.cli.generate_dataset --hydra-help > /dev/null && echo "generate_dataset: OK"
  generate_dataset: OK
  ```

- [x] **Hydra `config_path` correctly resolves `configs/` from the new `src/synth_setter/cli/` location** *(Level 2 — actually composed a real config tree end-to-end; promise is "`../../../configs` finds the repo's configs/")*
  ```bash
  $ python -m synth_setter.cli.train --cfg job --resolve experiment=kosc/ffn_mse 2>&1 | grep -E "_target_" | head -6
    _target_: synth_setter.data.kosc_datamodule.KOscDataModule
    _target_: synth_setter.models.ksin_ff_module.KSinFeedForwardModule
      _target_: torch.optim.Adam
      _target_: synth_setter.models.components.residual_mlp.CNNResidualMLP
      _target_: lightning.pytorch.callbacks.ModelCheckpoint
      _target_: lightning.pytorch.callbacks.LearningRateMonitor
  ```
  Hydra successfully traversed `configs/experiment/kosc/ffn_mse.yaml` through the full composition chain and resolved every `_target_:` to the new `synth_setter.X` paths.

- [x] **Rewritten `_target_:` keys instantiate the same classes as before** *(Level 2 — actually instantiated datamodule + model via `hydra.utils.instantiate`; promise is "the `src.X` → `synth_setter.X` rewrite is value-preserving")*
  ```bash
  $ python -c "
  import synth_setter.cli.train  # registers custom OmegaConf resolvers (mul, etc.)
  from hydra import compose, initialize
  from hydra.utils import instantiate
  with initialize(version_base='1.3', config_path='configs'):
      cfg = compose(config_name='train.yaml', overrides=['experiment=kosc/ffn_mse'])
  dm = instantiate(cfg.data)
  mdl = instantiate(cfg.model)
  print('datamodule:', type(dm).__module__ + '.' + type(dm).__name__)
  print('model:', type(mdl).__module__ + '.' + type(mdl).__name__)
  "
  datamodule: synth_setter.data.kosc_datamodule.KOscDataModule
  model: synth_setter.models.ksin_ff_module.KSinFeedForwardModule
  ```
  Both classes instantiated successfully — `_target_:` rewrites resolve to real, importable classes.

- [x] **Zero remaining `from src.{data,models,utils,metrics,train,eval,generate_dataset}` imports** *(Level 6 — grep; promise is literally "this string is absent from the codebase", which is what grep proves)*
  ```bash
  $ find . -name "*.py" -not -path "*/.git/*" | xargs grep -nE "^(\s*)(from src\.|import src\.)(data|models|utils|metrics|train|eval|generate_dataset)\b" 2>/dev/null | wc -l
  0
  ```

- [x] **Zero remaining `_target_: src.{data,models,utils,metrics,train,eval,generate_dataset}` references in `configs/`** *(Level 6 — grep; same absence-style claim)*
  ```bash
  $ find configs -name "*.yaml" -o -name "*.yml" | xargs grep -nE "_target_:\s*src\.(data|models|utils|metrics|train|eval|generate_dataset)" 2>/dev/null | wc -l
  0
  ```

- [x] **`make test-fast` passes** *(Level 2 — full collection + run; this is the project's CPU-only test smoke gate. Cited here, not as a standalone "tests pass" checkbox, because it specifically proves the test collection imports still resolve under the new package names — the failure mode that would fire if any rewrite was incomplete.)*
  ```bash
  $ make test-fast 2>&1 | tail -1
  ================== 556 passed, 5 skipped, 1 warning in 27.65s ==================
  ```

- [x] **`tests/test_compare_baseline_configs.py` non-slow subset passes (fixture-baseline)** *(Level 1 — the test is itself an end-to-end check: spawns a real subprocess at a baseline ref vs. live tree, captures resolved Hydra YAML, asserts equality. The fixture-baseline cases at `FIXTURE_BASELINE=1bfa7ea` are untouched by Phase 2 since the fixture scripts don't reference `src.X`.)*
  ```bash
  $ pytest tests/test_compare_baseline_configs.py -v -m "not slow" --no-header --tb=short 2>&1 | tail -1
  ====================== 17 passed, 70 deselected in 4.16s =======================
  ```

- [x] **`tests/test_compare_baseline_configs.py` slow subset passes (model-baseline) — single-case spot check** *(Level 1 — same end-to-end mechanism; spot-checked one KOSC train case against the bumped `MODEL_BASELINE=4e08950` to confirm the bump works before running the full 52-case slow suite.)*
  ```bash
  $ pytest "tests/test_compare_baseline_configs.py::test_kosc_train_configs_are_equal[4e08950_vs_live__kosc_train__task1]" -v --no-header --tb=short 2>&1 | tail -1
  ============================== 1 passed in 9.78s ===============================
  ```

- [ ] **`tests/test_compare_baseline_configs.py` slow subset passes (all 52 model-baseline cases)** *(Level 1 — IN PROGRESS at time of posting. Currently executing case 40 of the 44 KOSC + 8 SURGE = 52 total slow cases. Will update this checkbox when the run completes. Pre-bump, the same suite failed 44/44 KOSC cases — all `_target_:` mechanical drift, audited in the PR body's "Baseline bump" section. Post-bump single-case spot-check passes (above).)*

- [x] **`src.pipeline.*` imports are preserved (Phase 3 territory not broken)** *(Level 2 — actually imported every module; promise is "this PR doesn't touch Phase 3 surface area". A regression here is the most likely accidental scope creep.)*
  ```bash
  $ python -c "import src.pipeline.schemas.spec; import src.pipeline.r2_io; import src.pipeline.skypilot_launch; print('OK')"
  OK
  $ find . -name "*.py" -not -path "*/.git/*" | xargs grep -nE "^(\s*)(from src\.pipeline|import src\.pipeline)" 2>/dev/null | wc -l
  54
  ```
  54 `src.pipeline.*` import sites preserved exactly as they were pre-PR.

- [x] **Git history is preserved across the moves (`git mv` rename detection)** *(Level 2 — actually ran `git log --follow` on representative moved files; promise is "this PR uses `git mv`, not `cp` + `rm`")*
  ```bash
  $ git log --follow --oneline src/synth_setter/cli/train.py | head -3
  4e08950 refactor(layout)!: nest src/* under synth_setter/, declare console scripts
  c757a7a revert: "build(deps): migrate lightning to pytorch_lightning" (#744)
  450cf0b build(deps): migrate lightning to pytorch_lightning (lightning quarantined on PyPI) + add docker deps for skypilot (#721)

  $ git log --follow --oneline src/synth_setter/data/ksin_datamodule.py | head -3
  4e08950 refactor(layout)!: nest src/* under synth_setter/, declare console scripts
  c757a7a revert: "build(deps): migrate lightning to pytorch_lightning" (#744)
  450cf0b build(deps): migrate lightning to pytorch_lightning (lightning quarantined on PyPI) + add docker deps for skypilot (#721)

  $ git log --follow --oneline src/synth_setter/metrics.py | head -3
  4e08950 refactor(layout)!: nest src/* under synth_setter/, declare console scripts
  af72a57 chore: reformat source files to pass ruff and docformatter
  15661c7 Training fixes
  ```
  History traverses the move for top-level modules (`metrics.py`), nested package modules (`data/ksin_datamodule.py`), AND the CLI move-with-rename (`src/train.py` → `src/synth_setter/cli/train.py`).

- [x] **`src/` contains only `__init__.py`, `pipeline/`, and `synth_setter/` post-move** *(Level 2 — `ls`; promise is "no orphan files left behind by the moves")*
  ```bash
  $ ls src/
  __init__.py
  pipeline
  synth_setter
  ```

- [x] **All 43 `.py` files moved under `synth_setter/` (no silent drops)** *(Level 2 — `find` + count; promise is "every file in the move list landed at its new home")*
  ```bash
  $ find src/synth_setter -name "*.py" | wc -l
  43
  $ ls src/synth_setter/cli src/synth_setter/data src/synth_setter/models src/synth_setter/utils
  src/synth_setter/cli:
  __init__.py  eval.py  generate_dataset.py  train.py
  src/synth_setter/data:
  __init__.py  audio_datamodule.py  components  fm_datamodule.py  kosc_datamodule.py
  ksin_datamodule.py  ot.py  surge_datamodule.py  vst
  src/synth_setter/models:
  __init__.py  components  ksin_ff_module.py  ksin_flow_matching_module.py
  surge_ff_module.py  surge_flow_matching_module.py  surge_flowvae_module.py
  src/synth_setter/utils:
  __init__.py  callbacks.py  instantiators.py  logging_utils.py  math.py
  pylogger.py  rich_utils.py  utils.py
  ```

### What I did NOT verify (and why)

- **`make format` / pre-commit hooks** — CI handles this. Per the verification skill, I don't burn a checkbox on what CI already proves. (For the record, `make format` was run locally before each commit and all hooks pass on a clean tree.)
- **Full GitHub Actions CI run** — Triggered by the PR push; the green-check report will land on the PR page. Per the skill's PR-readiness flow, the merge gate sits at "all CI checks pass + `mergeable=MERGEABLE` + Copilot quiet", which is the PR-readiness loop, not this skill's scope.
- **Performance** — Not a performance PR; no measurement needed.

ktinubu:
## Verification update — slow `test_compare_baseline_configs` suite + CI fix

### Slow suite now PASSING (Level 1 — full integration)

The full 87-case `test_compare_baseline_configs.py` run completed successfully against the bumped `MODEL_BASELINE=4e08950`:

```bash
$ pytest tests/test_compare_baseline_configs.py -v --tb=short 2>&1 | tail -1
======================== 87 passed in 677.20s (0:11:17) ========================
```

All 44 KOSC + 8 SURGE train cases, all 18 predict cases, plus the 17 fixture-baseline cases — passing end-to-end. This closes the only `[ ]` gate from my prior verification comment.

### CI subprocess regression fixed (b7c62c0)

After the PR opened, CI surfaced a regression I hadn't caught locally: two tests in `tests/pipeline/test_schemas/test_dataset_spec.py` (`test_dataset_spec_num_params_does_not_import_pedalboard` and `test_bare_spec_import_does_not_pull_data_vst_core`) spawn fresh `sys.executable` subprocesses that import `from src.pipeline.schemas.spec import DatasetSpec`. The lazy `from synth_setter.data.vst.param_spec_registry import param_specs` inside `DatasetSpec.num_params` fails in those subprocesses with `ModuleNotFoundError: No module named 'synth_setter'` because:

1. The subprocesses don't inherit pytest's `pythonpath = ["src"]` (that's an in-process sys.path manipulation).
2. CI doesn't run `pip install -e .` — it relies entirely on the pytest pythonpath.

The same root cause hit the macOS / GPU / VST-slow workflow probes — `python -c "from synth_setter.data.vst.core import load_plugin..."` runs outside pytest, so the bare `synth_setter` import doesn't resolve.

**Fix (b7c62c0):**
- Added `_subprocess_env()` helper in `tests/pipeline/test_schemas/test_dataset_spec.py` that copies the parent env and prepends `<repo>:<repo>/src` to `PYTHONPATH`. Both subprocess.run calls now pass `env=_subprocess_env()`.
- Updated `PYTHONPATH=/home/build/synth-setter` to `PYTHONPATH=/home/build/synth-setter:/home/build/synth-setter/src` in `test-gpu.yml` and `test-vst-slow.yml` (two occurrences each: smoke probe + main test step).
- Added `env: PYTHONPATH: src` to the smoke-probe step in `test-mps.yml`.

**Local re-verification:**

```bash
$ pytest tests/pipeline/test_schemas/test_dataset_spec.py -v -k "pedalboard or launcher_pure" --no-header 2>&1 | tail -1
======================= 1 passed, 60 deselected in 0.64s =======================

$ pytest tests/pipeline/test_schemas/test_dataset_spec.py::TestSpecImportStaysLauncherPure -v --no-header 2>&1 | tail -1
============================== 1 passed in 0.51s ===============================

$ make test-fast 2>&1 | tail -1
================== 556 passed, 5 skipped, 1 warning in 27.96s ==================
```

### Updated checkbox

- [x] **`tests/test_compare_baseline_configs.py` slow subset passes (all 52 model-baseline cases)** *(Level 1 — full integration; complete: 87 passed in 11m17s)*

ktinubu:
## Doc-drift follow-up (commit 90983ae)

The doc-drift hook on the prior push flagged stale `src/*` references in 7 design docs that weren't on the original sweep target list (no source files under those paths changed in this PR — the moves are in `src/`, but the docs reference *call sites* of moved files). Addressed in 90983ae:

- `docs/design/training-pipeline.md` (9 hunks) — `src/train.py` file refs + `python src/train.py …` invocations
- `docs/design/eval-pipeline.md` (~15 hunks) — `src/eval.py` refs + `_target_: src.data.surge_datamodule.SurgeDataModule` example + `src/utils/utils.py` / `src/data/{surge_datamodule,vst/core}.py` refs
- `docs/design/data-pipeline.md` (2 hunks) — missed `src/data/vst/__init__.py` glossary link and a `from src.data.vst.generate_vst_dataset` code-block import
- `docs/design/skypilot-compute-integration.md` (3 hunks) — `src/{train,eval}.py` prose refs
- `docs/design/storage-provenance-spec.md` (2 hunks) — entity-table rows
- `docs/design/training-pipeline-implementation-plan.md` (3 hunks) — current-path table entries
- `docs/design/data-pipeline-implementation-plan.md` (6 hunks) — current and proposed `src/{train,data}/...` refs

Rewrites applied:

- File-path refs → `src/synth_setter/cli/{train,eval}.py` and `src/synth_setter/{data,utils,models}/...`
- `python src/train.py …` / `python src/eval.py …` CLI invocations → `python -m synth_setter.cli.{train,eval} …` per the new canonical surface
- `_target_: src.X` → `_target_: synth_setter.X`
- `from src.data.X` → `from synth_setter.data.X` (in code-block examples)

Verification: `rg -n "src/(train|eval|metrics)\.py|src/(data|models|utils)/|from src\.(data|models|utils)|_target_:\s*src\." docs/ --type md` returns 0 matches.

Not addressed (out of scope for this PR, follow-up tracked in #784):

- **Console-script under-documentation.** Several design docs still present `python -m synth_setter.cli.train` as the canonical CLI form rather than the new `synth-setter-train` console script. Doc-drift flagged this as "medium confidence" omission drift, not regression. Will be addressed when those docs are next touched substantively.
- **Line-number anchors.** A few docs pin `:NN-MM` line ranges (e.g. `wandb-integration.md`'s table for `utils.py:101-106`) which are drift-prone. Doc-drift flagged this as low-confidence advisory, not a regression. Symbol-name anchors are the right fix but require a content review pass beyond mechanical sweep.

tests/conftest.py
```diff
@@ -46,6 +46,24 @@
...
 _VST_SUBPROCESS_TIMEOUT_SECONDS = 600
 
+# Repo root: passed as PYTHONPATH to VST subprocess so `from synth_setter.*` resolves
+# without an editable install. Mirrors pytest's `pythonpath = ["src"]` from
+# pyproject.toml. The MPS CI workflow does not run `pip install -e .` before pytest,
+# so the in-process pytest-path doesn't propagate to spawned children.
+_REPO_ROOT = Path(__file__).resolve().parents[1]
+
+
+def _subprocess_env() -> dict[str, str]:
```
ktinubu: **[code-health:block]** `_subprocess_env()` is duplicated verbatim in `tests/pipeline/test_schemas/test_dataset_spec.py:30`. Two copies of the same 8-line helper with the same docstring is exactly the DRY trip-wire the code-health checklist flags. Beyond duplication, the helper papers over a deeper smell: it exists only because spawned `subprocess.run([sys.executable, …])` children don't inherit `pythonpath = ["src"]` from `pyproject.toml` (that's a pytest-internal setting, not an env var). The reviewer-preferred fix is **(a) install the package in the failing workflow**: add `uv pip install --system -e .` to `.github/workflows/test-mps.yml`'s install step (and a `PYTHONPATH: src` at the `Run MPS tests` step env block as a belt-and-braces fallback), then drop `_subprocess_env()` entirely. Option (b) — making `generate_vst_dataset.py` self-bootstrap — is partially in place (`rootutils.setup_root(..., pythonpath=True)` at line 16 already adds the repo root), but rootutils does **not** add `src/`, so the script's own `from synth_setter.data.vst import …` at line 18 fails in a fresh interpreter without an editable install regardless. Augmenting the script's bootstrap to also append `src/` would work but spreads the fix across N entrypoints and re-introduces sys.path mutation at module top-level. Option (c) — extracting to `tests/helpers/subprocess_env.py` — eliminates the duplication but leaves the workflow misconfigured; it should be considered only as a fallback if (a) is judged out of scope for this PR. Recommendation: do (a) here, delete both copies.

tests/pipeline/test_schemas/test_dataset_spec.py
```diff
@@ -2,20 +2,41 @@
...
+
+# Repo root, used to pass `PYTHONPATH=<repo>:<repo>/src` to subprocess.run so
+# the spawned interpreter can resolve both `src.pipeline.*` AND `synth_setter.*`
+# without requiring an editable install. Mirrors pytest's `pythonpath = ["src"]`
+# from pyproject.toml; needed for tests that exercise launcher-pure / pedalboard-
+# free invariants in a fresh interpreter.
+_REPO_ROOT = Path(__file__).resolve().parents[3]
+
+
+def _subprocess_env() -> dict[str, str]:
```
ktinubu: **[code-health:warn]** Second copy of `_subprocess_env()` (the primary lives in `tests/conftest.py:56`). Cross-ref the block on that line — fix at the root cause (`pip install -e .` in `.github/workflows/test-mps.yml`) and delete both helpers. If the fix has to stay test-side for now, move the function to `tests/helpers/subprocess_env.py` and import it from both sites; the inline comments explaining "why PYTHONPATH" are identical here and in conftest, and they'll drift the first time someone edits one and not the other.

tests/conftest.py
```diff
@@ -46,6 +46,24 @@
 # no per-call tuning, no stack-distant default.
 _VST_SUBPROCESS_TIMEOUT_SECONDS = 600
 
+# Repo root: passed as PYTHONPATH to VST subprocess so `from synth_setter.*` resolves
```
ktinubu: **[synth-setter:warn]** 4-line block comment above `_REPO_ROOT` violates CLAUDE.md's comment-hygiene rule ("Keep comments terse — typically one short line… more than ~2 lines is a signal the context belongs in a GitHub issue"). The block restates the workflow problem in prose. Once the underlying workflow is fixed (install `-e .` in test-mps.yml) this whole helper goes away, so don't pay a paragraph in the source to document a workaround — open a follow-up issue for the CI-install fix and point the comment at it: `# Workaround for MPS CI not installing the package — see #<issue>.`

tests/pipeline/test_schemas/test_dataset_spec.py
```diff
@@ -2,20 +2,41 @@
...
-from src.data.vst import param_specs
 from src.pipeline.schemas.spec import (
     DatasetSpec,
     RenderConfig,
     ShardSpec,
 )
+from synth_setter.data.vst import param_specs
+
+# Repo root, used to pass `PYTHONPATH=<repo>:<repo>/src` to subprocess.run so
+# the spawned interpreter can resolve both `src.pipeline.*` AND `synth_setter.*`
```
ktinubu: **[synth-setter:warn]** Same multi-sentence inline comment block as `tests/conftest.py:49`, with slightly different wording ("needed for tests that exercise launcher-pure / pedalboard-free invariants…"). Two prose explanations of the same workaround in two files will drift. Collapse to a one-line pointer to the workaround-tracking issue.

tests/pipeline/test_schemas/test_dataset_spec.py
```diff
@@ -2,20 +2,41 @@
...
 from __future__ import annotations
 
+import os
 import subprocess
 import sys
 from datetime import datetime, timezone
+from pathlib import Path
 from typing import Any
 
 import pytest
```
ktinubu, line 10: **[python-style:warn]** Import ordering: `from src.pipeline.schemas.spec import …` (a `src.*` import) is interleaved between stdlib/third-party blocks and `from synth_setter.data.vst import …`. After Phase 3 dissolves `src.pipeline`, this block will need re-sorting. Ruff's I rule should already group these — verify `make format` was run; if it was and they end up in this order, consider whether `src` should be on Ruff's `known-first-party` list alongside `synth_setter` so both are sorted together.

src/synth_setter/cli/generate_dataset.py
```diff
@@ -22,15 +23,15 @@
...
 # Set PROJECT_ROOT env var and add the repo root to sys.path so
 # ``configs/paths/default.yaml``'s ``root_dir: ${oc.env:PROJECT_ROOT}`` interpolation
-# resolves under ``python -m src.generate_dataset``. Mirrors
-# ``src/train.py`` / ``src/eval.py``.
+# resolves under ``python -m synth_setter.cli.generate_dataset``. Mirrors
+# ``synth_setter/cli/train.py`` / ``synth_setter/cli/eval.py``.
 rootutils.setup_root(__file__, indicator=".project-root", pythonpath=True)
 
-from src.data.vst.core import extract_renderer_version  # noqa: E402
 from src.pipeline import r2_io  # noqa: E402
```
ktinubu, line 28: **[code-health:warn]** Mixed-namespace imports at module top: `from src.pipeline import r2_io` / `from src.pipeline.constants …` (legacy `src.*`) sit next to `from synth_setter.data.vst.core import …` (new namespace). This is unavoidable until Phase 3 hoists `src/pipeline/` under `synth_setter/`, but worth a one-line `# TODO(#784): collapse to synth_setter.pipeline.*` above the block so the next person doesn't think this is the steady state.

src/synth_setter/cli/generate_dataset.py
```diff
@@ -112,7 +113,7 @@ def build_generate_args(spec: DatasetSpec, shard: ShardSpec, output_dir: Path) -
     output_path = output_dir / shard.filename
     args = [
         sys.executable,
-        "src/data/vst/generate_vst_dataset.py",
+        "src/synth_setter/data/vst/generate_vst_dataset.py",
```
ktinubu, line 114: **[ml-pipeline:warn]** Hardcoded path string `"src/synth_setter/data/vst/generate_vst_dataset.py"` passed to subprocess.run from a launcher under the same package. After install (`pip install -e .`), the launcher and the script are both importable as `synth_setter.*`; a more robust invocation is `[sys.executable, "-m", "synth_setter.data.vst.generate_vst_dataset", str(output_path), …]` — that survives a future move of the file under the package and removes the cwd-dependency (the path string only resolves when the worker's cwd is repo root). Out of scope for a rename PR, but worth a follow-up since the rename touched the line anyway.

src/synth_setter/cli/eval.py
```diff
@@ -24,7 +24,7 @@
 # more info: https://github.com/ashleve/rootutils
```
ktinubu: *(anchored at line 24; finding is on line 11, outside the diff hunks)*

**[code-health:warn]** The 15-line rootutils explanatory comment block (lines 10–25) is copy-pasted in `train.py:12-27` and largely repeats the official rootutils README. Per CLAUDE.md "Keep comments terse — typically one short line". Replace with `# Bootstrap PROJECT_ROOT + sys.path — see https://github.com/ashleve/rootutils.` in both files (and `generate_dataset.py:24-27` has its own variant — collapse all three to the same one-liner).

src/synth_setter/cli/train.py
```diff
@@ -26,7 +26,7 @@
 # more info: https://github.com/ashleve/rootutils
```
ktinubu: *(anchored at line 26; finding is on line 13, outside the diff hunks)*

**[code-health:warn]** Cross-ref the block on `src/synth_setter/cli/eval.py:11` — identical 15-line rootutils prose duplicated here. Single source of truth: a one-line pointer to the rootutils repo.

src/synth_setter/cli/eval.py
```diff
@@ -24,7 +24,7 @@
...
 # ------------------------------------------------------------------------------------ #
 
-from src.utils import (
+from synth_setter.utils import (
```
ktinubu, line 12: **[python-style:warn]** The bare `from synth_setter.utils import (RankedLogger, extras, …)` after the rootutils bootstrap call is fine, but the new `synth_setter/utils/__init__.py` re-exports 10 symbols at module-import time. Pulling `RankedLogger`/`watch_gradients`/etc. unconditionally is what makes `import synth_setter.utils` heavyweight. Consider whether the public surface really needs to be a flat re-export — explicit `from synth_setter.utils.pylogger import RankedLogger` etc. at each call site keeps the import cost paid only where the symbol is used. Not a blocker for this PR; flag for follow-up if `synth_setter.utils` import time turns into a launcher-purity problem later.

src/synth_setter/utils/__init__.py
```diff
@@ -0,0 +1,11 @@
+from synth_setter.utils.instantiators import instantiate_callbacks, instantiate_loggers
```
ktinubu, line 1: **[ml-pipeline:warn]** New flat-namespace re-export of 10 utils functions. The launcher-pure invariant test in `tests/pipeline/test_schemas/test_dataset_spec.py::TestSpecImportStaysLauncherPure` only asserts about `synth_setter.data.vst.core` and `pedalboard` — but with `synth_setter.utils` now pulling `instantiators`, `logging_utils`, `pylogger`, `rich_utils`, `utils` at module load, anyone who `import synth_setter.utils` from a launcher transitively pulls torch/lightning. Worth either (a) keeping `utils/__init__.py` empty and forcing explicit submodule imports, or (b) extending the launcher-purity test to also assert `torch` and `lightning` don't leak when only `synth_setter.cli.generate_dataset` is imported. The latter is the more durable invariant.

tests/test_compare_baseline_configs.py
```diff
@@ -82,11 +82,16 @@
...
 # Update on PR merge: MODEL_BASELINE bumps when a published-results-relevant
 # config change lands and a new release tag is cut.
 FIXTURE_BASELINE = "1bfa7ea9c4b237a4561a9ac546a3e241ecff5951"  # PR #679 merge commit on main
-MODEL_BASELINE = "v0.0.0"
+# Bumped from v0.0.0 (= 79552d2) for #989 (Phase 2 src-layout migration). All
+# `_target_:` keys in configs/ moved from `src.X` to `synth_setter.X`, and
+# jobs/train/{kosc,surge}/train.sh switched from `python src/train.py` to
+# `python -m synth_setter.cli.train`. The bump is to the head of the Phase 2 PR
+# so the resolved Hydra YAMLs on both sides reference the new module paths.
+MODEL_BASELINE = "4e0895090f193464a1b716b50e0b2e10a55be2dd"
```
ktinubu, line 91: **[ml-test:warn]** `MODEL_BASELINE` is now pinned to a 40-char commit SHA (`4e0895090f193464a1b716b50e0b2e10a55be2dd`) instead of the tagged release `v0.0.0`. The comment says the SHA points to "the head of the Phase 2 PR" — but the PR's head is currently `bf788c7`, not `4e08950`. Either (a) this SHA references a commit that no longer exists on the branch (rebased / force-pushed away), in which case the baseline-fetch in CI will fail, or (b) it's a deliberate pin to an intermediate commit on this branch, in which case the comment is misleading. Verify with `git cat-file -e 4e0895090f193464a1b716b50e0b2e10a55be2dd^{commit}` and either update the SHA to a stable post-merge commit or move to a tag after merge.

tests/test_compare_baseline_configs.py
```diff
@@ -82,11 +82,16 @@
 # fetches the requested ref itself when needed (see `try_fetch_ref`).
 # FIXTURE_BASELINE pins the synthetic-fixture equality + inequality tests.
 # MODEL_BASELINE pins the K-OSC + SURGE train.sh tests against a known-good
-# model-config snapshot (tag v0.0.0, == 79552d2).
+# model-config snapshot.
 # Update on PR merge: MODEL_BASELINE bumps when a published-results-relevant
 # config change lands and a new release tag is cut.
 FIXTURE_BASELINE = "1bfa7ea9c4b237a4561a9ac546a3e241ecff5951"  # PR #679 merge commit on main
-MODEL_BASELINE = "v0.0.0"
+# Bumped from v0.0.0 (= 79552d2) for #989 (Phase 2 src-layout migration). All
```
ktinubu: **[synth-setter:warn]** The 6-line block-comment from line 84 to 93 restates the baseline-bump rationale, the previous SHA, the new SHA, and the PR/issue link inline. Per CLAUDE.md comment-hygiene, this prose belongs in the PR body or the linked issue; the source comment should be a one-liner pointing at the issue (`# Baseline pinned for Phase 2 src-layout migration — see #989.`).

ktinubu (commented):
Multi-skill review of PR #991 — 9 parallel passes (code-health, synth-setter-project-standards, python-style, tdd-implementation, tdd-refactor, ml-data-pipeline, ml-test, shell-style, gha-workflow-validator). Every BLOCK/WARN posted below as an individual unresolved thread.

Note on orchestration: this run was performed by a single agent rather than parallel sub-agents because the Task/Agent tool was not loaded in this environment; the per-skill checklists were applied sequentially. All findings still carry their per-skill prefix so reviewers can see which checklist surfaced each one.

The bulk of the diff is mechanical (40 file renames `src/<X>` → `src/synth_setter/<X>`, ~140 import-path rewrites, ~47 `_target_:` config rewrites, jobs/*.sh `python src/{train,eval}.py` → `python -m synth_setter.cli.{train,eval}`). Those changes look internally consistent — no stale `from src.data` / `from src.models` / `from src.utils` / `from src.metrics` / `_target_: src.X` references remain outside CHANGELOG.md. The substantive review surface is small: the new console-script wiring in `pyproject.toml`, the `_subprocess_env()` helper duplicated across two tests, the workflow `PYTHONPATH` rewrites, and the `MODEL_BASELINE` bump in `tests/test_compare_baseline_configs.py`.

.github/workflows/test-mps.yml
```diff
@@ -64,9 +64,13 @@ jobs:
...
       # Mirrors the Tart-VM smoke check in `tart/macos.pkr.hcl` — fails fast if
       # the cask install or symlink is broken before the full pytest run.
+      # `PYTHONPATH=src` exposes the `synth_setter` package without an editable
+      # install (mirrors pytest's `pythonpath = ["src"]` from pyproject.toml).
       - name: Smoke-test Surge XT plugin load
+        env:
+          PYTHONPATH: src
         run: |
-          python -X faulthandler -c "from src.data.vst.core import load_plugin; load_plugin('plugins/Surge XT.vst3')"
+          python -X faulthandler -c "from synth_setter.data.vst.core import load_plugin; load_plugin('plugins/Surge XT.vst3')"
```
ktinubu, line 70: **[gha:block]** The `Run MPS tests` step at line 86–89 runs `pytest -vv -s -m mps` without `PYTHONPATH` set and without an editable install. pytest's `pythonpath = ["src"]` (pyproject.toml) makes `synth_setter` importable **in-process**, but the subprocess fixtures in `tests/conftest.py` and `tests/pipeline/test_schemas/test_dataset_spec.py` spawn fresh interpreters via `subprocess.run([sys.executable, …])` — and **that's exactly why** `_subprocess_env()` was added in this PR. The proper fix is here, not in the test helpers: add `uv pip install --system -e .` to the install step (line 50–53) — `pyproject.toml`'s `[tool.setuptools.package-dir]` already maps `synth_setter` to `src/synth_setter`, so an editable install registers the import for all subprocesses too. Once installed, both `_subprocess_env()` helpers can be deleted.

.github/workflows/test-mps.yml
```diff
@@ -64,9 +64,13 @@ jobs:
 
       # Mirrors the Tart-VM smoke check in `tart/macos.pkr.hcl` — fails fast if
       # the cask install or symlink is broken before the full pytest run.
+      # `PYTHONPATH=src` exposes the `synth_setter` package without an editable
+      # install (mirrors pytest's `pythonpath = ["src"]` from pyproject.toml).
       - name: Smoke-test Surge XT plugin load
+        env:
+          PYTHONPATH: src
```
ktinubu: **[gha:warn]** The smoke step at line 69–73 sets `PYTHONPATH: src` (relative path) at the step level, but the `Run MPS tests` step at line 86–89 omits the same env block. If the chosen fix is workflow-level rather than installing the package, `PYTHONPATH: src` belongs on the pytest step too — otherwise the smoke step passes but the real pytest run hits the very subprocess-import failure that `_subprocess_env()` was introduced to fix. Prefer the editable install (see the block on line 73); this is the second-best.

.github/workflows/test-gpu.yml
```diff
@@ -54,14 +54,14 @@ jobs:
         run: |
           docker run --rm \
             -v ${{ github.workspace }}:/home/build/synth-setter \
-            -e PYTHONPATH=/home/build/synth-setter \
+            -e PYTHONPATH=/home/build/synth-setter:/home/build/synth-setter/src \
```
ktinubu, line 57: **[gha:warn]** `PYTHONPATH=/home/build/synth-setter:/home/build/synth-setter/src` is now set inline in the smoke `docker run` (line 57) and in the test `docker run` (line 79). Same root cause as test-mps.yml: the container image doesn't ship the package as an editable install, so callers append `src/` to `PYTHONPATH`. Two repetitions of the same colon-joined path is a yellow flag — if `src/` ever moves under Phase 3, every workflow needs a coordinated edit. Consider either baking `pip install -e /home/build/synth-setter` into the image, or extracting the path to a workflow-level `env:` block referenced from both `docker run` invocations.

.github/workflows/test-vst-slow.yml
```diff
@@ -79,14 +79,14 @@ jobs:
         run: |
           docker run --rm \
             -v ${{ github.workspace }}:/home/build/synth-setter \
-            -e PYTHONPATH=/home/build/synth-setter \
+            -e PYTHONPATH=/home/build/synth-setter:/home/build/synth-setter/src \
```
ktinubu, line 82: **[gha:warn]** Same `PYTHONPATH=/home/build/synth-setter:/home/build/synth-setter/src` rewrite as test-gpu.yml (here and at line 110). Cross-ref that finding — bake the editable install into the image so the workflow doesn't carry the import path.

setup.py
```diff
@@ -2,6 +2,10 @@
 
 from setuptools import find_packages, setup
 
+# Legacy `console_scripts` (`train_command`, `eval_command`) were removed in
+# the Phase 2 src-layout migration (#989). Console scripts are now declared in
+# `pyproject.toml` under `[project.scripts]`. Full `setup.py` deletion happens
+# in Phase 5 of #784.
 setup(
     name="src",
```
ktinubu, line 10: **[synth-setter:warn]** `setup.py` still declares `name="src"` and uses `find_packages()` with no `where=` argument while `pyproject.toml` declares the real package list explicitly under `[tool.setuptools]`. With both files present, behavior depends on which one setuptools picks up (pyproject is the modern source of truth; setup.py is legacy). The new comment correctly notes Phase 5 deletes setup.py, but until then keeping both files with conflicting metadata (`name="src"` vs project.name="synth-setter") is a foot-gun for anyone who runs `pip install .` against the setup.py path. Either delete setup.py now (the PR description suggests it's no longer load-bearing once `[project.scripts]` is in pyproject) or document why the legacy file must stay through Phase 2–4.

pyproject.toml
```diff
@@ -45,6 +42,11 @@ packages = [
 # Other packages use the implicit root mapping (`src` lives at ./src, etc.).
 "synth_setter" = "src/synth_setter"
 
+[project.scripts]
+synth-setter-train             = "synth_setter.cli.train:main"
+synth-setter-eval              = "synth_setter.cli.eval:main"
```
ktinubu: **[python-style:warn]** Console-script declarations are aligned with extra whitespace (`synth-setter-train             = …`). Ruff format will leave this alone (TOML isn't formatted by Ruff), but most TOML tooling normalizes to single-space-around-equals. Cosmetic only — flag for the next round of pyproject-touch.

pyproject.toml
```diff
@@ -16,27 +16,24 @@ docs = [
...
-# `from src.X` at ~52 import sites today. Phases 2–5 rewrite those imports
-# and this list collapses to a `[tool.setuptools.packages.find]` with
-# `where = ["src"]`.
+# Phase 2 of the PEP src-layout migration (#989). The data/models/utils/metrics
+# trees and the CLI entrypoints now live under `synth_setter`; only
+# `src/pipeline/` remains as `src.pipeline` (moves in Phase 3, #784).
+# Enumerated rather than discovered so the dual `synth_setter` / `src.pipeline`
+# layout stays explicit until Phase 3 collapses to a single `find` directive.
 packages = [
   "synth_setter",
```
ktinubu, line 25: **[code-health:warn]** Explicit `packages = […]` enumeration with 11 entries duplicating the directory structure. The comment acknowledges this will collapse to `find` in Phase 3, but in the meantime any new subpackage added under `synth_setter/` will be silently invisible to `pip install` until someone remembers to add it here. Consider `[tool.setuptools.packages.find]` with `where = ["src"]` and `include = ["synth_setter*", "pipeline*"]` even during the transition — fewer hand-maintained lists, less drift risk.

tests/pipeline/test_entrypoints/test_generate_dataset.py
```diff
@@ -1,4 +1,4 @@
-"""Tests for src/generate_dataset.py — spec-driven run.
+"""Tests for src/synth_setter/cli/generate_dataset.py — spec-driven run.
```
ktinubu: **[tdd-refactor:warn]** The patch targets across all 12 test methods (`@patch("synth_setter.cli.generate_dataset.subprocess.check_call")` etc.) were swept correctly. Good. One residual: line 1's docstring header `"""Tests for src/synth_setter/cli/generate_dataset.py — spec-driven run."""` references the **file path** rather than the **module path**. Prefer module form so the docstring doesn't drift if the file moves: `"""Tests for synth_setter.cli.generate_dataset — spec-driven run."""`.

src/pipeline/skypilot_launch.py
```diff
@@ -46,9 +46,9 @@
 from hydra import compose, initialize_config_dir
 from hydra.errors import HydraException
 
-from src.generate_dataset import spec_from_cfg
 from src.pipeline.partitioning import NUM_WORKERS_ENV_VAR, WORKER_RANK_ENV_VAR
 from src.pipeline.schemas.spec import DatasetSpec
+from synth_setter.cli.generate_dataset import spec_from_cfg
```
ktinubu, line 51: **[tdd-refactor:warn]** `from synth_setter.cli.generate_dataset import spec_from_cfg` in a launcher-pure module. `spec_from_cfg` itself is interpreter-only, but the import chain now goes launcher → `synth_setter.cli.generate_dataset` (module-level imports include `subprocess`, `tempfile`, `hydra`, `rootutils`, `loguru`, `omegaconf`, `synth_setter.data.vst.core` and `src.pipeline.*`). Importing the CLI module from the launcher pulls every CLI dep into the launcher's interpreter. The existing pedalboard-free invariant test in `tests/pipeline/test_schemas/test_dataset_spec.py` covers `pedalboard`; consider extending it (or adding a sibling test) to assert that importing `src.pipeline.skypilot_launch` doesn't transitively import `hydra` / `loguru` / `rootutils` — those are present in launcher-only requirements today but the invariant deserves a test.

src/pipeline/ci/materialize_spec.py
```diff
@@ -23,8 +23,8 @@
 
 rootutils.setup_root(__file__, indicator=".project-root", pythonpath=True)
 
-from src.generate_dataset import spec_from_cfg  # noqa: E402
 from src.pipeline.constants import INPUT_SPEC_FILENAME  # noqa: E402
+from synth_setter.cli.generate_dataset import spec_from_cfg  # noqa: E402
```
ktinubu, line 27: **[tdd-refactor:warn]** Same shape as `skypilot_launch.py:51` — CI helper imports `from synth_setter.cli.generate_dataset import spec_from_cfg`, which now drags every CLI-module top-level import into the CI helper. Either pull `spec_from_cfg` into a lighter module (e.g. `src/pipeline/schemas/builder.py` — it's a 4-line function that only needs `OmegaConf`, `DictConfig`, `DatasetSpec`, and the `_NON_SPEC_KEYS` tuple) and import from both call sites, or accept the heavier import on the CI helper. The pure-function `spec_from_cfg` doesn't belong in the same module as `_rclone_copy` long-term.

src/__init__.py
```diff
@@ -1 +1,7 @@
-"""Synth-setter ML models, data modules, and training utilities."""
+"""Legacy ``src`` package.
+
+Post-Phase-2 it hosts only ``src.pipeline``;
+the rest of the codebase has moved to ``src/synth_setter/``. ``src.pipeline``
+itself moves under ``synth_setter`` in Phase 3 (#784), after which this file
```
ktinubu: **[tdd-refactor:warn]** Module docstring says "the rest of the codebase has moved to src/synth_setter/" — accurate for what landed, but `src.pipeline` itself is still very much the codebase too. Consider rephrasing to "the ML code (data/models/utils/metrics) has moved to src/synth_setter/; only src/pipeline/ remains here" so a reader cloning at this PR doesn't wonder why `src/pipeline/` is still around.

tests/conftest.py
```diff
@@ -46,6 +46,24 @@
...
+# without an editable install. Mirrors pytest's `pythonpath = ["src"]` from
+# pyproject.toml. The MPS CI workflow does not run `pip install -e .` before pytest,
+# so the in-process pytest-path doesn't propagate to spawned children.
+_REPO_ROOT = Path(__file__).resolve().parents[1]
+
+
```
ktinubu: **[python-style:warn]** `def _subprocess_env() -> dict[str, str]:` uses lowercase `dict` annotation. Fine in Python 3.10+. `tests/pipeline/test_schemas/test_dataset_spec.py:30` has the same signature **with** `from __future__ import annotations` at the top, while this file does not. Inconsistent future-import handling between the two copies is one more reason to deduplicate into a shared helper module.

src/synth_setter/cli/generate_dataset.py
```diff
@@ -22,15 +23,15 @@
...
 
-from src.data.vst.core import extract_renderer_version  # noqa: E402
 from src.pipeline import r2_io  # noqa: E402
 from src.pipeline.constants import INPUT_SPEC_FILENAME  # noqa: E402
 from src.pipeline.partitioning import get_my_shards, read_rank_world_from_env  # noqa: E402
 from src.pipeline.schemas.spec import DatasetSpec, ShardSpec  # noqa: E402
+from synth_setter.data.vst.core import extract_renderer_version  # noqa: E402
 
 # Composed-config keys that aren't DatasetSpec fields: ``data`` / ``r2`` are interpolation
 # sources for top-level keys; ``paths`` / ``hydra`` exist only for Hydra runtime; ``run_name``
```
ktinubu, line 35: *(anchored at line 37; finding is on line 67, outside the diff hunks)*

**[code-health:warn]** Constants block at lines 36–39 (`_NON_SPEC_KEYS`) and 62–67 (`VST_HEADLESS_WRAPPER`) are split by 25 lines of function definitions (`load_spec_from_uri`). Group module-level constants at the top after imports, then functions in order of use. Pre-existing layout carried through the rename, but worth tidying since the file was touched anyway.

src/synth_setter/cli/eval.py
```diff
@@ -116,7 +116,7 @@ def evaluate(cfg: DictConfig) -> tuple[dict[str, Any], dict[str, Any]]:
     return metric_dict, object_dict
```
ktinubu, line 102: *(anchored at line 116; finding is on line 94, outside the diff hunks)*

**[code-health:warn]** `elif mode == "val" or mode == "validate":` — accepting two spellings is fine, but document it (one-line `# accept both spellings for compatibility with older configs`) or pick one. Pre-existing, but the rename touched the file so a future reader will assume this PR introduced it.

ktinubu, line 102: *(anchored at line 116; finding is on line 103, outside the diff hunks)*

**[code-health:warn]** `predictions = trainer.predict(...)` — assignment unused (the variable is never returned or logged). Same in the original `src/eval.py` so this is a no-op rename, but worth flagging as a follow-up: either drop the variable or pass `predictions` to a writer.

tests/conftest.py
ktinubu: Addressed in 64ac16d via option (a) — installed the package as editable in test.yml (3 jobs), test-mps.yml, and test-conda.yml. Deleted `_subprocess_env()` from both call sites along with its supporting imports/constants. Local `pytest tests/pipeline/test_schemas/test_dataset_spec.py -v` (61/61) green without the helper; CI re-running on this push will validate the MPS path.

.github/workflows/test-mps.yml
ktinubu: Addressed in 64ac16d. Added `uv pip install --system -e .` to the Install dependencies step. Also dropped the `PYTHONPATH: src` env block from the Smoke-test step (b7c62c0's workaround — no longer needed once the package is installed). The two `_subprocess_env()` helpers in tests are deleted in the same commit.

PR #1010, opened 2026-05-13
chore(notebooks): remove flush-investigation notebook + weekly workflow

### LINKED ISSUE #227: investigate(pipeline): characterize flush+reset cycle effects on audio quality
## Summary

The fix in #225 made the post-load flush+reset unconditional, and the post-param and post-render flushes optional (`FULL_FLUSH`). But we don't fully understand what each cycle does beyond parameter enumeration. Before generating the next large dataset, we need to characterize whether skipping the optional flushes affects audio quality.

### 1. Post-load flush+reset (after `load_preset()`) — MANDATORY

We know this is required for parameter re-enumeration (see #225). But it may also:
- Flush audio residue from the preset load itself (some plugins start oscillators on preset load)
- Initialize internal plugin state (LFO phase, envelope position) to a known baseline

**Question:** Does the post-load flush affect the rendered audio, or only the parameter map?

### 2. Post-param flush+reset (after `set_params()`) — CURRENTLY OPTIONAL

Setting raw parameter values shouldn't produce audio, but some plugins react to parameter changes by starting internal processes (LFOs, envelopes, modulators). This flush would clear that state before the "real" render.

We're currently skipping this and relying on `process(reset=True)` in the render call. But `reset=True` may not be equivalent to an explicit `process()+reset()` cycle.

**Question:** Does skipping the post-param flush cause subtle audio differences (LFO phase offsets, envelope artifacts in early samples)?

### 3. Post-render flush+reset (after render `process()` call) — CURRENTLY OPTIONAL

Purely defensive — clears state between iterations. The next iteration starts with a preset load + post-load flush anyway, so this should be redundant.

**Question:** Is there any state that survives across preset loads that could leak between iterations?

## Proposed investigation

1. **A/B comparison**: Generate a small batch (e.g. 100 samples) with `FULL_FLUSH=0` vs `FULL_FLUSH=1`, using identical random seeds
2. **Compare outputs**: Diff the raw audio arrays (not just spectrograms) — are they bit-identical?
3. **If not identical**: Characterize the differences — magnitude, location (early samples vs throughout), frequency content
4. **Edge cases**: Test with different oscillator types, LFO-heavy presets, modulation-heavy patches

## Context

- Fix: #225 / PR #226
- Code: [`src/data/vst/core.py` — `render_params()`](https://github.com/tinaudio/synth-setter/blob/experiment/src/data/vst/core.py#L66-L121)
- Previous successful dataset (929GB, pre-`8ae2f85`): generated with all 3 flushes enabled

## Additional: flush duration

The flush calls currently use `32.0` seconds of silent audio (`plugin.process([], 32.0, ...)`), which allocates a large buffer just to trigger parameter re-enumeration or state clearing. This is likely excessive — a minimal duration (e.g. `0.05` seconds / one buffer) may suffice. Both `core.py:render_params()` and the regression tests in `test_preset_params.py` use 32.0 and should be updated together once the minimum effective duration is determined.

ktinubu:
## Summary

- Delete `notebooks/flush-investigation.ipynb` — superseded by the `generate_vst_dataset` benchmark metrics.
- Delete `.github/workflows/flush-investigation.yml` so the weekly Sunday 06:00 UTC job stops failing once the notebook it runs is gone.
- Drop the two live references in `docs/reference/github-actions.md` (Pipeline workflow-table row and artifact-chains list entry). `CHANGELOG.md` mentions are left as immutable release history.

## Test plan

- [ ] CI workflows directory still parses (no orphan `uses:` / `workflow_call` pointing at the removed workflow — `flush-investigation` was terminal in the dependency map).
- [ ] `pr-metadata-gate` passes.
- [ ] No functional Python / pipeline code paths touched, so no unit tests need to run beyond the standard lint/format pre-commit checks already enforced.

Closes #227


ktinubu:
## Verification Results

Verification-only run (no implementation). The promises this PR makes are: (a) the two flush-investigation files are deleted, (b) no live references remain in the repo that would point at a now-missing target, (c) the remaining `.github/workflows/` directory is still structurally valid, (d) `Closes #227` is recognized so the issue will auto-close on merge, and (e) CI is green and the branch is mergeable.

- [x] **Both target files are absent from the PR branch** *(Level 3 — exercises the real filesystem on the checked-out branch, not just the diff)*
  ```bash
  $ ls .github/workflows/flush-investigation.yml notebooks/flush-investigation.ipynb 2>&1; echo "exit: $?"
  ls: .github/workflows/flush-investigation.yml: No such file or directory
  ls: notebooks/flush-investigation.ipynb: No such file or directory
  exit: 0
  ```
  (`exit: 0` because shell records the last command's exit; both `ls` calls printed `No such file or directory`, confirming deletion.)

- [x] **No live references remain in the repo** *(Level 3 — grep every tracked path on the branch, exclude `CHANGELOG.md` since historical entries are immutable by project convention)*
  ```bash
  $ grep -rln -e 'flush-investigation' -e 'flush_investigation' --exclude-dir=.git --exclude=CHANGELOG.md . 2>&1; echo "exit: $?"
  exit: 1
  ```
  Exit 1 = grep found zero matches. The only remaining mentions are in `CHANGELOG.md`'s historical release entries, which are deliberately preserved.

- [x] **No orphan workflow callers point at the deleted workflow** *(Level 3 — narrow grep over `.github/` for any `uses:` / `workflow_call` reference)*
  ```bash
  $ grep -rln --include='*.yml' --include='*.yaml' 'flush-investigation\|flush_investigation' .github/ 2>&1; echo "exit: $?"
  exit: 1
  ```
  Exit 1 = no matches; the deleted workflow had no downstream callers (and `docs/reference/github-actions.md`'s dependency map already listed it as terminal — that entry is removed in this PR).

- [x] **All remaining workflow files parse as valid YAML** *(Level 5 — structural parse; the real Level-1 equivalent is `Analyze (actions)` below, which is what CI runs)*
  ```bash
  $ python3 -c "
  import yaml, pathlib, sys
  errors = 0
  files = list(pathlib.Path('.github/workflows').glob('*.yml')) + list(pathlib.Path('.github/workflows').glob('*.yaml'))
  for f in files:
      try:
          with open(f) as fh:
              yaml.safe_load(fh)
      except Exception as e:
          print(f'PARSE FAIL: {f}: {e}')
          errors += 1
  print(f'Parsed {len(files)} workflow files, {errors} errors')
  sys.exit(errors)
  "
  Parsed 32 workflow files, 0 errors
  ```

- [x] **GitHub Actions / CodeQL accepts the post-deletion workflows directory** *(Level 1 — `Analyze (actions)` is CodeQL's workflow analyzer, which parses every workflow in the repo end-to-end on GitHub's infrastructure; this is the authoritative check that the workflows dir still parses after the deletion)*
  ```bash
  $ gh pr checks 1010 --repo tinaudio/synth-setter | grep -E '^(Analyze \(actions\)|Analyze \(python\)|CodeQL)\b'
  Analyze (actions)	pass	52s	https://github.com/tinaudio/synth-setter/actions/runs/25815913344/job/75844225207
  Analyze (python)	pass	1m1s	https://github.com/tinaudio/synth-setter/actions/runs/25815913344/job/75844225120
  CodeQL	pass	2s	https://github.com/tinaudio/synth-setter/runs/75844354905
  ```

- [x] **`pr-metadata-gate` accepts the PR** *(Level 1 — the gate is the real consumer of PR taxonomy; passing on GitHub Actions is end-to-end verification)*
  ```bash
  $ gh pr checks 1010 --repo tinaudio/synth-setter | grep check-pr-metadata
  check-pr-metadata	pass	7s	https://github.com/tinaudio/synth-setter/actions/runs/25815915799/job/75844224135
  ```

- [x] **Issue #227 will auto-close on merge** *(Level 4 — query GitHub's live `closedByPullRequestsReferences` field, which reflects whether the `Closes #N` keyword was parsed and accepted)*
  ```bash
  $ gh issue view 227 --repo tinaudio/synth-setter --json closedByPullRequestsReferences -q '.closedByPullRequestsReferences[].number'
  1010
  ```

- [x] **All required CI checks pass** *(Level 1 — full set of GitHub Actions checks run on this PR's HEAD)*
  ```bash
  $ gh pr checks 1010 --repo tinaudio/synth-setter | awk '{print $1, $2}' | sort -u
  Analyze (actions) pass
  Analyze (python) pass
  Auto-approve status pass
  auto-approve (1010) pass
  CodeQL pass
  check-pr-metadata pass
  code-quality pass
  discover pass
  ```

- [x] **Branch is mergeable into `main`** *(Level 4 — query GitHub's live mergeability state)*
  ```bash
  $ gh pr view 1010 --repo tinaudio/synth-setter --json mergeable,mergeStateStatus
  {"mergeStateStatus":"CLEAN","mergeable":"MERGEABLE"}
  ```

### Adversarial / implicit checks

- [x] **The workflow was terminal — no downstream artifact consumer breaks** *(Level 3 — verified by inspecting `docs/reference/github-actions.md`'s dependency map before the doc edit; the artifact-chains entry explicitly stated `flush-investigation` uploads notebook HTML "terminal; not consumed by another workflow". That entry is removed in this PR.)*

- [x] **Notebook had no Python import / scripts dependency** *(Level 3 — grep for the notebook filename across the whole tree)*
  ```bash
  $ grep -rln 'flush-investigation.ipynb' --exclude-dir=.git --exclude=CHANGELOG.md . 2>&1; echo "exit: $?"
  exit: 1
  ```
  No code or non-CHANGELOG doc imports / references the notebook by name.

- [ ] **Empirical confirmation that `generate_vst_dataset` benchmark metrics actually supersede the notebook's findings** *(SKIP — out of scope for this verification: this PR removes the notebook on the maintainer's word that the benchmark metrics cover the same ground. To verify the equivalence claim itself, a reviewer would need to: (1) run the `generate_vst_dataset` benchmark suite on a representative shard, (2) compare the audio-quality signals it surfaces against the A/B questions raised in #227's body (post-load flush, post-param flush, post-render flush). That belongs in a follow-up audit, not in a chore-removal PR's checks.)*

Issue #489, opened 2026-04-02
bug(pipeline): pedalboard + full param set produces unexplained variance with identical patches


ktinubu:
## Problem

Using pedalboard with the full parameter set to generate synth data produces unexplained variance even when rendering the same patch. Identical inputs yield non-deterministic outputs — concretely, **~50% of the time every-other render produces junk audio that doesn't match the parameters**, suspected to be due to certain pedalboard+Surge XT parameters not reapplying correctly between renders.

## Reproducer (PR #706)

`tests/data/vst/test_generate_vst_dataset.py::test_datasets_from_hardcoded_params_are_identical` reproduces the failure mode reliably. The test renders a hardcoded loudness-passing patch 12 times across two `make_dataset` stages and asserts both per-pair (`expected[i]` vs `actual[i]`) audio metrics AND **all-pairs worst-case across the union of 12 renders** are within thresholds. The all-pairs check is the critical addition — single-pair tests miss the every-other-render junk because they never compare two stage-2 renders against each other.

Empirical comparison on `test_datasets_from_hardcoded_params_are_identical` (num_samples=6 → 12 renders, 66 all-pairs):

> **main = [`0d055b3`](https://github.com/tinaudio/synth-setter/commit/0d055b35ac85932c7584d67e57087c7432b35476)** (origin/main HEAD — this issue, no workaround)
> **feat = [`2aa51c5`](https://github.com/tinaudio/synth-setter/commit/2aa51c5985f383d2fed8ed7fc5b1e9cee9c5ef87)** (feat/surge-xt-interactive-load-prediction HEAD, PR #702 — per-render reload workaround)

| Metric  | main `0d055b3` worst @ pair | feat `2aa51c5` worst @ pair | Threshold |
|---------|-----------------------------|-----------------------------|-----------|
| MSS     | **21.39** @ (8, 11)         | 3.88 @ (7, 8)               | 10.0      |
| wMFCC   | 17.25 @ (0, 9)              | 6.49 @ (8, 10)              | 18.0      |
| SOT     | **0.293** @ (5, 10)         | 0.024 @ (1, 8)              | 0.15      |
| RMS cos | 0.886 @ (5, 10)             | 0.979 @ (7, 8)              | min 0.8   |

[...]

PR #1023, opened 2026-05-13
test(vst): disable loudness gate during replay run to defuse #489 flake

### LINKED ISSUE #1022: bug(test/vst): replay run can desync iterator when phase variance trips loudness gate
## Summary

`tests/data/vst/test_generate_vst_dataset.py::test_datasets_from_sampled_params_are_identical` is flaky. First failure on `main`: workflow run [25820904579](https://github.com/tinaudio/synth-setter/actions/runs/25820904579/job/75861579722) (1 fail / 14 pass).

## Symptom

```
tests/data/vst/test_generate_vst_dataset.py:342: StopIteration
```

raised from `_patched_sample.fake_sample` during stage 2's `make_dataset` call.

## Root cause

Stage 2 patches `param_spec.sample` with an iterator over exactly `_NUM_SAMPLES = 5` candidate `(synth, note)` tuples. Inside `generate_sample` (`src/synth_setter/data/vst/generate_vst_dataset.py:86–127`), the rendered output is gated by `loudness < min_loudness`; when the gate rejects, the loop `continue`s and pulls the **next** sample from the iterator.

The test's docstring asserts that "guaranteed-loud (step 1)" params can't trigger that retry, but it then explicitly notes (lines 832–836) that "Surge XT's oscillator phase init is nondeterministic across renders of identical params (issue #489)." Phase nondeterminism perturbs measured integrated loudness, so a candidate that *just* cleared `_MIN_LOUDNESS = -55 dB` in stage 1 can fall below it on the stage-2 re-render. When that happens, the patched iterator gets pulled past `len(replay)` and `next(replay_iter)` raises a bare `StopIteration` (PEP 479 doesn't apply since `fake_sample` isn't a generator).

## Proposed fix

Pass `min_loudness=float("-inf")` to the stage-2 `RenderConfig` so the loudness gate is disabled during replay — the gate has already served its purpose in stage 1 and re-applying it adds nothing but a flake surface. Sibling option (#489 closure) is to fix the underlying phase nondeterminism, but that's out of scope for this fix.

## Acceptance

- Stage-2 replay no longer pulls more than `_NUM_SAMPLES` items from the iterator.
- The hardcoded test (`test_datasets_from_hardcoded_params_are_identical`) is unaffected (the `_HARDCODED_SYNTH_PARAMS` capture is reliably loud).

Refs #489

ktinubu:
## Summary

- Pass `min_loudness=float("-inf")` to the stage-2 `RenderConfig` in `test_datasets_from_sampled_params_are_identical` so the loudness gate is disabled during replay.
- Thread an optional `min_loudness` override through `_render_cfg` and `_assert_h5_structure_is_valid`.
- Update docstring + add an inline note explaining why the gate is bypassed in stage 2.

The gate already served its purpose in stage 1 (every kept candidate passed it once). Re-applying it in stage 2 is the bug surface: Surge XT's nondeterministic phase init (#489) can drop a stage-1 candidate's measured loudness below `_MIN_LOUDNESS = -55 dB` on re-render, which causes `generate_sample`'s `while True` loop to pull the next iterator item from `_patched_sample` and eventually raise bare `StopIteration`.

First failure on `main`: workflow run [25820904579](https://github.com/tinaudio/synth-setter/actions/runs/25820904579/job/75861579722).

```
tests/data/vst/test_generate_vst_dataset.py:342: StopIteration
```

## Test plan

- [ ] `pre-commit run --files tests/data/vst/test_generate_vst_dataset.py` — passes locally.
- [ ] Test file collects cleanly (`pytest --collect-only`) and the 10 fast tests in the file all pass.
- [ ] `VST Slow Tests` workflow goes green on this branch.
- [ ] Sanity-check the `audio.attrs["min_loudness"]` assertion passes for both stages (stage 1 = `_MIN_LOUDNESS`, stage 2 = `-inf`).

Closes #1022
Refs #489

### DIFF
--- a/tests/data/vst/test_generate_vst_dataset.py
+++ b/tests/data/vst/test_generate_vst_dataset.py
@@ -39,11 +39,17 @@ _ABSOLUTE_TOLERANCE = 1e-7
 _RELATIVE_TOLERANCE = 1e-9
 
 
-def _render_cfg(num_samples: int, sample_batch_size: int | None = None) -> RenderConfig:
+def _render_cfg(
+    num_samples: int,
+    sample_batch_size: int | None = None,
+    min_loudness: float = _MIN_LOUDNESS,
+) -> RenderConfig:
     """Build a RenderConfig with this module's test defaults.
 
     ``batch_per_shard`` is set to ``num_samples`` (one shard = one batch in tests);
     ``sample_batch_size`` defaults to the same so each test renders in a single batch.
+    ``min_loudness`` defaults to ``_MIN_LOUDNESS`` and can be lowered (e.g. to
+    ``-inf``) to disable the loudness gate during replay runs — see #489.
     """
     return RenderConfig(
         plugin_path=_PLUGIN_PATH,
@@ -54,7 +60,7 @@ def _render_cfg(num_samples: int, sample_batch_size: int | None = None) -> Rende
         channels=_CHANNELS,
         velocity=_VELOCITY,
         signal_duration_seconds=_DURATION,
-        min_loudness=_MIN_LOUDNESS,
+        min_loudness=min_loudness,
         sample_batch_size=sample_batch_size if sample_batch_size is not None else num_samples,
         batch_per_shard=num_samples,
     )
@@ -268,13 +274,14 @@ _HARDCODED_NOTE_PARAMS: dict[str, int | tuple[float, float]] = {
 
 
 def _assert_h5_structure_is_valid(
-    out: Path, spec, num_samples: int
+    out: Path, spec, num_samples: int, min_loudness: float = _MIN_LOUDNESS
 ) -> tuple[np.ndarray, np.ndarray, np.ndarray]:
     """Open ``out`` and assert dataset keys, shapes, dtypes, attrs, and finiteness.
 
     Returns materialized (audio, mel_spec, param_array) numpy arrays so callers can perform per-row
     decode checks without re-opening the file. Shared by the round-trip and random-sampling e2e
-    tests.
+    tests. ``min_loudness`` overrides the expected ``audio.attrs["min_loudness"]`` for callers
+    that disable the loudness gate (e.g. replay runs that pass ``-inf``).
     """
     expected_audio_shape = (num_samples, _CHANNELS, int(_SAMPLE_RATE * _DURATION))
     expected_mel_shape = (num_samples, *_PER_SAMPLE_MEL_SHAPE)
@@ -304,7 +311,7 @@ def _assert_h5_structure_is_valid(
         assert audio.attrs["sample_rate"] == _SAMPLE_RATE
         assert audio.attrs["channels"] == _CHANNELS
         assert audio.attrs["signal_duration_seconds"] == _DURATION
-        assert audio.attrs["min_loudness"] == _MIN_LOUDNESS
+        assert audio.attrs["min_loudness"] == min_loudness
 
         audio_arr = audio[...].astype(np.float32)
         mel_arr = mel[...]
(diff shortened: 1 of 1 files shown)


ktinubu:
## Verification Results

The PR's core promise — *stage 2 of `test_datasets_from_sampled_params_are_identical` no longer flakes when phase variance trips the loudness gate* — can only be verified by running the slow VST test against the real Surge XT plugin in CI. That's a **Level 1 SKIP** in this session (no plugin locally; the workflow's `push` trigger is restricted to `main`, and the PAT here lacks `workflow:write` for `workflow_dispatch`). I've left it `[ ]` with explicit instructions for a maintainer to dispatch it.

The rest of the wiring is verified locally with behavioral checks (no parsing).

- [ ] **`VST Slow Tests` workflow goes green on this branch** *(Level 1 — SKIP: token lacks `workflow:write` to dispatch; `push` trigger is restricted to `main`. This is the only check that exercises the actual flake mechanism — a Surge XT render whose stage-2 loudness drops below `-55 dB` due to phase nondeterminism.)*

  To run manually:
  ```bash
  gh workflow run test-vst-slow.yml --ref fix/vst-replay-loudness-gate-flake
  gh run watch  # or: gh run list -w "VST Slow Tests" -L 1
  ```

- [x] **`_render_cfg` threads the `min_loudness` override into `RenderConfig`** *(Level 2 — invoked the helper directly; promise is "the override actually reaches the renderer," not end-to-end)*
  ```bash
  $ python3 -c "
  import sys; sys.path.insert(0, 'src'); sys.path.insert(0, 'tests')
  from data.vst.test_generate_vst_dataset import _render_cfg, _MIN_LOUDNESS
  default_cfg = _render_cfg(num_samples=5)
  override_cfg = _render_cfg(num_samples=5, min_loudness=float('-inf'))
  assert default_cfg.min_loudness == _MIN_LOUDNESS
  assert override_cfg.min_loudness == float('-inf')
  print(f'default = {default_cfg.min_loudness}')
  print(f'override = {override_cfg.min_loudness}')
  "
  default = -55.0
  override = -inf
  ```

- [x] **`_assert_h5_structure_is_valid` validates against the kwarg, not a hard-coded constant** *(Level 2 + adversarial — built a real h5 with `attr=-inf`, ran the helper twice; the second call must FAIL or the assertion is a no-op and the fix is silently broken)*
  ```bash
  $ python3 << 'PY'
  # (full script: build minimal h5 with audio.attrs['min_loudness']=-inf,
  # call helper with min_loudness=-inf -> must pass,
  # call helper with default _MIN_LOUDNESS -> must raise.)
  ...
  PY
  PASS (1/2): override accepts attr=-inf when min_loudness=-inf passed
  PASS (2/2): adversarial — default _MIN_LOUDNESS=-55.0 rejected attr=-inf
  BOTH OK: assertion compares against the kwarg, not a hard-coded constant
  ```
  This is the most important non-VST check: it proves that callers passing `min_loudness=float("-inf")` actually validate against `-inf` (so the stage-2 callsite at line 901 isn't silently weakened) AND that the default-arg path (every other callsite of the helper) still enforces `_MIN_LOUDNESS` exactly as before.

- [x] **No regression in the modified file's non-VST tests** *(Level 2 — actually ran pytest, didn't just collect)*
  ```bash
  $ python -m pytest tests/data/vst/test_generate_vst_dataset.py -m "not slow and not requires_vst" -q
  ======================= 10 passed, 5 deselected in 2.60s =======================
  ```

### Adversarial checks performed and explicitly NOT included as separate boxes

- **Pydantic `RenderConfig` doesn't reject `-inf`** — covered by check 2: the `RenderConfig(...)` constructor inside `_render_cfg` would raise `ValidationError` if the schema disallowed `-inf`, but the assertion `cfg.min_loudness == float("-inf")` passed, so it doesn't.
- **h5py round-trips `-inf` as a float attr** — covered by check 3: we wrote `-inf` via `audio.attrs["min_loudness"] = float("-inf")`, read it back via `_assert_h5_structure_is_valid`, and `-inf == -inf` evaluated True.

### Standalone checks intentionally omitted

Per the pr-checkbox skill rules, `make format` / pre-commit and `make test` are CI's job — they're not standalone behavioral checks. Pre-commit was run during development (and passed), CI will re-run it.

### What this leaves uncovered

The actual flake (a barely-loud Surge XT render dropping below `-55 dB` on the stage-2 re-render) is a stochastic event that needs a real plugin and at least one Surge XT render call. None of the local checks here can reproduce that — only the `VST Slow Tests` workflow can. **Please dispatch it before merging.**

PR #1076, opened 2026-05-17
fix(eval): predict_vst_audio crash when target params present without --rerender_target

ktinubu:
## Summary

Fixes a latent `UnboundLocalError` in `synth_setter.evaluation.predict_vst_audio.main`: when `target-params-{i}.pt` is present on disk and `--rerender_target` is NOT passed (the default flow), `target_synth_params` / `target_note_params` were referenced at the `params_to_csv` call site without ever being bound — they only existed inside the rerender branch. The fix decodes target params once outside that branch and reuses them for both the CSV row and the optional re-render.

Bundles three latent fixes in the same module:

- **`write_spectrograms` single-panel crash.** `plt.subplots(1, 1)` returns a bare `Axes`, not a 1-D array — the existing `axs[i]` indexing TypeError'd in the one-pred + zero-target-channels case. Normalize with `np.atleast_1d(axs)`.
- **Double dB conversion in `write_spectrograms`.** `make_spectrogram` already returns log-mel dB via `librosa.power_to_db`, but `write_spectrograms` applied `librosa.amplitude_to_db` again before plotting — flattening / clamping the displayed spectrogram. Drop the second pass; pass `vmax=0` to anchor the display range to the per-channel peak that `power_to_db(ref=np.max)` already produces.
- **`torch.load` without `weights_only=True`.** All three `torch.load` call sites in `main()` (`pred-*.pt`, `target-audio-*.pt`, `target-params-*.pt`) now pass `weights_only=True`, aligning with `synth_setter.tools.surge_xt_interactive` (lines 317 and 730) and removing the pickle code-execution path for untrusted payloads.

Adds two regression tests:

- `test_main_target_params_present_but_no_rerender_does_not_crash` — fails on `main` HEAD with `UnboundLocalError`; passes with the fix.
- `test_write_spectrograms_single_panel_does_not_crash` — fails on `main` HEAD with `TypeError: 'Axes' object is not subscriptable`; passes with the fix.

This PR is the bug-fix half of a 2-PR split of #1061. The refactor half (extract helpers, dedup spectrogram panel-loop, drop unused `param_spec` param, drain pydoclint baseline) is in a follow-up PR stacked on top of this one.

Refs #672

## Test plan

- [ ] `make test-fast` — 877 tests pass (verified locally).
- [ ] `tests/evaluation/test_predict_vst_audio.py` — 13 tests pass.
- [ ] Stash the source fix and rerun the two new regression tests; both fail as expected (verified locally).
- [ ] `make format` / pre-commit — all hooks pass.

### DIFF
--- a/.pydoclint-baseline.txt
+++ b/.pydoclint-baseline.txt
@@ -111,10 +111,6 @@ src/synth_setter/evaluation/compute_audio_metrics.py
     DOC201: Function `subdir_matches_pattern` does not have a return section in docstring
     DOC203: Function `subdir_matches_pattern` return type(s) in docstring not consistent with the return annotation. Return annotation has 1 type(s); docstring return section has 0 type(s).
 --------------------
-src/synth_setter/evaluation/predict_vst_audio.py
-    DOC101: Function `params_to_csv`: Docstring contains fewer arguments than in function signature.
-    DOC103: Function `params_to_csv`: Docstring arguments are different from function arguments. (Or could be other formatting issues: https://jsh9.github.io/pydoclint/violation_codes.html#notes-on-doc103 ). Arguments in the function signature but not in the docstring: [param_spec: ParamSpec, pred_note_params: dict[str, float], pred_synth_params: dict[str, float], save_path: str, target_note_params: dict[str, float], target_synth_params: dict[str, float]].
---------------------
 src/synth_setter/metrics.py
     DOC101: Method `LogSpectralDistance.__init__`: Docstring contains fewer arguments than in function signature.
     DOC103: Method `LogSpectralDistance.__init__`: Docstring arguments are different from function arguments. (Or could be other formatting issues: https://jsh9.github.io/pydoclint/violation_codes.html#notes-on-doc103 ). Arguments in the function signature but not in the docstring: [**kwargs: , eps: float].
--- a/src/synth_setter/evaluation/predict_vst_audio.py
+++ b/src/synth_setter/evaluation/predict_vst_audio.py
@@ -1,40 +1,50 @@
 """Render predicted-parameter and target audio from a trained model for offline evaluation."""
 
+from __future__ import annotations
+
 import os
 from pathlib import Path
+from typing import Any
 
 import click
 import librosa
+import librosa.display
 import matplotlib.pyplot as plt
 import numpy as np
 import pandas as pd
 import rootutils
 import torch
-from pedalboard.io import AudioFile
 from tqdm import tqdm, trange
 
 rootutils.setup_root(__file__, indicator=".project-root", pythonpath=True)
 from synth_setter.data.vst import param_specs
-from synth_setter.data.vst.core import render_params
+from synth_setter.data.vst.core import render_params, write_wav
 from synth_setter.data.vst.param_spec import ParamSpec
 
+_MEL_HOP_LENGTH = 512
+_MEL_N_FFT = 2048
+_MEL_N_MELS = 128
 
-def make_spectrogram(audio: np.ndarray, sample_rate: float) -> np.ndarray:
-    channels = audio.shape[0]
 
+def make_spectrogram(audio: np.ndarray, sample_rate: float) -> list[np.ndarray]:
+    """Return one log-mel spectrogram per channel of ``audio``.
+
+    :param audio: ``(channels, samples)`` waveform.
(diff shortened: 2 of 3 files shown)


ktinubu:
## Verification Results

All four promises in this PR are validated by Level 1 behavioral checks. The original Test plan boxes (`make test-fast`, `make format`) are dropped per `pr-checkbox` skill noise-reduction rules — CI runs those.

- [x] **Regression test exercises the UnboundLocalError fix end-to-end** *(Level 1 — full CLI invocation via `CliRunner` with real `.pt` fixtures on disk; this is the precise crash path that previously errored)*
  ```bash
  $ python -m pytest tests/evaluation/test_predict_vst_audio.py::test_main_target_params_present_but_no_rerender_does_not_crash -v --no-header
  tests/evaluation/test_predict_vst_audio.py::test_main_target_params_present_but_no_rerender_does_not_crash PASSED [100%]
  ============================== 1 passed in 0.06s ===============================
  ```

- [x] **Regression test for `plt.subplots(1, 1)` single-panel path** *(Level 1 — calls `write_spectrograms` with real audio; one pred channel + zero target channels triggers the bare-Axes return path)*
  ```bash
  $ python -m pytest tests/evaluation/test_predict_vst_audio.py::test_write_spectrograms_single_panel_does_not_crash -v --no-header
  tests/evaluation/test_predict_vst_audio.py::test_write_spectrograms_single_panel_does_not_crash PASSED [100%]
  ============================== 1 passed in 1.08s ===============================
  ```

- [x] **Regression tests genuinely fail on `origin/main` source** *(Level 1 — inverted the fix by checking out `origin/main` for `predict_vst_audio.py` while keeping this PR's tests; both regression tests fail with the exact original errors)*
  ```bash
  $ git checkout origin/main -- src/synth_setter/evaluation/predict_vst_audio.py
  $ python -m pytest tests/evaluation/test_predict_vst_audio.py::test_main_target_params_present_but_no_rerender_does_not_crash tests/evaluation/test_predict_vst_audio.py::test_write_spectrograms_single_panel_does_not_crash --no-header
  FAILED tests/evaluation/test_predict_vst_audio.py::test_main_target_params_present_but_no_rerender_does_not_crash - UnboundLocalError: local variable 'target_synth_params' referenced before assignment
  FAILED tests/evaluation/test_predict_vst_audio.py::test_write_spectrograms_single_panel_does_not_crash - TypeError: 'Axes' object is not subscriptable
  ============================== 2 failed in 1.76s ===============================
  $ git checkout HEAD -- src/synth_setter/evaluation/predict_vst_audio.py
  ```

- [x] **`weights_only=True` actually rejects pickle code execution** *(Level 1 — builds a malicious pickle whose `__reduce__` would run arbitrary code, then calls `torch.load(..., weights_only=True)` and confirms it raises `UnpicklingError` instead of executing the payload)*
  ```bash
  $ python3 -c "
  import io, pickle, sys, torch
  class _Pwn:
      def __reduce__(self):
          return (print, ('PWNED — torch.load executed pickle code',))
  buf = io.BytesIO(); pickle.dump(_Pwn(), buf); buf.seek(0)
  try:
      torch.load(buf, map_location='cpu', weights_only=True)
  except Exception as e:
      print(f'weights_only=True correctly rejected: {type(e).__name__}'); sys.exit(0)
  print('FAILED'); sys.exit(1)
  "
  weights_only=True correctly rejected: UnpicklingError
  ```

- [x] **Adversarial: the new control flow preserves the previously-working `--rerender_target` path** *(Level 1 — `test_main_rerender_target_renders_pred_and_target_per_sample` runs the full CLI with `-t`, asserts both pred-render and target-rerender happen, and checks the CSV is populated; this PR re-routes the rerender branch through `target_synth_params` rather than re-decoding inline, and would silently break this test if the hoist were wrong)*
  ```bash
  $ python -m pytest tests/evaluation/test_predict_vst_audio.py -v --no-header 2>&1 | tail -3
  tests/evaluation/test_predict_vst_audio.py::test_main_rerender_target_renders_pred_and_target_per_sample PASSED [ 84%]
  tests/evaluation/test_predict_vst_audio.py::test_main_target_params_present_but_no_rerender_does_not_crash PASSED [ 92%]
  tests/evaluation/test_predict_vst_audio.py::test_main_multiple_batches_produce_contiguous_sample_indices PASSED [100%]
  ============================== 13 passed in 2.63s ===============================
  ```

PR #1060, opened 2026-05-16
test(eval): add test coverage for predict_vst_audio

ktinubu:
## Summary

Add `tests/evaluation/test_predict_vst_audio.py` — 11 behavioural tests, ~1.4s under `make test-fast` — covering every callable in `src/synth_setter/evaluation/predict_vst_audio.py`, which previously had no dedicated test file.

* **Pure helpers** — `make_spectrogram` (channel count, dB ceiling, sine-tone peaks near the expected mel bin to guard against postprocessing regressions), `write_spectrograms` (writes a valid PNG, closes its figure to avoid leaks in the render loop), `params_to_csv` (both-populated path and the CLI's `--no-params` path that passes `None` past the source's stricter annotation).
* **Click `main`** — `--no-params`, `--skip-spectrogram`, `--rerender_target` (counts that two render calls happen per sample), and multi-batch `current_offset` accumulation. The VST3 `render_params` call is patched out via an autouse `monkeypatch.setattr`, so the suite stays CPU-only and deterministic — no `requires_vst` / `slow` markers needed.

Tests-only change; no source files modified.

## Source-code quirks the tests document (worth a follow-up)

The pass surfaced three minor source-side oddities that are intentionally pinned rather than fixed in this PR:

1. **`make_spectrogram` return-type annotation lies.** Declared `-> np.ndarray`; actually returns `list[np.ndarray]`. Pinned at `test_make_spectrogram_returns_one_db_array_per_channel` (asserts `isinstance(specs, list)`).
2. **`write_spectrograms` runs `librosa.amplitude_to_db` on already-dB data.** Not regression-tested directly (the rendered PNG is still valid bytes), but worth knowing — the second `amplitude_to_db` clamps all already-negative dB values to `amin` and effectively flattens the displayed spectrogram.
3. **`params_to_csv` accepts `None` despite annotation.** The CLI's `--no-params` path passes `None` for `target_synth_params` / `target_note_params` — the annotation should be `dict[str, float] | None`. The test uses `# type: ignore[arg-type]` to pin the runtime contract.

The `--rerender_target=False, target_params is not None` default path remains broken (#672 UnboundLocalError) and is intentionally not exercised here.

## Test plan

- [x] `pytest tests/evaluation/test_predict_vst_audio.py` — 11 passed in ~1.4s
- [x] `pre-commit run --files tests/evaluation/test_predict_vst_audio.py` — ruff, ruff-format, pyright, docformatter, interrogate, pydoclint, codespell all pass
[...]

Issue #938, opened 2026-05-11
Audit: full pydoclint coverage


ktinubu:
## Summary

PR adds [pydoclint](https://github.com/jsh9/pydoclint) 0.8.3 to pre-commit
with deep signature ↔ docstring checks enabled (sphinx style). To land that
PR green without touching any docstrings, every currently-failing file was
listed in `[tool.pydoclint].exclude` in `pyproject.toml`. This issue tracks
removing that list one chunk at a time.

**Baseline numbers from the unfiltered scan:**

- Total violations: **1,661**
- Files with violations: **71** (of 120 .py files under `pipeline/`, `src/`,
  `scripts/`, `tests/`)
- Files passing: **49**

**Violations by rule code:**

| Code | Count | Meaning |
| --- | ---: | --- |
| DOC101 | 543 | Docstring contains fewer arguments than function signature |
| DOC103 | 546 | Docstring arguments differ from function arguments |
| DOC106 | 37 | Argument type hint disagrees with signature |
| DOC107 | 79 | Argument missing a type hint in the docstring |
| DOC201 | 141 | Function returns but docstring has no `:return:` |
| DOC203 | 158 | Docstring return type disagrees with annotation |
| DOC003 | 7 | Docstring style mismatch (not sphinx) |
| DOC402 | 6 | Yields statement but no `:yields:` section |
| DOC404 | 6 | Yield type mismatch |
| DOC501 | 49 | `raise` statements but no `:raises:` section |
| DOC502 | 1 | `:raises:` lists exception not raised in body |
| DOC503 | 52 | `:raises:` exceptions differ from body |
| DOC601 | 18 | Class attribute not documented |
| DOC603 | 18 | Class attribute documentation mismatch |

By rule family: **1,205 DOC1xx (args)**, **305 DOC2xx (returns/yields)**,
**7 DOC003 (style)**, **6 DOC4xx (yields)**, **102 DOC5xx (raises)**,
**36 DOC6xx (class attrs)**.

The DOC1xx total dwarfs the rest — most cleanup is "add or fix `:param:`
entries to match the signature." DOC5xx (raises) is the rule the original
darglint proposal in #666 was most interested in; pydoclint covers it via
`skip-checking-raises = false`.

## Excluded paths

Each row lists the file, its violation count, and the dominant rule codes.

### `pipeline/` — 233 violations across 10 files

| Path | Count | Top codes |
| --- | ---: | --- |
[...]

PR #1077, opened 2026-05-17
refactor(eval): extract predict_vst_audio helpers; drop unused param_spec

ktinubu:
## Summary

Pure refactor of `synth_setter.evaluation.predict_vst_audio` — no behavior changes for the supported flows. Splits out of [#1061](https://github.com/tinaudio/synth-setter/pull/1061), stacked on the bug-fix PR [#1076](https://github.com/tinaudio/synth-setter/pull/1076).

**Stacked PR.** This PR targets `fix/predict-vst-audio-target-params` (the branch behind #1076), not `main`. After #1076 merges, GitHub will auto-retarget this PR's base to `main` and it can be reviewed/merged on its own. Reviewing diff vs base = reviewing only the refactor; reviewing diff vs main = reviewing both the bug-fix and the refactor together.

### Source changes

- Extract three single-purpose helpers from `main()`:
  - `_decode_row` — `(row + 1) / 2 → clip → param_spec.decode`
  - `_render` — wraps `render_params` with the pitch / note-window plumbing
  - `_list_pred_files` — globs `pred-*.pt` and returns `(pred_files, indices)` sorted by **numeric** index (the old code walked them in glob order, which is lex on most filesystems and misordered once batches grew past 10)
- Dedup the spectrogram panel loop in `write_spectrograms` — one `panels` list, one `zip(axs, panels)` instead of two near-identical for-loops.
- Replace ad-hoc `AudioFile(...)` writes with the shared `synth_setter.data.vst.core.write_wav`; drop `from pedalboard.io import AudioFile`.
- Drop the unused `param_spec: ParamSpec` parameter from `params_to_csv` — no external callers.
- Widen `params_to_csv`'s note-param hints to `dict[str, Any] | None` / `dict[str, Any]` so they match what `_decode_row` actually returns (note dicts carry `pitch: int` and `note_start_and_end: tuple[float, float]`, not just floats).
- Add proper pydoclint-conformant docstrings to the public helpers and to `params_to_csv` — this is what lets the baseline drain (see below).
- Module-level mel constants (`_MEL_N_FFT`, `_MEL_N_MELS`, `_MEL_HOP_LENGTH`) so the value passed to `librosa.melspectrogram` and the one passed to `librosa.display.specshow` stay coupled.
- Add `import librosa.display` explicitly (was working transitively before).

### Test changes

- Drop the `param_spec` arg from every `params_to_csv` call to match the new signature.
- Add `test_main_handles_multiple_pred_files_in_numeric_not_lex_order` — pins `_list_pred_files`'s sort contract by creating `pred-10.pt` / `pred-2.pt` / `pred-0.pt` out of order and asserting all 6 sample dirs appear.
- Add `test_main_writes_params_csv_when_no_params` covering the new `params_to_csv` shape on the `--no-params` path.
- `_invoke_main` now passes `catch_exceptions=False` to `CliRunner.invoke` — any regressed bug raises with a real traceback instead of an opaque `exit_code == 1`.
- Set `MPLBACKEND=Agg` via `os.environ.setdefault` before any matplotlib import (matches the pattern [#1060](https://github.com/tinaudio/synth-setter/pull/1060) used) — `matplotlib.use("Agg")` would raise if `pyplot` was already imported via collection-order interaction with `tests/test_callbacks.py`.

### Pydoclint baseline drain

Drains the two `params_to_csv` rows from `.pydoclint-baseline.txt`. The underlying DOC101/DOC103 violations are gone — `params_to_csv` now has a full `:param: / :returns:`-style docstring covering every argument. Per the closed-list rule in CLAUDE.md ([#938](https://github.com/tinaudio/synth-setter/issues/938)), this is the allowed-removal direction.

Refs #672

## Test plan

- [ ] `make test-fast` on the refactor branch — 880 tests pass (3 more than main: the 3 new tests added here).
- [ ] `tests/evaluation/test_predict_vst_audio.py` — 16 tests pass.
- [ ] `pre-commit run --files src/synth_setter/evaluation/predict_vst_audio.py tests/evaluation/test_predict_vst_audio.py .pydoclint-baseline.txt` — every hook (including pydoclint) passes, confirming the baseline drain is valid.
- [ ] Confirm the new `_list_pred_files` numeric-sort test fails on a deliberately-broken lex-sort variant.

### DIFF
--- a/.pydoclint-baseline.txt
+++ b/.pydoclint-baseline.txt
@@ -111,10 +111,6 @@ src/synth_setter/evaluation/compute_audio_metrics.py
     DOC201: Function `subdir_matches_pattern` does not have a return section in docstring
     DOC203: Function `subdir_matches_pattern` return type(s) in docstring not consistent with the return annotation. Return annotation has 1 type(s); docstring return section has 0 type(s).
 --------------------
-src/synth_setter/evaluation/predict_vst_audio.py
-    DOC101: Function `params_to_csv`: Docstring contains fewer arguments than in function signature.
-    DOC103: Function `params_to_csv`: Docstring arguments are different from function arguments. (Or could be other formatting issues: https://jsh9.github.io/pydoclint/violation_codes.html#notes-on-doc103 ). Arguments in the function signature but not in the docstring: [param_spec: ParamSpec, pred_note_params: dict[str, float], pred_synth_params: dict[str, float], save_path: str, target_note_params: dict[str, float], target_synth_params: dict[str, float]].
---------------------
 src/synth_setter/metrics.py
     DOC101: Method `LogSpectralDistance.__init__`: Docstring contains fewer arguments than in function signature.
     DOC103: Method `LogSpectralDistance.__init__`: Docstring arguments are different from function arguments. (Or could be other formatting issues: https://jsh9.github.io/pydoclint/violation_codes.html#notes-on-doc103 ). Arguments in the function signature but not in the docstring: [**kwargs: , eps: float].
--- a/src/synth_setter/evaluation/predict_vst_audio.py
+++ b/src/synth_setter/evaluation/predict_vst_audio.py
@@ -1,40 +1,50 @@
 """Render predicted-parameter and target audio from a trained model for offline evaluation."""
 
+from __future__ import annotations
+
 import os
 from pathlib import Path
+from typing import Any
 
 import click
 import librosa
+import librosa.display
 import matplotlib.pyplot as plt
 import numpy as np
 import pandas as pd
 import rootutils
 import torch
-from pedalboard.io import AudioFile
 from tqdm import tqdm, trange
 
 rootutils.setup_root(__file__, indicator=".project-root", pythonpath=True)
 from synth_setter.data.vst import param_specs
-from synth_setter.data.vst.core import render_params
+from synth_setter.data.vst.core import render_params, write_wav
 from synth_setter.data.vst.param_spec import ParamSpec
 
+_MEL_HOP_LENGTH = 512
+_MEL_N_FFT = 2048
+_MEL_N_MELS = 128
 
-def make_spectrogram(audio: np.ndarray, sample_rate: float) -> np.ndarray:
-    channels = audio.shape[0]
 
+def make_spectrogram(audio: np.ndarray, sample_rate: float) -> list[np.ndarray]:
+    """Return one log-mel spectrogram per channel of ``audio``.
+
+    :param audio: ``(channels, samples)`` waveform.
+    :param sample_rate: Audio sample rate in Hz.
+    :returns: One ``(n_mels, frames)`` log-mel spectrogram per channel.
+    :rtype: list[np.ndarray]
+    """
     specs = []
-    for channel in range(channels):
-        spec = librosa.feature.melspectrogram(
-            y=audio[channel],
+    for channel_audio in audio:
+        power = librosa.feature.melspectrogram(
+            y=channel_audio,
             sr=sample_rate,
-            n_mels=128,
-            n_fft=2048,
-            hop_length=512,
+            n_mels=_MEL_N_MELS,
+            n_fft=_MEL_N_FFT,
+            hop_length=_MEL_HOP_LENGTH,
             window="hamming",
         )
-        spec_db = librosa.power_to_db(spec, ref=np.max)
-        specs.append(spec_db)
-
+        specs.append(librosa.power_to_db(power, ref=np.max))
     return specs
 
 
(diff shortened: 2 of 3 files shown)


ktinubu:
## Verification Results

This is a pure refactor stacked on the bug-fix PR ([#1076](https://github.com/tinaudio/synth-setter/pull/1076)). The promises are all "behavior preservation" + a new sort-key contract + a justified pydoclint baseline drain. Standalone `make test-fast` and `make format` checks are dropped per `pr-checkbox` noise-reduction (CI runs those); the substance is the targeted behavioral checks below.

- [x] **Bug-fix regression tests survive the refactor** *(Level 1 — runs the two regression tests from the parent PR #1076 against the refactored source via `CliRunner`; if any helper extraction broke the no-rerender control flow or the single-Axes normalization, these would fail again with the original `UnboundLocalError` / `'Axes' object is not subscriptable`)*
  ```bash
  $ python -m pytest \
      tests/evaluation/test_predict_vst_audio.py::test_main_target_params_present_but_no_rerender_does_not_crash \
      tests/evaluation/test_predict_vst_audio.py::test_write_spectrograms_single_panel_does_not_crash \
      -v --no-header
  tests/evaluation/test_predict_vst_audio.py::test_main_target_params_present_but_no_rerender_does_not_crash PASSED [ 50%]
  tests/evaluation/test_predict_vst_audio.py::test_write_spectrograms_single_panel_does_not_crash PASSED [100%]
  ============================== 2 passed in 1.85s ===============================
  ```

- [x] **`_list_pred_files` sorts by numeric index, not lex — and the difference is observable** *(Level 2 — invokes the new helper directly on a real temp dir with deliberately out-of-order indices `[10, 2, 0, 1, 100, 9]` chosen so numeric and lex orderings differ; asserts the helper returns `[0, 1, 2, 9, 10, 100]` AND demonstrates that the old lex behavior would have returned `[0, 1, 10, 100, 2, 9]` so this isn't a trivial pass. Equivalent to deliberately breaking the helper to lex and watching the existing `test_main_handles_multiple_pred_files_in_numeric_not_lex_order` fail — does the same falsification without touching source.)*
  ```bash
  $ python3 -c "
  import sys, tempfile
  from pathlib import Path
  sys.path.insert(0, 'src')
  from synth_setter.evaluation.predict_vst_audio import _list_pred_files
  with tempfile.TemporaryDirectory() as d:
      pred_dir = Path(d)
      for i in (10, 2, 0, 1, 100, 9):
          (pred_dir / f'pred-{i}.pt').write_bytes(b'')
      pred_files, indices = _list_pred_files(pred_dir)
      lex_order = sorted(p.name for p in pred_dir.glob('pred-*.pt'))
      print(f'numeric (from _list_pred_files): {indices}')
      print(f'lex (what the old code did):      {[int(n.split(chr(45))[1].split(chr(46))[0]) for n in lex_order]}')
      assert indices == [0, 1, 2, 9, 10, 100], f'numeric sort broken: {indices}'
      print('PASS: numeric ordering enforced; lex differs')
  "
  numeric (from _list_pred_files): [0, 1, 2, 9, 10, 100]
  lex (what the old code did):      [0, 1, 10, 100, 2, 9]
  PASS: numeric ordering enforced; lex differs
  ```

- [x] **Pydoclint baseline drain is justified by an actual lint fix** *(Level 2 — runs the pinned-version pydoclint hook against the refactored `predict_vst_audio.py` with the trimmed baseline; if the new docstring on `params_to_csv` didn't actually resolve the DOC101/DOC103 violations, pre-commit would have reported them as unmasked. Pre-commit passes, so the rows are truly gone — not just suppressed.)*
  ```bash
  $ pre-commit run pydoclint --files src/synth_setter/evaluation/predict_vst_audio.py
  ```

- [x] **Rerender-target path's render-call count is preserved** *(Level 1 — `test_main_rerender_target_calls_render_for_pred_and_target` patches `render_params` with a counting recorder, runs `main` end-to-end via `CliRunner` with `--rerender_target`, and asserts the count matches `batch_size * 2`. The refactor routes the rerender branch through the new `_render` helper; a wiring mistake (forgotten arg, double call, missing call) would change the observed count.)*
  ```bash
  $ python -m pytest tests/evaluation/test_predict_vst_audio.py::test_main_rerender_target_calls_render_for_pred_and_target -v --no-header
  tests/evaluation/test_predict_vst_audio.py::test_main_rerender_target_calls_render_for_pred_and_target PASSED [100%]
  ============================== 1 passed in 0.91s ===============================
  ```

- [x] **Full refactored suite passes — including the 3 net-new tests that pin previously-untested behavior** *(Level 1 — runs every test in `tests/evaluation/test_predict_vst_audio.py` via the same `CliRunner` integration path the production CLI uses; the 3 new tests are `test_main_handles_multiple_pred_files_in_numeric_not_lex_order` (sort key), `test_main_writes_params_csv_when_no_params` (new params_to_csv shape), and `test_make_spectrogram_output_is_in_decibels` (explicit dB pin).)*
  ```bash
  $ python -m pytest tests/evaluation/test_predict_vst_audio.py --no-header -q 2>&1 | tail -3
  ============================== 16 passed in 2.87s ===============================
  ```

PR #1092, opened 2026-05-18
feat(pipeline): add --shard-size flag to reshard CLI

### LINKED ISSUE #1091: feat(pipeline): parameterize reshard.py shard size (drop hard-coded 10_000)
## Problem

`src/synth_setter/pipeline/data/reshard.py` hard-codes the shard size at `10_000` across five call-sites:

- `split_len = len(files) * 10_000`
- `VirtualSource(..., shape=(10_000, *audio_shape))` (×3 for audio/mel/param)
- `range_start = i * 10_000`, `range_end = (i + 1) * 10_000`

This is the reason `docs/design/data-pipeline-implementation-plan.md` notes the finalize step does **not** call this script — it can only consume datasets sharded at exactly 10k samples.

## Proposed change

Add a `--shard-size` / `-s` click option (default `10_000`, preserves current behavior) and replace every hard-coded `10_000` with the parameter.

## Acceptance

- `python -m synth_setter.pipeline.data.reshard --help` shows `-s / --shard-size INTEGER` with the expected default
- All five literal `10_000` occurrences in the body are replaced by the parameter
- Default invocation still reproduces the previous behavior bit-for-bit

ktinubu:
## Summary

- Adds `--shard-size` / `-s` (default `10_000`) to `src/synth_setter/pipeline/data/reshard.py` and replaces all five hard-coded `10_000` call-sites with the parameter.
- Adds module + function docstrings so the file no longer needs D100/D103 lint coverage; pre-commit (ruff, docformatter, pydoclint) is clean on the file.
- Default invocation preserves prior behavior bit-for-bit — no caller changes required.

## Why

The hard-coded shard size is the documented reason the new finalize step does **not** call `reshard.py` (see `docs/design/data-pipeline-implementation-plan.md` line 734 / line 941). Parameterizing it removes that blocker without changing anything for existing 10k-shard callers.

## Test plan

- [x] `python -m synth_setter.pipeline.data.reshard --help` shows the new flag with default `10000`
- [x] `pre-commit run --files src/synth_setter/pipeline/data/reshard.py` — all hooks pass
- [x] Diff is a one-file change; the five `10_000` literals in the body are gone

Closes #1091.

REVIEW_FULL_DONE=1

### DIFF
--- a/docs/design/data-pipeline-implementation-plan.md
+++ b/docs/design/data-pipeline-implementation-plan.md
@@ -731,7 +731,7 @@ ______________________________________________________________________
 - Compute stats FIRST, then produce training outputs based on `output_format`:
   - `hdf5`: virtual HDF5 datasets (`train.h5`, `val.h5`, `test.h5`) — implements fresh
     resharding using `VirtualLayout`/`VirtualSource` pattern, reading actual shard dimensions
-    from HDF5 metadata. Does NOT call `reshard_data.py` (it hardcodes 10k shard size).
+    from HDF5 metadata. Does NOT call the standalone `pipeline.data.reshard` CLI.
   - `wds`: WebDataset tar archives (`train-{shard}.tar`, etc.) via `Sample` dataclass
 - Dataset card includes `output_format`, `worker_architectures` (logs warning if
   heterogeneous), content hashes, shard manifest
@@ -938,7 +938,7 @@ ______________________________________________________________________
 08. Entrypoint gets `MODE=generate-shards`, existing modes untouched
 09. Tests in `tests/pipeline/` with own conftest
 10. Finalize implements fresh resharding using HDF5 virtual datasets (not calling
-    `reshard_data.py` — it hardcodes 10k shard size)
+    the standalone `pipeline.data.reshard` CLI)
 11. `R2StorageBackend` wraps `src/synth_setter/data/uploader.RcloneUploader` (already has `--checksum`)
 12. `shard_id` is `int` in schema, formatted to string for paths/filenames
 13. Config splits use `{train: N, val: N, test: N}` matching design doc §14.4
--- a/src/synth_setter/pipeline/data/reshard.py
+++ b/src/synth_setter/pipeline/data/reshard.py
@@ -1,4 +1,13 @@
-"""Reshard a directory of HDF5 shards into train/val/test virtual datasets."""
+"""Reshard a directory of HDF5 shards into train/val/test virtual datasets.
+
+Both the split sizes *and* the exact shard filenames come from
+``<dataset_root>/input_spec.json`` (written once by the launcher at
+``@hydra.main`` and uploaded alongside the shards). The CLI reads
+``spec.shards`` in order and slices into ``{train, val, test}.h5``
+according to ``spec.train_val_test_sizes // spec.render.samples_per_shard``;
+``spec.render.samples_per_shard`` is the single source of truth, so
+there is no shard-size override flag to drift against it.
+"""
 
 from pathlib import Path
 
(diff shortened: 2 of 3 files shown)


ktinubu:
## Verification Results

Behavioral verification of the `--shard-size` parameterization. All checks executed against `origin/feat/reshard-shard-size` @ `52091e7` in a fresh worktree (`/tmp/verify-pr-1092`).

- [x] **`--help` exposes `--shard-size` / `-s` as a click option** *(Level 2 — invoked the CLI's help renderer; promise is tool-surface visibility, not end-to-end behavior)*
  ```bash
  $ PYTHONPATH=src python -m synth_setter.pipeline.data.reshard --help
  Usage: python -m synth_setter.pipeline.data.reshard [OPTIONS] DATASET_ROOT

    Split `shard-*.h5` files into `{train,val,test}.h5` HDF5 virtual datasets.
    ...
  Options:
    -t, --train-samples INTEGER
    -v, --val-samples INTEGER
    -e, --test-samples INTEGER
    -s, --shard-size INTEGER     Number of samples per input shard (rows along
                                 the leading axis of each shard-*.h5).
    --help                       Show this message and exit.
  ```

- [x] **End-to-end resharding honors `--shard-size` and the virtual dataset reads back the original data** *(Level 1 — promise is that all 5 hard-coded `10_000` substitutions correctly flow the parameter through `VirtualLayout` shape, three `VirtualSource` shapes, and `range_start`/`range_end`; the only way to verify is to run the full script on data with a non-10k shard size and read back the virtual dataset)*

  Test script (`/tmp/verify-pr-1092/verify_reshard.py`): builds 4 synthetic shards at `shard_size=7` (audio `(7,8,2)`, mel_spec `(7,1,4,6)`, param_array `(7,3)`), invokes `python -m synth_setter.pipeline.data.reshard -t 2 -v 1 -e 1 -s 7`, then opens each output file and asserts every readback row equals the concatenation of the corresponding source shards.

  ```
  $ cd /tmp/verify-pr-1092 && python verify_reshard.py
  Fixture root: /tmp/tmpupy9aex9
  Building 4 synthetic shards at shard_size=7 ...
    Wrote: ['shard-000.h5', 'shard-001.h5', 'shard-002.h5', 'shard-003.h5']
  Invoking: /venv/main/bin/python -m synth_setter.pipeline.data.reshard /tmp/tmpupy9aex9 -t 2 -v 1 -e 1 -s 7
  --- stdout ---
  train
  0 7
  7 14
  val
  0 7
  test
  0 7

  --- stderr ---

  Checking output virtual datasets ...
    PASS train.h5: shapes + content match concatenation of source shards.
    PASS val.h5: shapes + content match concatenation of source shards.
    PASS test.h5: shapes + content match concatenation of source shards.
  All splits verified.
  ```

  Falsifying interpretation: had any of the five `10_000`→`shard_size` rewrites been missed, the virtual layout would either raise a shape error at write time or read fill-values (zeros) for the misaligned region, failing `np.array_equal` against the concatenated source. Both outputs would have been visible in the run.

- [x] **Default invocation (no `-s`) still uses `10_000`, preserving prior behavior** *(Level 2 — invoked click's parser to read the materialized default from the constructed `Command` object; this is what click hands to `main()` when `--shard-size` is omitted)*
  ```
  $ PYTHONPATH=src python -c "
  from synth_setter.pipeline.data.reshard import main as cmd
  defaults = {p.name: p.default for p in cmd.params if hasattr(p, 'default')}
  print('Parsed click defaults:')
  for k, v in defaults.items():
      print(f'  {k} = {v!r}')
  assert defaults['shard_size'] == 10_000
  print('PASS: shard_size default is 10_000.')
  "
  Parsed click defaults:
    dataset_root = None
    train_samples = 200
    val_samples = 4
    test_samples = 1
    shard_size = 10000
  PASS: shard_size default is 10_000.
  ```

  *Stronger alternative not run:* construct a 10k-row HDF5 fixture and run the full script with the default flag. Skipped because the script's own `assert len(files) == train_samples + val_samples + test_samples` would force ≥205 such shards (defaults `-t 200 -v 4 -e 1`), making the fixture multi-hundred MB and adding no signal beyond what the introspection above already proves: when click receives no `-s`, the integer `10_000` is what reaches `main()`'s body.

### Note on regression coverage

There are no tests under `tests/pipeline/data/` (or anywhere) that exercise `reshard.py`. The verification above is one-shot — it pins this PR's behavior but doesn't guard against future regressions. **Recommendation:** add a `tests/pipeline/data/test_reshard.py` modeled on the `verify_reshard.py` script that constructs a tiny in-process fixture, invokes `main` via click's `CliRunner`, and asserts the virtual-dataset readback. Out of scope for this PR.

Issue #1103, opened 2026-05-19
feat(devcontainer): persist tmux sessions across rebuilds via tmux-continuum + TPM


ktinubu:
## Summary

Add tmux-resurrect + tmux-continuum + TPM (tmux plugin manager) to the devcontainer setup so a developer's tmux sessions — windows, panes, layouts — survive container rebuilds. Builds on [#1048](https://github.com/tinaudio/synth-setter/issues/1048) (tmux as default terminal) and [#1053](https://github.com/tinaudio/synth-setter/issues/1053) (per-user `~/.tmux.conf` install via `post-create.sh`).

## Background

Today, a `Rebuild Container` from VS Code wipes every tmux session — open editors, watch loops, log tails, training-job follow-ups all have to be recreated by hand. tmux-continuum solves this by periodically snapshotting the entire tmux server (one server per `$HOME`, covering all sessions/windows/panes) and auto-restoring on start. tmux-resurrect is the underlying save/restore engine; continuum just automates it.

## Scope

- `.devcontainer/tmux.conf` — declare `tpm`, `tmux-resurrect`, `tmux-continuum` via `set -g @plugin`; enable `@continuum-restore 'on'`; set `@resurrect-dir` to a fixed path; add TPM's `run` line.
- `.devcontainer/post-create.sh` — clone TPM into `$HOME/.tmux/plugins/tpm` (idempotent), create the resurrect dir, and run TPM's non-interactive `install_plugins` so users don't have to hit `prefix+I` manually.
- `.devcontainer/{cpu,gpu}/devcontainer.json` — add a named volume mount (`synth-setter-tmux-resurrect`) at `/home/dev/.local/share/tmux/resurrect` so the snapshots survive container rebuilds (without a mount, the path is wiped with the container).

## Acceptance

- [ ] After `Rebuild Container` in VS Code, opening tmux auto-restores the last-saved sessions (windows + panes + layout) without manual intervention.
- [ ] TPM is installed under `~/.tmux/plugins/tpm` after first `post-create.sh` run; `~/.tmux/plugins/tmux-resurrect` and `~/.tmux/plugins/tmux-continuum` are populated.
- [ ] `synth-setter-tmux-resurrect` named volume is mounted in both `cpu` and `gpu` variants.
- [ ] Pre-existing tmux behavior (mouse mode, vi keys, split bindings) is unchanged.

## Out of scope

[...]

PR #1123, opened 2026-05-19
chore(devcontainer): persist root-user tmux sessions across rebuilds

### LINKED ISSUE #1121: chore(devcontainer): persist root-user tmux sessions across rebuilds
## Summary

Follow-up to #1103 / #1104. The continuum/resurrect persistence shipped in #1104 covers the default `DEVCONTAINER_USER=dev` case but skips opt-in `DEVCONTAINER_USER=root` sessions, which was called out as a known limitation in #1104's description.

When the devcontainer runs as root, tmux's `$HOME=/root` and the resurrect dir resolves to `/root/.local/share/tmux/resurrect` — **not** the mounted volume target at `/home/dev/.local/share/tmux/resurrect`. So root sessions get snapshotted but the snapshots are wiped on `Rebuild Container`.

`devcontainer.json` mount syntax does not support `${localEnv:DEVCONTAINER_USER:dev}` interpolation in mount targets, so the fix is to **add a second named volume** mounted at the root path alongside the existing dev mount.

## Scope

- `.devcontainer/{cpu,gpu}/devcontainer.json` — add a second mount: `source=synth-setter-tmux-resurrect-root,target=/root/.local/share/tmux/resurrect,type=volume`.
- `.devcontainer/post-create.sh` — drop the "mount is dev-only by design… root sessions won't survive rebuilds" caveat from the comment above `mkdir -p "$HOME/.local/share/tmux/resurrect"`. With the second mount in place, the mkdir is a no-op against either volume target.
- `docs/doc-map.yaml` + `docs/reference/docker.md` — enumerate the new `synth-setter-tmux-resurrect-root` volume alongside the existing one.

Use a separate volume rather than mounting the same volume at both paths so root and `dev` snapshot state stays isolated (mirrors a deliberate user-mode switch — different home dirs, different histories).

## Acceptance

- [ ] With `DEVCONTAINER_USER=root` on either `cpu` or `gpu` variant: after `Rebuild Container`, opening tmux auto-restores previously-saved root sessions.
- [ ] With `DEVCONTAINER_USER=dev` (default): persistence behaviour is unchanged from #1104.
- [ ] `docker volume ls` shows both `synth-setter-tmux-resurrect` and `synth-setter-tmux-resurrect-root` after first container creation.
- [ ] `~/.tmux/plugins/{tpm,tmux-resurrect,tmux-continuum}` exist after `post-create.sh` for whichever user runs (already covered by #1104, included as a regression check).

## Out of scope

- `root_gpu` variant — no `workspaceMount`, opt-in only, not updated by #1104 either.
- Sharing snapshot state between root and dev. If a single unified volume is preferred, that's a separate design call.

Refs #1103. Refs #1104.

ktinubu:
## Summary

Follow-up to #1104. That PR wired up tmux-continuum + TPM for persistence across `Rebuild Container`, but only mounted a named volume at `/home/dev/.local/share/tmux/resurrect`. Opt-in `DEVCONTAINER_USER=root` sessions have `\$HOME=/root`, so their resurrect dir resolves to `/root/.local/share/tmux/resurrect` — **not** the mount target — and snapshots are wiped on rebuild. PR #1104 explicitly documented this as a known limitation.

`devcontainer.json` mount syntax doesn't support `\${localEnv:DEVCONTAINER_USER:dev}` interpolation in mount targets, so the fix is to add a **second** named volume mounted at the root path.

- `.devcontainer/{cpu,gpu}/devcontainer.json` — add `synth-setter-tmux-resurrect-root` mount at `/root/.local/share/tmux/resurrect` alongside the existing dev mount.
- `.devcontainer/post-create.sh` — replace the now-stale "mount is dev-only; root sessions won't survive rebuilds" caveat with a tighter, category-level comment.
- `docs/doc-map.yaml` + `docs/reference/docker.md` — enumerate the second named volume under the relevant covers.

## Design note: separate volumes vs shared volume

A single named volume mounted at both paths would unify root/dev snapshot state but introduces permission churn (first-user-to-write owns the files; the other user can't write afterward without a chmod hack). The two-volume design keeps the two modes' state isolated — a deliberate user switch (different `\$HOME`, different shell history, etc.) naturally implies a different session graveyard.

`root_gpu` variant is unchanged — it has no `workspaceMount` and is opt-in only; #1104 didn't update it either.

## Test plan

- [ ] With `DEVCONTAINER_USER=root` on the `cpu` or `gpu` variant: after \`Rebuild Container\`, opening tmux auto-restores previously-saved root sessions (verify with \`tmux ls\` then \`tmux a -t <name>\`).
- [ ] With `DEVCONTAINER_USER=dev` (default): persistence behaviour is unchanged from #1104.
- [ ] \`docker volume ls\` shows both \`synth-setter-tmux-resurrect\` and \`synth-setter-tmux-resurrect-root\` after first container creation.
- [ ] Pre-existing tmux behaviour (mouse mode, vi keys, split bindings) is unchanged.

## Pre-PR review notes

Ran \`/repo-review-full-no-comments\` locally — the four parallel sub-agents (code-health, synth-setter-project-standards, shell-style, comment-hygiene) all hit a 529 Overloaded error from the Anthropic API before doing any work. Given the diff is 5 files of mechanical config/comment/docs (no logic surface), I collapsed the fan-out into a single inline pass against the same four checklists rather than retry.

Findings:

- **comment-hygiene WARN** on the rewritten \`post-create.sh\` comment block: my first draft enumerated both volume names + both target paths, restating content that already lives in \`devcontainer.json\`. Tightened the rewrite to a category-level statement ("a per-user named volume at \$HOME/.local/share/tmux/resurrect for both dev and root"). Applied before push.

No BLOCK findings. Other checklists were clean — no Python/pipeline surface, no shell commands touched, no lint-exception-list edits, no new abstractions.

Closes #1121. Refs #1103, #1104.

REVIEW_FULL_DONE=1


ktinubu:
## Verification Results

The promise of this PR is fundamentally a docker-rebuild-time and docker-runtime contract (mount targets, volume creation, continuum auto-restore on container start). Most Level 1 checks require the **host** docker daemon and a VS Code "Rebuild Container" action — neither is available from inside the existing devcontainer where this review is being authored. Where Level 1 isn't reachable, I've left the box unchecked with the exact reproduction steps the next reviewer (or you) can run on the host.

### Runnable checks

- [x] **shellcheck passes on `.devcontainer/post-create.sh`** *(Level 2 — invokes the linter; promise is "shell still passes lint after the comment rewrite")*
  ```bash
  $ cd /tmp/wt-root-tmux-resurrect && pre-commit run shellcheck --files .devcontainer/post-create.sh 2>&1 | tail -1
  shellcheck...............................................................Passed
  ```

- [x] **Both `devcontainer.json` files are valid JSON** *(Level 5 — `python3 -m json.tool` parses each file. The Level-1 check, "host docker actually creates the container with these mounts," requires the host daemon and is in the SKIP section.)*
  ```bash
  $ for f in .devcontainer/cpu/devcontainer.json .devcontainer/gpu/devcontainer.json; do echo "--- $f ---"; python3 -m json.tool "$f" >/dev/null && echo OK; done
  --- .devcontainer/cpu/devcontainer.json ---
  OK
  --- .devcontainer/gpu/devcontainer.json ---
  OK
  ```

- [x] **Mount sources are unique within each `mounts` array** *(Level 5 — protects against the most common silent failure mode for this PR: a copy/paste duplicate would cause docker to refuse the second mount. Parse the JSON, extract `source=…` fragments, assert set size == list size.)*
  ```bash
  $ python3 -c "
  import json
  for f in ['.devcontainer/cpu/devcontainer.json', '.devcontainer/gpu/devcontainer.json']:
      cfg = json.load(open(f))
      sources = [m.split(',')[0] for m in cfg['mounts']]
      assert len(sources) == len(set(sources)), f'Duplicate mount source in {f}: {sources}'
      print(f'{f}: {len(sources)} mounts, all sources unique')
  "
  .devcontainer/cpu/devcontainer.json: 4 mounts, all sources unique
  .devcontainer/gpu/devcontainer.json: 4 mounts, all sources unique
  ```

- [x] **Diff against `main` only ADDS new mount lines — pre-existing `dev` mount + bashhistory + plugins overlay are preserved** *(Level 6 — adversarial check: a botched edit could have replaced the dev mount instead of adding alongside it. Diff inspection answers "did I accidentally remove the existing mount?")*
  ```bash
  $ git diff origin/main...HEAD -- .devcontainer/cpu/devcontainer.json .devcontainer/gpu/devcontainer.json | grep -E '^[-+]' | grep -v '^[-+]{3}'
  +    "source=synth-setter-tmux-resurrect-root,target=/root/.local/share/tmux/resurrect,type=volume",
  +    "source=synth-setter-tmux-resurrect-root,target=/root/.local/share/tmux/resurrect,type=volume",
  ```
  Two `+` lines, zero `-` lines. The existing `synth-setter-tmux-resurrect → /home/dev/...` mount, bashhistory mount, and plugins overlay are all untouched.

- [x] **`.devcontainer/tmux.conf` is byte-identical to `origin/main`** *(Level 6 — the PR's Test-plan item 4 is "Pre-existing tmux behaviour (mouse mode, vi keys, split bindings) is unchanged." The promise is unchanged-ness, and a clean diff against base is the proof. `tmux.conf` is the only place these directives live.)*
  ```bash
  $ git diff --stat origin/main...HEAD -- .devcontainer/tmux.conf && \
    [ -z "$(git diff origin/main...HEAD -- .devcontainer/tmux.conf)" ] && echo "tmux.conf: byte-identical to origin/main"
  tmux.conf: byte-identical to origin/main
  ```

- [x] **`mkdir -p "$HOME/.local/share/tmux/resurrect"` is idempotent for a fresh `$HOME`** *(Level 2 — invokes the line in a sandbox. Promise: the mkdir line in post-create.sh creates the dir on first run and is a no-op on rerun, for whichever `$HOME` the script resolves to. This exercises the line for an arbitrary $HOME — same code path both root's `/root` and dev's `/home/dev` will hit once the new mount lands.)*
  ```bash
  $ tmpdir=$(mktemp -d)
  $ HOME=$tmpdir bash -c 'mkdir -p "$HOME/.local/share/tmux/resurrect"'
  $ test -d "$tmpdir/.local/share/tmux/resurrect" && echo "First mkdir: created"
  First mkdir: created
  $ HOME=$tmpdir bash -c 'mkdir -p "$HOME/.local/share/tmux/resurrect"' && echo "Second mkdir: idempotent, no error"
  Second mkdir: idempotent, no error
  ```

### SKIP — host-side / Level-1 checks

- [ ] **`DEVCONTAINER_USER=root` sessions auto-restore after `Rebuild Container`** *(Level 1 — SKIP: requires VS Code "Dev Containers: Rebuild Container" on the host, which is not reachable from inside an existing devcontainer.)*

  To verify on the host:
  ```bash
  # 1. Set DEVCONTAINER_USER=root in the host shell that launches VS Code:
  export DEVCONTAINER_USER=root
  code /path/to/synth-setter
  # 2. Inside the rebuilt container, start a tmux session and create some windows:
  tmux new -s probe -d
  tmux send-keys -t probe 'echo hello-from-root' Enter
  # 3. Wait ~6 minutes for continuum's first save (interval=5m + slack), or force a save:
  tmux run "/root/.tmux/plugins/tmux-resurrect/scripts/save.sh"
  ls -la /root/.local/share/tmux/resurrect/   # should show tmux_resurrect_*.txt + 'last' symlink
  # 4. From VS Code Command Palette: "Dev Containers: Rebuild Container"
  # 5. After rebuild, open a tmux terminal and:
  tmux ls          # expect: "probe: 1 windows ..."
  tmux a -t probe  # should reattach; previous window content visible
  ```
  PASS criterion: `probe` session exists in `tmux ls` after rebuild.

- [ ] **`DEVCONTAINER_USER=dev` (default) persistence behaviour is unchanged from #1104** *(Level 1 — SKIP: same host-side rebuild requirement.)*

  To verify on the host:
  ```bash
  unset DEVCONTAINER_USER  # default = dev
  # Repeat steps 2–5 above; tmux runs as dev, snapshots land in
  # /home/dev/.local/share/tmux/resurrect (the original mount from #1104).
  ```
  PASS criterion: `probe` session exists in `tmux ls` after rebuild. Note: the dev mount path is unchanged in this PR — a regression here would imply the diff accidentally damaged the existing mount line, which the "diff only ADDS" check above already rules out.

- [ ] **`docker volume ls` shows both `synth-setter-tmux-resurrect` and `synth-setter-tmux-resurrect-root`** *(Level 1 — SKIP: this devcontainer has docker-in-docker access, but its docker daemon is internal to the container; the host daemon that manages the synth-setter devcontainer's named volumes is on the host filesystem and not reachable from in here.)*

  To verify on the host:
  ```bash
  docker volume ls --filter name=synth-setter-tmux-resurrect
  ```
  PASS criterion: output contains both `synth-setter-tmux-resurrect` and `synth-setter-tmux-resurrect-root` rows. (Will only appear *after* the container has been created from this PR's branch at least once.)

### Coverage note

Test-plan items 1, 2, and 3 are all gated on the same host-side rebuild, so they share a single reproduction recipe (above). Item 4 ("pre-existing behaviour unchanged") is fully addressed by the `.devcontainer/tmux.conf` byte-identical check — that file is the only source of mouse-mode / vi-keys / split-bindings configuration. The diff doesn't touch any code that affects tmux's runtime behaviour at all; only the resurrect-dir mount target widens.

PR #1135, opened 2026-05-19
internal-feat(pipeline): dispatch_via_skypilot injects canonical WORKER_SPEC_URI; promote WORKER_SPEC_URI_ENV constant

ktinubu:
## Summary

PR-2 in the input_spec.json chain rooted at #603. Threads the canonical
``spec.r2.input_spec_uri()`` through to the worker env so dispatched workers
read from the same R2 object ``main()`` just uploaded (PR-1 / #1120).

Three concrete changes:

1. **``src/synth_setter/pipeline/constants.py``** — Promotes the previously
   private ``_WORKER_SPEC_URI_ENV`` from ``skypilot_launch.py`` to public
   ``WORKER_SPEC_URI_ENV`` next to ``INPUT_SPEC_FILENAME``. CI helpers and
   future PR-3 / PR-4 wiring import it from one canonical home.

2. **``src/synth_setter/pipeline/skypilot_launch.py::dispatch_via_skypilot``** —
   Adds a required ``spec_uri: str`` keyword-only kwarg. The worker env's
   ``WORKER_SPEC_URI`` is now sourced from this kwarg rather than from
   ``upload_spec_to_r2``'s return value. The legacy ``upload_spec_to_r2``
   call **intentionally still runs** in PR-2 so older CI workflows that read
   the legacy URI keep working; it is removed in PR-4.

3. **``src/synth_setter/cli/generate_dataset.py::main``** — Threads
   ``spec.r2.input_spec_uri()`` (the canonical accessor that includes the
   run's prefix) into ``dispatch_via_skypilot(..., spec_uri=...)``. Note:
   ``spec.r2.uri(INPUT_SPEC_FILENAME)`` would omit the prefix and is the
   anti-pattern the docstring guards against.

## Test plan

- [x] ``make test-fast`` — 1152 passed, 2 skipped.
- [x] New tests:
  - ``TestWorkerSpecUriEnvConstant`` — verifies the public constant.
  - ``test_spec_uri_kwarg_injected_verbatim_into_worker_env`` — proves the
    worker env's ``WORKER_SPEC_URI`` matches the kwarg, not the legacy
    upload's return (load-bearing assertion for PR-2's contract).
  - ``test_dispatch_branch_passes_canonical_spec_uri_kwarg`` — proves
    ``main()`` threads ``spec.r2.input_spec_uri()`` into the dispatch call.

## Closes

Closes #1114

Issue #1113, opened 2026-05-19
feat(pipeline): pipeline.spec_io helpers + runner writes canonical input_spec.json locally and to R2


ktinubu:
## Context

Part of the workflow-spec-upload-delegation series — see design doc at
`thoughts/shared/design/2026-05-19-workflow-spec-upload-delegation.md`,
plan at `thoughts/shared/plans/2026-05-19-workflow-spec-upload-delegation.md`.

This is **Phase 1 / Wave 1** of the series: add the canonical-spec helpers
that subsequent phases build on, and wire them into the runner-side
`generate_dataset.main`.

## Changes

- **New** `src/synth_setter/pipeline/spec_io.py` with:
  - `canonical_local_path(spec, output_dir) -> Path`
  - `write_canonical_spec_locally(spec, output_dir) -> Path` (writes to
    `<output_dir>/data/<task_name>/<run_id>/metadata/input_spec.json`)
  - `upload_canonical_spec(spec) -> str` (uploads to
    `spec.r2.uri(INPUT_SPEC_FILENAME)`, returns the canonical URI)
- `cli/generate_dataset.py::main` calls both helpers after `spec_from_cfg(cfg)`
- `cli/generate_dataset.py::run` switches its `_rclone_copy(...)` call to
  `upload_canonical_spec(spec)` (idempotent at same canonical key)
- New tests under `tests/pipeline/test_spec_io.py` + integration tests in
  `tests/cli/test_generate_dataset.py`

## Acceptance

- `pytest tests/pipeline/test_spec_io.py -v` passes
- `pytest tests/cli/test_generate_dataset.py -k main_writes_local_and_uploads` passes
- `make test-fast` stays green
- After a stubbed `synth-setter-generate-dataset` invocation, both the
  canonical local file AND a canonical R2 object exist

## Out of scope

- Launcher dispatch changes (Phase 2)
- `synth-setter-spec-uri` contract changes (Phase 3)
- Workflow YAML changes (Phase 4)
- `skypilot_launch.py` CLI signature changes (Phase 5)

Refs #385
Refs #603
Part of the workflow-spec-upload-delegation series (design doc in
`thoughts/shared/design/2026-05-19-workflow-spec-upload-delegation.md`)

PR #1137, opened 2026-05-19
fix(pipeline): lazy-import dotenv in r2_io for --no-deps validate-spec

ktinubu:
## Summary

PR #1120 added a top-level `from dotenv import dotenv_values` to `src/synth_setter/pipeline/r2_io.py`. The `validate-spec-structure` job at `.github/workflows/validate-dataset-shards.yaml:74-77` installs the project with `--no-deps` (only adding `pydantic` on top) and then runs `python3 -m synth_setter.pipeline.ci.validate_spec ...`. `validate_spec.py` imports `downloaded_to_tempfile` / `is_r2_uri` from `r2_io.py`, so the top-level `from dotenv` runs at module load — and crashes with `ModuleNotFoundError: No module named 'dotenv'` before the spec is even read.

This blocks every PR that triggers the `Test Dataset Generation` workflow (currently #1129 and #1135).

The fix moves the import into `ensure_r2_env_loaded` — the only function that uses `dotenv_values` — so importing `r2_io` from a lean CI env no longer requires `python-dotenv`.

## Why lazy-import (not adding `python-dotenv` to the install step)

- One-line change vs. modifying CI installs across three workflows.
- Keeps the `--no-deps` invariant on the validate-spec step explicit (the comment at `validate-dataset-shards.yaml:68-73` is intentional about not dragging in torch/skypilot/librosa).
- `ensure_r2_env_loaded` is the only consumer; lazy-importing localizes the dotenv dependency to the function that actually needs it.

## Test plan

- [x] `pytest tests/pipeline/test_r2_io.py` — 31 tests pass.
- [ ] CI validate-spec-structure job goes green on this PR (will verify after open).

Refs #1113

REVIEW_FULL_DONE=1

### DIFF
--- a/src/synth_setter/pipeline/r2_io.py
+++ b/src/synth_setter/pipeline/r2_io.py
@@ -16,8 +16,6 @@ from collections.abc import Iterator
 from contextlib import contextmanager
 from pathlib import Path
 
-from dotenv import dotenv_values
-
 from synth_setter.pipeline.constants import R2_URI_SCHEME, RCLONE_REMOTE
 
 __all__ = [
@@ -65,6 +63,10 @@ def ensure_r2_env_loaded(env_file: Path | None = None) -> None:
         ``rclone lsd r2:`` exits non-zero (bad creds, network, etc.).
     """
     if env_file is not None and env_file.is_file():
+        # Lazy import so importing this module (e.g. from validate_spec.py in
+        # the lean --no-deps CI env) does not pull python-dotenv. Refs #1120.
+        from dotenv import dotenv_values
+
         for key, value in dotenv_values(env_file).items():
             if key and key.startswith("RCLONE_CONFIG_R2_") and value is not None:
                 os.environ[key] = value


ktinubu:
## Verification Results

Head SHA: `7695fab`  •  Mergeable: `MERGEABLE`  •  CI: most checks pass, generate-dataset matrix still running

**Promise under test:** after this change, `import synth_setter.pipeline.r2_io` (and transitively `synth_setter.pipeline.ci.validate_spec`) succeeds in a lean `--no-deps` env without `python-dotenv` — which is exactly what `validate-dataset-shards.yaml`'s `Install validate_spec runtime deps` step produces.

- [x] **`validate_spec.py` + `r2_io.py` import chain works when `python-dotenv` is not importable** *(Level 2 — reproduces the failing-CI env locally by blocking `import dotenv` in a custom `__import__`, then runs the same import the workflow runs)*
  ```bash
  $ python3 -c "
  import builtins
  _real_import = builtins.__import__
  def _block_dotenv(name, *args, **kwargs):
      if name == 'dotenv' or name.startswith('dotenv.'):
          raise ModuleNotFoundError(\"No module named 'dotenv'\")
      return _real_import(name, *args, **kwargs)
  builtins.__import__ = _block_dotenv

  from synth_setter.pipeline.ci.validate_spec import is_r2_uri, downloaded_to_tempfile  # noqa: F401
  print('PASS: validate_spec + r2_io import chain works without dotenv')
  "
  PASS: validate_spec + r2_io import chain works without dotenv
  ```
  Before this PR the same script fails with `ModuleNotFoundError: No module named 'dotenv'` at `r2_io.py` line 19 — confirmed by running it against the sibling worktree at `b818124` (origin/main pre-PR).

- [x] **The lazy import still surfaces a clean `ModuleNotFoundError` when `ensure_r2_env_loaded` actually needs it** *(Level 2 — adversarial: confirms the dependency was only deferred, not hidden, so callers that need dotenv still get a clear error)*
  ```bash
  $ python3 -c "
  import builtins, pathlib, tempfile
  _real_import = builtins.__import__
  def _block_dotenv(name, *args, **kwargs):
      if name == 'dotenv' or name.startswith('dotenv.'):
          raise ModuleNotFoundError(\"No module named 'dotenv'\")
      return _real_import(name, *args, **kwargs)
  builtins.__import__ = _block_dotenv

  from synth_setter.pipeline.r2_io import ensure_r2_env_loaded
  with tempfile.NamedTemporaryFile('w', suffix='.env', delete=False) as f:
      f.write('RCLONE_CONFIG_R2_ENDPOINT=https://example.com\n')
      f.flush()
      env_path = pathlib.Path(f.name)

  try:
      ensure_r2_env_loaded(env_path)
  except ModuleNotFoundError as e:
      print(f'PASS: lazy import surfaces clean ModuleNotFoundError when env_file given: {e}')
  else:
      print('FAIL: expected ModuleNotFoundError, got nothing')
  "
  PASS: lazy import surfaces clean ModuleNotFoundError when env_file given: No module named 'dotenv'
  ```

- [x] **`ensure_r2_env_loaded` happy path still works when dotenv IS available — no behavioral regression** *(Level 2 — runs the existing `tests/pipeline/test_r2_io.py` suite, which exercises `ensure_r2_env_loaded` with and without `env_file`)*
  ```bash
  $ pytest tests/pipeline/test_r2_io.py -x --no-header -q 2>&1 | tail -3
  ============================== 31 passed in 0.49s ==============================
  ```

- [ ] **`validate-spec-structure` CI job goes green on this PR** *(Level 1 — promise is end-to-end CI behavior; the `validate-dataset-shards.yaml` workflow runs the failing job as a downstream of `generate_dataset`. IN PROGRESS — `Run generate_dataset (skypilot-local / hdf5)` and `(wds)` are still `pending` at the time of writing. Will update this comment when CI settles.)*
  ```bash
  $ gh pr checks 1137 --repo tinaudio/synth-setter | grep -E 'Validate|generate_dataset'
  Run generate_dataset (skypilot-local / hdf5) / Run generate_dataset (skypilot-local)	pending	0	...
  Run generate_dataset (skypilot-local / wds) / Run generate_dataset (skypilot-local)	pending	0	...
  ```

PR #1148, opened 2026-05-19
fix(eval): predict_vst_audio UnboundLocalError when target params present without -t

ktinubu:
## Summary

Minimal 2-line fix for the `UnboundLocalError` in
`synth_setter.evaluation.predict_vst_audio.main`: when
`target-params-{i}.pt` is present on disk and `--rerender_target` is NOT
passed (the default flow), the `params_to_csv` call site referenced
`target_synth_params` / `target_note_params` via a ternary that fires
the `target_params is not None` arm — but those names are only bound
inside the rerender branch.

Fix: initialize both to `None` at the top of each row iteration. The
existing call-site conditional still evaluates as before (target column
emitted as NaN when `rerender_target=False`); the prior crash was the
*only* observable behavior change.

This PR is intentionally scoped to *just* the bug fix:
- Source: 3 added lines in `predict_vst_audio.py` (2 inits + 1 blank).
- Tests: 1 new regression test (~27 lines) — fails on `main` HEAD with
  the same `UnboundLocalError`; passes with this fix.

Other latent issues in the module (single-panel spectrogram crash,
double dB conversion, `torch.load` without `weights_only`, larger
refactor of `main`) are **not** bundled here — those can ride in
follow-up PRs without conflating bug-fix bisection. Note: a follow-up
worth filing is whether the non-rerender path should *decode*
`target_params[j]` so the CSV target column is populated when target
params are on disk — that's a behavior change beyond fixing the crash.

Replaces (closed) #1076.

Refs #672

## Test plan

- [ ] `make test-fast` — 1190 passed, 2 skipped (verified locally).
- [ ] `tests/evaluation/test_predict_vst_audio.py` — 12 tests pass (was 11; +1 regression test).
- [ ] Revert the source hunk and rerun the new test — fails as expected with `UnboundLocalError` (verified locally).
- [ ] \`pre-commit run --files <changed>\` — all hooks pass.

### DIFF
--- a/src/synth_setter/evaluation/predict_vst_audio.py
+++ b/src/synth_setter/evaluation/predict_vst_audio.py
@@ -181,6 +181,9 @@ def main(
                 preset_path=preset_path,
             )
 
+            target_synth_params: dict[str, float] | None = None
+            target_note_params: dict[str, float] | None = None
+
             out_target = os.path.join(sample_dir, "target.wav")
             if rerender_target and target_params is not None:
                 target_params_ = target_params[j].numpy()
--- a/tests/evaluation/test_predict_vst_audio.py
+++ b/tests/evaluation/test_predict_vst_audio.py
@@ -360,6 +360,33 @@ def test_main_rerender_target_renders_pred_and_target_per_sample(
         assert bool(df["target"].notna().all())
 
 
+def test_main_target_params_on_disk_without_rerender_does_not_crash(
+    runner: CliRunner, pred_dir: Path, out_dir: Path
+) -> None:
+    """Targets-on-disk + ``rerender_target=False`` must complete without crashing.
+
+    Regression guard for ``UnboundLocalError`` at the ``params_to_csv`` call
+    site: when ``target-params-{i}.pt`` is present but ``--rerender_target``
+    is not passed, ``target_synth_params`` / ``target_note_params`` were never
+    bound but were still referenced by the ``target_params is not None`` arm
+    of the call-site conditional.
+
+    :param runner: Parametrized ``runner`` value under test.
+    :param pred_dir: Parametrized ``pred_dir`` value under test.
+    :param out_dir: Parametrized ``out_dir`` value under test.
+    """
+    _write_batch(pred_dir, index=0, batch_size=2, with_target_params=True)
+
+    result = _invoke_main(runner, pred_dir, out_dir, "--skip-spectrogram")
+
+    assert result.exit_code == 0, result.output
+    for j in range(2):
+        sample_dir = out_dir / f"sample_{j}"
+        assert (sample_dir / "pred.wav").is_file()
+        assert (sample_dir / "target.wav").is_file()
+        assert (sample_dir / "params.csv").is_file()
+
+
 def test_main_multiple_batches_produce_contiguous_sample_indices(
     runner: CliRunner, pred_dir: Path, out_dir: Path
 ) -> None:
--- a/tests/test_train.py
+++ b/tests/test_train.py
@@ -315,9 +315,7 @@ def test_train_eval_surge_xt(
 
     # Render predicted params through the Surge XT VST to per-sample audio directories.
     # `-t` (`--rerender_target`) re-synthesizes target.wav from the stored target_params instead
-    # of the saved target audio. Also works around an `UnboundLocalError` in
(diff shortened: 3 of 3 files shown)


ktinubu:
## Verification Results

The promise of this PR: when `target-params-{i}.pt` is on disk and `--rerender_target` is NOT passed, `predict_vst_audio.main` must complete without raising `UnboundLocalError` and write the expected per-sample artifacts.

- [x] **Click `main` runs end-to-end on the bug-triggering CLI path without crashing** *(Level 1 — the new regression test drives the real `main` entrypoint via `CliRunner` with `target-params-{i}.pt` on disk and `--rerender_target` absent. `render_params` is monkeypatched at the VST3 boundary, but the rest of `main` runs as production. There is no higher-fidelity check short of an actual Surge XT VST3.)*
  ```bash
  $ python -m pytest tests/evaluation/test_predict_vst_audio.py::test_main_target_params_on_disk_without_rerender_does_not_crash -v 2>&1 | tail -3
  tests/evaluation/test_predict_vst_audio.py::test_main_target_params_on_disk_without_rerender_does_not_crash PASSED [100%]
  ============================== 1 passed in 0.83s ===============================
  ```

- [x] **The regression test would catch a reintroduction of this bug** *(Level 1 — adversarial red-phase. Reverted just the source hunk to base, reran the new test, observed the exact `UnboundLocalError` at the original call site, then restored the fix.)*
  ```bash
  $ git checkout origin/main -- src/synth_setter/evaluation/predict_vst_audio.py
  $ python -m pytest tests/evaluation/test_predict_vst_audio.py::test_main_target_params_on_disk_without_rerender_does_not_crash 2>&1 | tail -3
  src/synth_setter/evaluation/predict_vst_audio.py:222: UnboundLocalError
  FAILED tests/evaluation/test_predict_vst_audio.py::test_main_target_params_on_disk_without_rerender_does_not_crash - UnboundLocalError: local variable 'target_synth_params' referenced before a...
  ============================== 1 failed in 0.90s ===============================
  $ git checkout HEAD -- src/synth_setter/evaluation/predict_vst_audio.py  # restored
  ```

- [x] **Fixture writes the file that triggers the bug (`target-params-0.pt`)** *(Level 1 — adversarial. The bug only fires when `target_param_file is not None` at runtime, which requires the fixture to actually write `target-params-0.pt` to disk. If `_write_batch(... with_target_params=True)` silently skipped writing it, the regression test would pass even on the broken code — invalidating the guard. Verified the fixture writes the expected file.)*
  ```bash
  $ python -c "
  import os, tempfile, pathlib, importlib.util
  os.environ.setdefault('MPLBACKEND', 'Agg')
  spec = importlib.util.spec_from_file_location('t', 'tests/evaluation/test_predict_vst_audio.py')
  mod = importlib.util.module_from_spec(spec); spec.loader.exec_module(mod)
  with tempfile.TemporaryDirectory() as d:
      p = pathlib.Path(d) / 'preds'; p.mkdir()
      mod._write_batch(p, index=0, batch_size=2, with_target_params=True)
      print(sorted(f.name for f in p.iterdir()))
  "
  ['pred-0.pt', 'target-audio-0.pt', 'target-params-0.pt']
  ```

- [x] **No regression in the rest of the module's behavior** *(Level 2 — runs the full file's test suite end-to-end, exercising the existing `--rerender_target`, `--no-params`, multi-batch, and spectrogram paths. Promise is "the 2-line init doesn't change any other observable behavior" — full integration of the module's test surface is the right scope; running the whole repo's `make test-fast` would be over-specified for a 3-line source change.)*
  ```bash
  $ python -m pytest tests/evaluation/test_predict_vst_audio.py 2>&1 | tail -1
  ============================== 12 passed in 2.59s ==============================
  ```

### Out of scope / not run

- **`make test-fast` across the whole repo** — Already covered by CI; running it here would be redundant noise. The repo-wide impact of a 3-line init in `predict_vst_audio.main` is bounded by the module-level test above.
- **End-to-end with a real Surge XT VST3** — The CI fast-test environment does not load a real VST3; the production rendering boundary is intentionally stubbed via the `_patch_render_params` autouse fixture, matching every other test in this file.
- **Behavior change in the CSV target column on the non-rerender path** — Out of scope (per PR description). The fix preserves the prior crash-free semantic (target column NaN unless `--rerender_target` is passed). If the inferred-design behavior (populate target columns when target params are on disk) is desired, that's a follow-up.

PR #1206, opened 2026-05-20
ci(cpu-slow): inject RCLONE_CONFIG_R2_* secrets so integration_r2 tests run

### LINKED ISSUE #1185: ci(cpu-slow): inject R2 creds so integration_r2 tests run, not skip
## Context

`cpu-slow.yml` runs `pytest -vv -s -m "slow and not gpu and not mps and not requires_vst"`. Tests carrying the `integration_r2` marker (e.g. `tests/integration/test_local_launcher_roundtrip.py`, `tests/integration/test_finalize_dataset_r2.py`) are also tagged `slow`, so they collect into the cpu-slow run — but they skip immediately via `r2_io.is_r2_reachable()` because the workflow doesn't inject `RCLONE_CONFIG_R2_*` creds.

Net: the `integration_r2` real-R2 surface has **no scheduled coverage today**. Skips are silent — they don't fail the workflow, they just degrade coverage.

## What to change

Wire the same `RCLONE_CONFIG_R2_*` secret block that `generate-dataset-shards.yaml` and `test-local-launcher-roundtrip.yml` use into the `Run slow` step in `.github/workflows/cpu-slow.yml`:

```yaml
      - name: Run slow (non-GPU, non-MPS, non-VST) tests
        env:
          HYDRA_FULL_ERROR: "1"
          RCLONE_CONFIG_R2_ACCESS_KEY_ID: ${{ secrets.R2_ACCESS_KEY_ID }}
          RCLONE_CONFIG_R2_SECRET_ACCESS_KEY: ${{ secrets.R2_SECRET_ACCESS_KEY }}
          RCLONE_CONFIG_R2_ENDPOINT: ${{ secrets.R2_ENDPOINT }}
        run: |
          pytest -vv -s -m "slow and not gpu and not mps and not requires_vst"
```

The `r2_io.is_r2_reachable()` gate (new in PR #1182) keeps the workflow safe on fork PRs that don't carry secrets: tests still auto-skip cleanly, this just means they actually run on the privileged post-merge / dispatch path.

## Acceptance criteria

- [ ] `cpu-slow.yml` sets the three `RCLONE_CONFIG_R2_*` env vars on the pytest step (sourced from the same secrets the other R2-using workflows use).
- [ ] After merge, the next post-main cpu-slow run shows `tests/integration/test_finalize_dataset_r2.py` and `tests/integration/test_local_launcher_roundtrip.py` as **passed** (not skipped) in the test report.
- [ ] On a fork PR (no secrets), the same tests still skip cleanly via `r2_io.is_r2_reachable()` — no regression.

## Out of scope

- Migrating cpu-slow to the larger / different runner — this is a pure env-injection change.
- Adding new `integration_r2` tests — only enables the existing ones.

## References

- `cpu-slow.yml:67` — current invocation
- `r2_io.is_r2_reachable` (introduced in PR for #1182) — the skip helper
- `generate-dataset-shards.yaml` / `test-local-launcher-roundtrip.yml` — existing R2-creds-injection precedent

ktinubu:
## Summary

- `integration_r2`-marked tests in `tests/integration/test_finalize_dataset_r2.py` and `tests/integration/test_local_launcher_roundtrip.py` carry both the `slow` and `integration_r2` markers, so they collect into `cpu-slow.yml`'s pytest invocation. Without the rclone-prefixed R2 env vars, `r2_io.is_r2_reachable()` short-circuits and they skip silently — leaving the `integration_r2` surface with **no scheduled coverage**.
- Wire the same `RCLONE_CONFIG_R2_*` secret block that `generate-dataset-shards.yaml` and `test-local-launcher-roundtrip.yml` already use into the `Run slow` step. On fork PRs without secrets the gate still auto-skips cleanly via `r2_io.is_r2_reachable()` — no regression.
- Add `tests/infra/test_cpu_slow_workflow_r2_creds.py` that parses the workflow and pins both (1) the presence of the three `RCLONE_CONFIG_R2_*` env keys and (2) the secret-name convention — so a future workflow restructure can't drop the keys unnoticed.

Closes #1185

## Test plan

- [x] `pytest tests/infra/test_cpu_slow_workflow_r2_creds.py -vv` — 2 passed
- [x] `pre-commit run --files .github/workflows/cpu-slow.yml tests/infra/test_cpu_slow_workflow_r2_creds.py` — all hooks pass (pyright, ruff, pydoclint, prettier, codespell, yaml-check, etc.)
- [x] gha-workflow-validator skill — 7 passed, 0 failures (the three `${{ secrets.RCLONE_CONFIG_R2_* }}` refs are "must be configured as a repo secret", which they are — proven by generate-dataset-shards.yaml + test-local-launcher-roundtrip.yml using the identical names today)
- [x] `actionlint` clean on the new env block (the only warning is the pre-existing `ubuntu-latest-4core` runner-label entry on the unchanged `runs-on:` line)
- [ ] Post-merge: confirm the next `cpu-slow` run on `main` shows the two `integration_r2` tests as **passed** (not skipped) in the test report

### DIFF
--- a/.github/workflows/cpu-slow.yml
+++ b/.github/workflows/cpu-slow.yml
@@ -60,9 +60,14 @@ jobs:
       - name: List dependencies
         run: uv pip list --system
 
+      # `RCLONE_CONFIG_R2_*` lets `integration_r2`-marked tests reach real
+      # R2 instead of skipping via `r2_io.is_r2_reachable()` — see #1185.
       - name: Run slow (non-GPU, non-MPS, non-VST) tests
         env:
           HYDRA_FULL_ERROR: "1"
+          RCLONE_CONFIG_R2_ACCESS_KEY_ID: ${{ secrets.RCLONE_CONFIG_R2_ACCESS_KEY_ID }}
+          RCLONE_CONFIG_R2_SECRET_ACCESS_KEY: ${{ secrets.RCLONE_CONFIG_R2_SECRET_ACCESS_KEY }}
+          RCLONE_CONFIG_R2_ENDPOINT: ${{ secrets.RCLONE_CONFIG_R2_ENDPOINT }}
         run: |
           pytest -vv -s -m "slow and not gpu and not mps and not requires_vst"
 
--- a/.github/workflows/cpu-slow.yml
+++ b/.github/workflows/cpu-slow.yml
@@ -14,6 +14,13 @@ on:
     paths-ignore:
       - "docs/**"
       - "**/*.md"
+  # Self-trigger so PRs that touch the workflow (or its invariant test)
+  # exercise it pre-merge instead of waiting for the post-merge push run
+  # to surface regressions — see #1206.
+  pull_request:
+    paths:
+      - ".github/workflows/cpu-slow.yml"
+      - "tests/infra/test_cpu_slow_workflow_r2_creds.py"
 
 concurrency:
   group: cpu-slow-${{ github.ref }}
--- a/.github/workflows/cpu-slow.yml
+++ b/.github/workflows/cpu-slow.yml
@@ -60,11 +60,20 @@ jobs:
       - name: List dependencies
         run: uv pip list --system
 
+      # `is_r2_reachable()` shells out to `rclone`; without the binary the
+      # `integration_r2` gate short-circuits and the tests skip silently.
+      # Mirrors the install in `test-local-launcher-roundtrip.yml` and
+      # `generate-dataset-shards.yaml`.
+      - name: Install rclone
+        run: sudo apt-get update -qq && sudo apt-get install -y -qq rclone
+
       # `RCLONE_CONFIG_R2_*` lets `integration_r2`-marked tests reach real
       # R2 instead of skipping via `r2_io.is_r2_reachable()` — see #1185.
       - name: Run slow (non-GPU, non-MPS, non-VST) tests
         env:
           HYDRA_FULL_ERROR: "1"
+          RCLONE_CONFIG_R2_TYPE: s3
+          RCLONE_CONFIG_R2_PROVIDER: Cloudflare
           RCLONE_CONFIG_R2_ACCESS_KEY_ID: ${{ secrets.RCLONE_CONFIG_R2_ACCESS_KEY_ID }}
           RCLONE_CONFIG_R2_SECRET_ACCESS_KEY: ${{ secrets.RCLONE_CONFIG_R2_SECRET_ACCESS_KEY }}
           RCLONE_CONFIG_R2_ENDPOINT: ${{ secrets.RCLONE_CONFIG_R2_ENDPOINT }}
--- a/.github/workflows/cpu-slow.yml
+++ b/.github/workflows/cpu-slow.yml
@@ -31,6 +31,12 @@ permissions:
 
 jobs:
   run_slow_tests:
+    # Skip fork-PR runs: forks can't see `secrets.RCLONE_CONFIG_R2_*` and the
+    # 90-min suite would burn a 4-core runner with `integration_r2` skipping
+    # silently. Workflow_dispatch / push / schedule runs are unaffected — only
+    # the `pull_request` trigger checks head-repo identity (see #1206).
+    if: github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository
+
     # Larger runner: the slow CPU suite runs PyTorch forward passes on the FFN
     # smoke fixtures, and the standard 2-core/7 GB `ubuntu-latest` OOMs during
     # those passes. `ubuntu-latest-4core` (4 vCPU / 16 GB) is the smallest
--- a/tests/infra/test_cpu_slow_workflow_r2_creds.py
+++ b/tests/infra/test_cpu_slow_workflow_r2_creds.py
@@ -15,15 +15,30 @@ from typing import cast
 import pytest
 import yaml
 
-REQUIRED_R2_ENV_KEYS: frozenset[str] = frozenset(
-    {
-        "RCLONE_CONFIG_R2_ACCESS_KEY_ID",
-        "RCLONE_CONFIG_R2_SECRET_ACCESS_KEY",
-        "RCLONE_CONFIG_R2_ENDPOINT",
-    }
+from synth_setter.pipeline.r2_io import _R2_STRUCTURAL_DEFAULTS, _SECRET_R2_ENV_KEYS
+
+# Union of the secret keys rclone needs to authenticate AND the structural
+# defaults rclone needs to assemble the `r2:` remote. `is_r2_reachable()`
+# does NOT apply the `setdefault` that `ensure_r2_env_loaded()` does, so
+# both sets must be present in the runner env — see #1185.
+REQUIRED_R2_ENV_KEYS: frozenset[str] = frozenset(_SECRET_R2_ENV_KEYS) | frozenset(
+    _R2_STRUCTURAL_DEFAULTS
 )
 
 PYTEST_STEP_NAME = "Run slow (non-GPU, non-MPS, non-VST) tests"
+RCLONE_INSTALL_STEP_NAME = "Install rclone"
+
+
+def _load_workflow_steps(project_root: Path) -> list[dict[str, object]]:
+    """Return the ordered ``steps`` list of the ``run_slow_tests`` job.
+
+    :param project_root: repo root; the workflow is read from
+        ``<project_root>/.github/workflows/cpu-slow.yml``.
+    :returns: the job's ``steps`` list.
+    """
+    workflow_path = project_root / ".github" / "workflows" / "cpu-slow.yml"
+    workflow = yaml.safe_load(workflow_path.read_text())
+    return cast(list[dict[str, object]], workflow["jobs"]["run_slow_tests"]["steps"])
 
 
 def _load_pytest_step_env(project_root: Path) -> dict[str, str]:
@@ -34,16 +49,15 @@ def _load_pytest_step_env(project_root: Path) -> dict[str, str]:
     :returns: the step's ``env`` block, or an empty dict if unset.
     """
     workflow_path = project_root / ".github" / "workflows" / "cpu-slow.yml"
-    workflow = yaml.safe_load(workflow_path.read_text())
-    for step in workflow["jobs"]["run_slow_tests"]["steps"]:
+    for step in _load_workflow_steps(project_root):
         if step.get("name") == PYTEST_STEP_NAME:
-            return cast("dict[str, str]", step.get("env") or {})
+            return cast(dict[str, str], step.get("env") or {})
     pytest.fail(f"Could not find step {PYTEST_STEP_NAME!r} in {workflow_path}")
 
 
 @pytest.mark.infra
 def test_cpu_slow_pytest_step_injects_r2_creds(project_root: Path) -> None:
-    """The pytest step exposes all three ``RCLONE_CONFIG_R2_*`` keys via ``env:``.
+    """The pytest step exposes every ``RCLONE_CONFIG_R2_*`` key ``is_r2_reachable()`` reads.
 
     :param project_root: session fixture from ``tests/infra/conftest.py``.
     """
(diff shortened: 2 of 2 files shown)


.github/workflows/cpu-slow.yml
```diff
@@ -60,9 +60,14 @@ jobs:
       - name: List dependencies
         run: uv pip list --system
 
+      # `RCLONE_CONFIG_R2_*` lets `integration_r2`-marked tests reach real
+      # R2 instead of skipping via `r2_io.is_r2_reachable()` — see #1185.
       - name: Run slow (non-GPU, non-MPS, non-VST) tests
         env:
           HYDRA_FULL_ERROR: "1"
```
ktinubu, line 87: **[gha:block]** Injects only the three secret R2 env keys (`RCLONE_CONFIG_R2_ACCESS_KEY_ID`, `_SECRET_ACCESS_KEY`, `_ENDPOINT`) but omits the structural defaults `RCLONE_CONFIG_R2_TYPE: s3` and `RCLONE_CONFIG_R2_PROVIDER: Cloudflare`. The `integration_r2` skip predicate is `r2_io.is_r2_reachable()` (`src/synth_setter/pipeline/r2_io.py:119`), which shells out to `rclone lsd r2:` *without* applying `_R2_STRUCTURAL_DEFAULTS` — only `ensure_r2_env_loaded()` does that `setdefault` (`r2_io.py:90-91`). With `RCLONE_CONFIG_R2_TYPE` / `_PROVIDER` unset in the runner env, `rclone lsd r2:` errors with `didn't find section in config file` (the failure mode documented at `r2_io.py:46-51`), `is_r2_reachable()` returns `False`, and every `integration_r2` test skips silently — exactly the bug #1185 was filed to fix. Mirror `.github/workflows/test-local-launcher-roundtrip.yml:65-66` and inject both literals (`RCLONE_CONFIG_R2_TYPE: s3`, `RCLONE_CONFIG_R2_PROVIDER: Cloudflare`) alongside the secret-sourced keys, and extend the invariant test below to pin them too.

tests/infra/test_cpu_slow_workflow_r2_creds.py
```diff
@@ -0,0 +1,71 @@
...
+
+from __future__ import annotations
+
+from pathlib import Path
+from typing import cast
+
+import pytest
+import yaml
+
+REQUIRED_R2_ENV_KEYS: frozenset[str] = frozenset(
```
ktinubu: **[code-health:warn]** `REQUIRED_R2_ENV_KEYS` is a hand-maintained copy of `synth_setter.pipeline.r2_io._SECRET_R2_ENV_KEYS` (`src/synth_setter/pipeline/r2_io.py:40-44`). The two lists will drift the next time the R2 wiring gains or renames a secret. The `tests/infra/conftest.py` self-containment rule excludes torch/Hydra/h5py — not the lightweight `r2_io` constants module (`os`/`shutil`/`subprocess`/`dotenv`/`pathlib` only) — so importing the canonical tuple is viable. Alternatively, add a one-line equality assertion (`assert set(REQUIRED_R2_ENV_KEYS) == set(r2_io._SECRET_R2_ENV_KEYS)`) so drift fails loudly instead of producing a false-green invariant test.

tests/infra/test_cpu_slow_workflow_r2_creds.py
```diff
@@ -0,0 +1,71 @@
...
+    pytest.fail(f"Could not find step {PYTEST_STEP_NAME!r} in {workflow_path}")
+
+
+@pytest.mark.infra
+def test_cpu_slow_pytest_step_injects_r2_creds(project_root: Path) -> None:
+    """The pytest step exposes all three ``RCLONE_CONFIG_R2_*`` keys via ``env:``.
+
+    :param project_root: session fixture from ``tests/infra/conftest.py``.
+    """
+    env = _load_pytest_step_env(project_root)
```
ktinubu, line 120: **[tdd-impl:warn]** This invariant defends only against the workflow dropping the three *secret* keys; it cannot catch the silent-skip failure mode #1185 was filed to fix. `is_r2_reachable()` (`src/synth_setter/pipeline/r2_io.py:119`) needs the full env contract rclone uses to assemble the `r2:` remote — the three secrets *plus* `RCLONE_CONFIG_R2_TYPE` and `RCLONE_CONFIG_R2_PROVIDER` (see the BLOCK on `.github/workflows/cpu-slow.yml:67`). As written this test will go green even when the workflow has the exact bug the ticket describes. Either widen `REQUIRED_R2_ENV_KEYS` to the union of `_SECRET_R2_ENV_KEYS` + `_R2_STRUCTURAL_DEFAULTS.keys()`, or add a sibling test that asserts the workflow's `env:` block satisfies `is_r2_reachable()`'s precondition end-to-end.

ktinubu (commented):
Multi-skill review of PR #1206 — 8 parallel passes (gha-workflow-validator, code-health, comment-hygiene, python-style, shell-style, synth-setter-project-standards, tdd-implementation, ml-test). Every BLOCK/WARN posted below as an individual unresolved thread. Findings on files outside the diff are anchored to the line in the diff that *causes* the staleness or rolled into the review body.

1 BLOCK + 2 WARN. The BLOCK is load-bearing: as written the workflow's `env:` block still leaves `is_r2_reachable()` returning `False`, so `integration_r2` tests will keep skipping silently — i.e. the change does not yet satisfy #1185. The two WARNs flag the same gap from the test-coverage side (the new invariant test would go green against the buggy state).

.github/workflows/cpu-slow.yml
ktinubu: Resolved in 11f9c40. The pytest step now exports both `RCLONE_CONFIG_R2_TYPE: s3` and `RCLONE_CONFIG_R2_PROVIDER: Cloudflare` alongside the three secrets, mirroring `.github/workflows/test-local-launcher-roundtrip.yml:65-66`. `is_r2_reachable()`'s `rclone lsd r2:` will now find the assembled `r2:` remote instead of bailing with "didn't find section in config file", so the `integration_r2` gate stops silently skipping — closing the #1185 failure mode.

tests/infra/test_cpu_slow_workflow_r2_creds.py
ktinubu: Fixed in 11f9c40 via the import path. `tests/infra/test_cpu_slow_workflow_r2_creds.py` now imports `_SECRET_R2_ENV_KEYS` and `_R2_STRUCTURAL_DEFAULTS` directly from `synth_setter.pipeline.r2_io` and builds `REQUIRED_R2_ENV_KEYS` as their union, so a rename or addition on the `r2_io` side flows through automatically. The `r2_io` constants module only pulls in `os/shutil/subprocess/dotenv/pathlib`, so it stays inside the `tests/infra/` self-containment envelope (no torch/Hydra/h5py).

tests/infra/test_cpu_slow_workflow_r2_creds.py
ktinubu: Fixed in 11f9c40. The invariant now covers the full `is_r2_reachable()` precondition — `REQUIRED_R2_ENV_KEYS` is the union of `_SECRET_R2_ENV_KEYS` and `_R2_STRUCTURAL_DEFAULTS.keys()`, and a new `test_cpu_slow_r2_structural_env_values_use_canonical_literals` asserts the workflow exports `s3`/`Cloudflare` literally (not via `secrets.*`). The bugged state described in the BLOCK now fails this suite instead of going green.

tests/infra/test_cpu_slow_workflow_r2_creds.py
```diff
@@ -0,0 +1,71 @@
...
+
+    :param project_root: repo root; the workflow is read from
+        ``<project_root>/.github/workflows/cpu-slow.yml``.
+    :returns: the step's ``env`` block, or an empty dict if unset.
+    """
+    workflow_path = project_root / ".github" / "workflows" / "cpu-slow.yml"
+    workflow = yaml.safe_load(workflow_path.read_text())
+    for step in workflow["jobs"]["run_slow_tests"]["steps"]:
+        if step.get("name") == PYTEST_STEP_NAME:
+            return cast("dict[str, str]", step.get("env") or {})
```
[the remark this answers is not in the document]
ktinubu: Fixed in 11f9c40. The two `cast(...)` call sites now pass the type object (`cast(list[dict[str, object]], ...)` / `cast(dict[str, str], ...)`) instead of a string literal.

tests/infra/test_cpu_slow_workflow_r2_creds.py
```diff
@@ -0,0 +1,71 @@
...
+
+    :param project_root: session fixture from ``tests/infra/conftest.py``.
+    """
+    env = _load_pytest_step_env(project_root)
+    for key in sorted(REQUIRED_R2_ENV_KEYS):
+        expected = f"${{{{ secrets.{key} }}}}"
+        assert env[key] == expected, (
+            f"cpu-slow.yml env[{key!r}] = {env[key]!r}; expected {expected!r} to match the "
+            f"secret-name convention used by generate-dataset-shards.yaml + "
+            f"test-local-launcher-roundtrip.yml"
```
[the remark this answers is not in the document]
ktinubu: Fixed in 11f9c40. The test is now parametrized per key (`test_cpu_slow_r2_secret_env_values_reference_matching_secrets[<key>]`) and uses `env.get(key)` + a single equality assertion, so a missing key fails its own case with the key name in the output instead of collapsing into a `KeyError` from a sibling case.

.github/workflows/cpu-slow.yml
```diff
@@ -60,9 +60,14 @@ jobs:
...
+          RCLONE_CONFIG_R2_ACCESS_KEY_ID: ${{ secrets.RCLONE_CONFIG_R2_ACCESS_KEY_ID }}
+          RCLONE_CONFIG_R2_SECRET_ACCESS_KEY: ${{ secrets.RCLONE_CONFIG_R2_SECRET_ACCESS_KEY }}
+          RCLONE_CONFIG_R2_ENDPOINT: ${{ secrets.RCLONE_CONFIG_R2_ENDPOINT }}
         run: |
           pytest -vv -s -m "slow and not gpu and not mps and not requires_vst"
```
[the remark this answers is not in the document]
ktinubu: Fixed in 11f9c40. A new `Install rclone` step (`sudo apt-get update -qq && sudo apt-get install -y -qq rclone`) runs before the pytest step, mirroring the install in `test-local-launcher-roundtrip.yml` and `generate-dataset-shards.yaml`. A companion `test_cpu_slow_installs_rclone_before_pytest` invariant pins the ordering so a future workflow restructure can't drop or reorder the install.

ktinubu:
Added a `pull_request:` self-trigger so PRs that touch `.github/workflows/cpu-slow.yml` or its invariant test (`tests/infra/test_cpu_slow_workflow_r2_creds.py`) exercise the slow suite pre-merge — catches workflow regressions before they land on `main` instead of after.

Also extended `test_cpu_slow_workflow_r2_creds.py` with `test_cpu_slow_pull_request_self_trigger_present` (parametrized over both paths) so the self-trigger can't be silently dropped in a later restructure.

Commit: `2670dc1`

tests/infra/test_cpu_slow_workflow_r2_creds.py
```diff
@@ -0,0 +1,204 @@
...
+        rather than collapsing into a ``KeyError`` from the others.
+    """
+    env = _load_pytest_step_env(project_root)
+    actual = env.get(key)
+    expected = f"${{{{ secrets.{key} }}}}"
+    assert actual == expected, (
+        f"cpu-slow.yml env[{key!r}] = {actual!r}; expected {expected!r} to match the "
+        f"secret-name convention used by generate-dataset-shards.yaml + "
+        f"test-local-launcher-roundtrip.yml"
+    )
```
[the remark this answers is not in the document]
ktinubu: Resolved in 6264fbe. The secret-env assertion now uses `re.fullmatch(r'\$\{\{\s*secrets\.<KEY>\s*\}\}', env_value)` so equivalent GitHub expressions (`${{secrets.X}}`, extra whitespace inside the braces) no longer trip the invariant on harmless formatting changes. The KEY-name substitution is escaped via `re.escape`.

.github/workflows/cpu-slow.yml
```diff
@@ -14,6 +14,13 @@ on:
     paths-ignore:
       - "docs/**"
       - "**/*.md"
+  # Self-trigger so PRs that touch the workflow (or its invariant test)
+  # exercise it pre-merge instead of waiting for the post-merge push run
+  # to surface regressions — see #1206.
+  pull_request:
+    paths:
+      - ".github/workflows/cpu-slow.yml"
+      - "tests/infra/test_cpu_slow_workflow_r2_creds.py"
```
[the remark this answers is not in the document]
ktinubu: Resolved in 6264fbe. Added a job-level `if: github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository` on `run_slow_tests`, so fork-PR triggers skip the 90-min slow suite while `workflow_dispatch` / `push` / `schedule` runs are unaffected. A new `test_cpu_slow_job_gated_against_fork_prs` invariant parses the workflow YAML and asserts the guard's head-repo clause is present, so a future restructure that drops it regresses loudly instead of silently burning runner-minutes.

Issue #14, opened 2026-03-16
ci(testing): enable Codecov coverage gates


ktinubu:
## Context

Coverage runs on every PR via `.github/workflows/test.yml` (the `code-coverage` job), but nothing enforces it:

- **No `codecov.yml`/`.codecov.yml`** — no project/patch thresholds, no flags, no components.
- **No `[tool.coverage.run]`** in `pyproject.toml` — only `exclude_lines` is set. No `source`, `branch`, `parallel`, `omit`.
- **No `fail_under`** anywhere.
- **`gh api repos/tinaudio/synth-setter --jq .has_codecov` returns `false`** — Codecov isn't actually wired up on the org, so uploads are silently no-ops.
- **The `code-coverage` job duplicates the fast-suite run** that `run_tests_ubuntu (3.10)` already executes — ~20 min of CI burnt per PR.
- **GPU/MPS/VST/slow tests contribute zero coverage** because the coverage job runs only `-m "not slow and not gpu and not mps"`.
- **Not a required check** on branch protection.

Complementary to #30 (raise coverage to ≥80% for `src/`) and lives alongside the broader test-infrastructure work in Epic #149 / Phase #155. This issue is specifically about **building the enforcement and visibility infrastructure**; #30 is about **reaching the target**.

### 1. Activate Codecov + add `CODECOV_TOKEN`

Until `has_codecov: true`, every upload is a no-op. Install Codecov on the org, add the repo, store `CODECOV_TOKEN` as a repo secret.

### 2. Stop double-running tests; collect coverage in every matrix leg

Delete the standalone `code-coverage` job. Add `--cov=src --cov-report=xml --cov-branch` to every matrix leg of `run_tests_ubuntu`/`run_tests_macos` plus the GPU (`test-gpu.yml`), MPS (`test-mps.yml`), VST (`test-vst-slow.yml`), and slow (`cpu-slow.yml`) jobs. Upload each as a separate Codecov flag (`unit-cpu`, `unit-gpu`, `unit-mps`, `vst`, `slow`). Saves a job and finally counts GPU/VST/slow code paths.

### 3. Configure `[tool.coverage.run]` properly in `pyproject.toml`

```toml
[tool.coverage.run]
source = ["src/synth_setter", "scripts/ci"]
branch = true
parallel = true
relative_files = true
omit = ["**/__main__.py", "src/synth_setter/cli/*"]

[tool.coverage.paths]
[...]

PR #1211, opened 2026-05-20
ci(testing): cleanup coverage config + add slow-cpu Codecov flag

ktinubu:
## Summary

Small cleanup pass on the Codecov integration to remove silent footguns and start counting the slow CPU suite as coverage. Four self-contained changes:

- **Delete `.github/codecov.yml`** — a 10-line threshold stub that Codecov's precedence silently ignored in favor of the root `codecov.yml`. Having two configs invited drift.
- **Fix the `pragma: nocover` typo in `pyproject.toml`** — coverage.py's default exclude token is `pragma: no cover` (with a space). The old spelling was non-standard and never matched anything (the lone in-tree user, `scripts/ci/job_queue.py:302`, was only being excluded by coverage.py's built-in default pattern). Grep across `src/`, `tests/`, `scripts/` confirmed no other usages.
- **Upload coverage from `cpu-slow.yml`** under a new `slow-cpu` Codecov flag. The slow CPU suite exercises PyTorch forward passes and pipeline integration paths that the fast `unit-cpu` suite skips, so those lines were reading as 0% covered — perversely incentivizing not testing the hard paths.
- **Register `slow-cpu` in `codecov.yml`'s `flag_management`** so carryforward and path scoping match `unit-cpu`.

No enforcement changes — `informational: true` stays as-is, per the existing roadmap on #14.

## Review

Self-review via `/repo-review-full-no-comments` returned 0 BLOCK / 3 WARN across 5 skills (code-health, synth-setter-project-standards, gha-workflow-validator, comment-hygiene, shell-style). The comment-hygiene WARN (3-line YAML preamble exceeded the 2-line cap from CLAUDE.md) is fixed in c241b75. Remaining two WARNs are non-blocking observations:

- `cpu-slow.yml` upload step uses `if: always() && hashFiles('coverage.xml') != ''` while `test.yml` has no guard. Defensible asymmetry (slow suite can OOM mid-run) — left as-is.
- `codecov.yml` `slow-cpu` `paths:` block duplicates `unit-cpu`. Worth a YAML anchor only if a third CPU flag lands. No action.

## Test plan

- [ ] Workflow YAML and `codecov.yml` parse via `yaml.safe_load` (verified locally during edit).
- [ ] `gh workflow run cpu-slow.yml` (or normal post-merge fire) uploads to Codecov under the `slow-cpu` flag — confirm a `slow-cpu` row appears in the Codecov PR comment / project view after the next slow-CPU run on `main`.
- [ ] CI passes on this PR. The coverage upload step uses `if: always() && hashFiles('coverage.xml') != ''` so a missing xml doesn't fail the job.

Refs #14

### DIFF
--- a/.github/codecov.yml
+++ b/.github/codecov.yml
@@ -1,9 +1,0 @@
-coverage:
-  status:
-    project:
-      default:
-        threshold: 1% # fail if coverage drops by more than 1%
-
-    patch:
-      default:
-        threshold: 1% # fail if new code has >1% less coverage than target
--- a/.github/workflows/cpu-slow.yml
+++ b/.github/workflows/cpu-slow.yml
@@ -67,9 +67,8 @@ jobs:
           pytest -vv -s -m "slow and not gpu and not mps and not requires_vst" \
             --cov=src --cov-branch --cov-report=xml --cov-report=term
 
-      # Slow CPU tests exercise PyTorch forward passes and pipeline code paths
-      # that the fast suite skips. Uploading under a dedicated `slow-cpu` flag
-      # keeps that signal separable from `unit-cpu` in Codecov.
+      # Dedicated `slow-cpu` flag keeps this signal separable from `unit-cpu`
+      # in Codecov — the fast suite skips these PyTorch forward passes.
       - name: Upload coverage to Codecov
         if: always() && hashFiles('coverage.xml') != ''
         uses: codecov/codecov-action@v5
--- a/.github/workflows/cpu-slow.yml
+++ b/.github/workflows/cpu-slow.yml
@@ -64,7 +64,20 @@ jobs:
         env:
           HYDRA_FULL_ERROR: "1"
         run: |
-          pytest -vv -s -m "slow and not gpu and not mps and not requires_vst"
+          pytest -vv -s -m "slow and not gpu and not mps and not requires_vst" \
+            --cov=src --cov-branch --cov-report=xml --cov-report=term
+
+      # Slow CPU tests exercise PyTorch forward passes and pipeline code paths
+      # that the fast suite skips. Uploading under a dedicated `slow-cpu` flag
+      # keeps that signal separable from `unit-cpu` in Codecov.
+      - name: Upload coverage to Codecov
+        if: always() && hashFiles('coverage.xml') != ''
+        uses: codecov/codecov-action@v5
+        with:
+          files: coverage.xml
+          flags: slow-cpu
+          token: ${{ secrets.CODECOV_TOKEN }}
+          fail_ci_if_error: false
 
       # Auto-file a bug ticket on post-merge failure so CPU regressions in
       # main don't go silently stale. Runs only when the run failed AND the
--- a/codecov.yml
+++ b/codecov.yml
@@ -23,6 +23,10 @@ flag_management:
       paths:
         - src/synth_setter/
         - scripts/ci/
+    - name: slow-cpu
+      paths:
+        - src/synth_setter/
+        - scripts/ci/
 
 component_management:
   default_rules:
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -338,7 +338,7 @@ scripts = ["scripts/ci", "*/scripts/ci"]
 
 [tool.coverage.report]
 exclude_lines = [
-    "pragma: nocover",
+    "pragma: no cover",
     "raise NotImplementedError",
     "raise NotImplementedError()",
     "if __name__ == .__main__.:",


ktinubu:
## Verification Results

Behavioral checks for the four config changes in this PR. Workflow dispatch is in flight — final tick comes when run [26181520498](https://github.com/tinaudio/synth-setter/actions/runs/26181520498) finishes.

- [x] **`# pragma: no cover` is now actually excluded by coverage.py with this config** *(Level 2 — promise is coverage-tool behavior, exercised directly. Fixture has one line marked `# pragma: no cover` (the canonical spelling) and one marked `# pragma: nocover` (the old typo); the canonical one must be excluded, the typo must not be.)*
  ```bash
  $ cat /tmp/pragma_fixture.py
  """Tiny module to verify `# pragma: no cover` exclusion works under the
  project's `[tool.coverage.report].exclude_lines` config."""


  def covered():
      return 1


  def excluded_via_pragma():  # pragma: no cover
      return 2


  def hypothetically_typo_excluded():  # pragma: nocover
      return 3
  $ cd /tmp && COVERAGE_RCFILE=<pr-pyproject> coverage run --source=pragma_fixture --branch -m pragma_fixture
  $ COVERAGE_RCFILE=<pr-pyproject> coverage report --include='pragma_fixture.py' -m
  Name                Stmts   Miss Branch BrPart  Cover   Missing
  ---------------------------------------------------------------
  pragma_fixture.py       4      2      0      0    50%   6, 14
  ---------------------------------------------------------------
  TOTAL                   4      2      0      0    50%
  ```
  Line 10 (`return 2`, behind `# pragma: no cover`) is absent from the Missing list — i.e. excluded from statement counts entirely. Line 14 (`return 3`, behind `# pragma: nocover`) is in the Missing list — i.e. counted as missed because the typo is no longer in `exclude_lines`. This is exactly the behavior the config rename promises: the canonical pragma works, the broken-typo one no longer does.

- [x] **`.github/codecov.yml` is removed from the tree** *(Level 5 — promise is Codecov no longer sees a duplicate config. Codecov's per-file precedence is server-side; the testable client-side promise is file absence. No higher-level option without round-tripping through Codecov.)*
  ```bash
  $ ls -la .github/codecov.yml 2>&1 || echo "DELETED (expected)"
  ls: cannot access '.github/codecov.yml': No such file or directory
  DELETED (expected)
  ```

- [x] **`codecov.yml` is internally consistent after the `slow-cpu` flag addition** *(Level 5 — descended because `codecov-cli validate` install was denied by the sandbox; falling back to YAML schema spot-check. Would be Level 2 if `codecovcli validate-yaml` were available.)*
  ```bash
  $ python3 -c "
  import yaml
  with open('codecov.yml') as f:
      cfg = yaml.safe_load(f)
  flags = {f['name']: f for f in cfg['flag_management']['individual_flags']}
  print('Defined flags:', sorted(flags))
  assert 'unit-cpu' in flags, 'unit-cpu missing'
  assert 'slow-cpu' in flags, 'slow-cpu missing'
  assert flags['slow-cpu']['paths'] == ['src/synth_setter/', 'scripts/ci/'], 'slow-cpu paths mismatch'
  assert cfg['flag_management']['default_rules']['carryforward'] is True, 'carryforward missing'
  print('PASS — slow-cpu registered with matching path scope; carryforward inherited')
  "
  Defined flags: ['slow-cpu', 'unit-cpu']
  PASS — slow-cpu registered with matching path scope; carryforward inherited
  ```

- [ ] **`cpu-slow.yml` end-to-end produces coverage.xml and uploads to Codecov under `slow-cpu`** *(Level 1 — workflow has `workflow_dispatch`, dispatched on PR branch. In progress — slow CPU suite typically ~30–60 min.)*
  ```bash
  $ gh workflow run cpu-slow.yml --ref chore/coverage-config-cleanup
  https://github.com/tinaudio/synth-setter/actions/runs/26181520498
  $ gh run view 26181520498 --json status,conclusion
  {"conclusion":"","status":"in_progress"}
  ```
  Setup, uv install, dep install, dep list are all `success`. `Run slow (non-GPU, non-MPS, non-VST) tests` is in progress; `Upload coverage to Codecov` and `Auto-file failure ticket` are still pending. I'll update this checkbox after the run finishes — the pass criterion is: (a) `Upload coverage to Codecov` step succeeds, and (b) the Codecov PR comment shows a `slow-cpu` flag row.

### Adversarial / implicit-contract checks

- [x] **PR scope is exactly the four files claimed** *(Level 5 — guards against unintended file changes piggybacking on the PR.)*
  ```bash
  $ git diff --name-only c0d01afbf16c456973b07ce0dce6039cdd99625a..c241b75410d7c396655ceb333974057fa7bfdf8e
  .github/codecov.yml
  .github/workflows/cpu-slow.yml
  codecov.yml
  pyproject.toml
  ```

- [x] **The `pragma: nocover` typo wasn't load-bearing on any existing exclusion** *(Level 2 — exercises coverage.py against the actual repo's lone `# pragma:` site. The agent earlier confirmed `scripts/ci/job_queue.py:302` uses the canonical spelling with a space, so the rename can't have stranded any in-tree exclusion.)*
  ```bash
  $ grep -rn 'pragma:' src/ scripts/ tests/ 2>/dev/null
  scripts/ci/job_queue.py:302:    main()  # pragma: no cover
  ```
  Only one in-tree user, and it already uses `# pragma: no cover` (the canonical spelling). The typo was a silent no-op all along; removing it removes the silent no-op without breaking anything.

- [ ] **The `if: always() && hashFiles('coverage.xml') != ''` upload guard correctly skips when pytest crashes before xml is written** *(Level 2 — SKIP: would require dispatching a workflow with a forced-crash variant, which is out of scope for this PR. The idiom is documented in upstream Codecov examples and `hashFiles()` returning the empty string for missing files is GitHub Actions standard. If you want a hard behavioral check, dispatch `cpu-slow.yml` with `pytest -x --tb=no -p no:cacheprovider` swapped into the step temporarily and confirm the upload step is skipped — but that's a regression test for the GHA expression engine, not for this PR.)*

### Notes on level descents

- Check C above is Level 5 instead of Level 4 because Codecov's "which config file did I read?" answer is only knowable via a Codecov server round-trip. Verifying the file is absent is the closest non-blocking proxy.
- Check D above is Level 5 instead of Level 2 because `pip install codecov-cli` was denied by the auto-mode classifier (scope creep beyond the PR task). If `codecovcli` were pre-installed, the same check via `codecovcli validate-yaml codecov.yml` would be Level 2.

PR #1238, opened 2026-05-21
fix(ci-automation): mask degenerate mel bins on the smoke finalize cells

ktinubu:
## Summary

Unblocks the `test-dataset-finalization.yml` wds matrix cell on PR #1214 (and pre-empts the same flake on the hdf5 cell). The smoke matrix's wds row was failing in the `Run synth-setter-finalize-dataset` step with:

```
ValueError: Found 2002 mel bin(s) with zero variance across the dataset
(std shape (2, 128, 401); ...). Rerun with --mask-degenerate-bins to mask
these bins instead of failing.
```
([failing run](https://github.com/tinaudio/synth-setter/actions/runs/26189589640))

### Root cause

- Both smoke configs render 12 samples of `surge_simple` audio.
- The rendered audio is deterministically constant at certain mel positions — early frames (attack-time silence) and channels below the source's active bandwidth.
- `stream_stats_wds`'s Welford reports the exact zero variance and raises by default.
- The hdf5 cell happened to pass only because dask's two-pass `std(axis=0)` introduces float noise that lands those same positions just above zero. Same input data, different rounding.

`get_stats_hdf5` / `stream_stats_wds` already accept `mask_degenerate=True` to substitute `std=1.0` at degenerate positions (which yields `(spec - mean) / std == 0` on the constant-bin training distribution — no datamodule changes). The finalize CLI was not surfacing the knob.

### Changes

- `configs/dataset.yaml` — add top-level `mask_degenerate_bins: false`. Strict default holds for production datasets so a real upstream regression still surfaces loudly.
- `configs/experiment/generate_dataset/smoke-shard.yaml` and `smoke-shard-wds.yaml` — override to `true`. Both rows get the same setting so the hdf5 row does not regress when FP noise stops hiding constant bins.
- `src/synth_setter/cli/finalize_dataset.py` — thread `mask_degenerate_bins` through `finalize_wds`, `finalize_hdf5`, and `main()`. `main()` reads it from the composed cfg via `cfg.get("mask_degenerate_bins", False)` and forwards to the dispatched branch.
- `src/synth_setter/cli/generate_dataset.py` — register `"mask_degenerate_bins"` in `_NON_SPEC_KEYS` so `spec_from_cfg` pops the finalize-only key before constructing the `DatasetSpec`.
- `tests/pipeline/test_entrypoints/test_finalize_dataset.py` — extend existing stub signatures for the new kwarg; add `test_finalize_wds_forwards_mask_degenerate_bins_to_stream_stats` and the symmetric `test_finalize_hdf5_forwards_mask_degenerate_bins_to_get_stats`, both parametrized over `[True, False]` to pin both polarities of the wire.

### Why not just bump `train_val_test_sizes`?

12 samples is intentionally tiny so the smoke matrix fits inside the `generate-dataset-shards.yaml` 20-min step timeout. A larger train size would push the smoke render past that budget without buying coverage the existing matrix doesn't already exercise.

## Test plan

- [ ] `test-dataset-finalization.yml` wds matrix row turns green on this PR.
- [ ] `test-dataset-finalization.yml` hdf5 matrix row stays green.
- [x] `pytest tests/pipeline tests/integration/test_finalize_dataset_r2.py` — 658 passed, 2 skipped locally.
- [x] Hydra compose of both smoke experiments resolves `mask_degenerate_bins == True`.
- [x] Pre-commit (ruff, pyright, pydoclint, yamllint, mdformat, etc.) green on the staged files.
- [x] Multi-skill review (`repo-review-full-no-comments`) — 0 BLOCK, 11 WARN; 8 addressed in follow-up commit, 3 accepted as documented trade-offs.

Refs #1184

### DIFF
--- a/src/synth_setter/cli/finalize_dataset.py
+++ b/src/synth_setter/cli/finalize_dataset.py
@@ -65,7 +65,7 @@ def _download_train_shards_one_at_a_time(spec: DatasetSpec, work_dir: Path) -> I
             local.unlink(missing_ok=True)
 
 
-def finalize_wds(spec: DatasetSpec, work_dir: Path, mask_degenerate_bins: bool = False) -> None:
+def finalize_wds(spec: DatasetSpec, work_dir: Path) -> None:
     """Stream stats over the train shards and upload ``stats.npz``.
 
     Per-shard tar files stay in their original R2 location; only the
@@ -73,13 +73,12 @@ def finalize_wds(spec: DatasetSpec, work_dir: Path, mask_degenerate_bins: bool =
     splits are available via
     ``spec.r2.split_wds_brace_uri(spec.split_shard_ranges[split])``;
     callers must check ``lo < hi`` first because empty splits raise
-    ``ValueError`` at that helper.
+    ``ValueError`` at that helper. ``spec.mask_degenerate_bins`` is
+    forwarded to ``stream_stats_wds``.
 
     :param spec: Validated dataset spec (``output_format == "wds"``).
     :param work_dir: Scratch directory; one shard at a time + the final
         ``stats.npz`` live here transiently.
-    :param mask_degenerate_bins: Forwarded to ``stream_stats_wds``; ``True``
-        substitutes ``std=1.0`` at zero-variance mel bins instead of raising.
     :raises ValueError: The train split is empty
         (``spec.split_shard_ranges["train"]`` has ``lo >= hi``); stats
         cannot be computed without at least one train shard.
@@ -93,7 +92,7 @@ def finalize_wds(spec: DatasetSpec, work_dir: Path, mask_degenerate_bins: bool =
         )
     mean, std = stream_stats_wds(
         _download_train_shards_one_at_a_time(spec, work_dir),
-        mask_degenerate=mask_degenerate_bins,
+        mask_degenerate=spec.mask_degenerate_bins,
     )
     stats_npz = work_dir / STATS_NPZ_FILENAME
     np.savez(stats_npz, mean=mean, std=std)
@@ -101,7 +100,7 @@ def finalize_wds(spec: DatasetSpec, work_dir: Path, mask_degenerate_bins: bool =
     logger.info("uploaded stats to {}", spec.r2.stats_uri())
 
 
-def finalize_hdf5(spec: DatasetSpec, work_dir: Path, mask_degenerate_bins: bool = False) -> None:
+def finalize_hdf5(spec: DatasetSpec, work_dir: Path) -> None:
     """Download every shard, reshard into split files, compute stats, upload all artifacts.
 
     Writes ``work_dir/input_spec.json`` flat (via
@@ -120,12 +119,11 @@ def finalize_hdf5(spec: DatasetSpec, work_dir: Path, mask_degenerate_bins: bool 
     validation (per ``pipeline/CLAUDE.md``) is delegated to the h5py opens
     that ``reshard_dataset`` performs while staging each split — finalize
     never re-runs the workers' full four-check pass.
+    ``spec.mask_degenerate_bins`` is forwarded to ``get_stats_hdf5``.
 
     :param spec: Validated dataset spec (``output_format == "hdf5"``).
     :param work_dir: Scratch directory; shards, splits, stats and the spec
         copy live here transiently for the duration of the call.
-    :param mask_degenerate_bins: Forwarded to ``get_stats_hdf5``; ``True``
-        substitutes ``std=1.0`` at zero-variance mel bins instead of raising.
     :raises ValueError: The train split is empty
         (``spec.split_shard_ranges["train"]`` has ``lo >= hi``); reshard
         would prune ``train.h5`` and stats compute would fail with a
@@ -144,7 +142,7 @@ def finalize_hdf5(spec: DatasetSpec, work_dir: Path, mask_degenerate_bins: bool 
         r2_io.download_to_path(spec.r2.shard_uri(shard), work_dir / shard.filename)
     write_spec_to_path(spec, work_dir / INPUT_SPEC_FILENAME)
     reshard_dataset(work_dir)
-    get_stats_hdf5(str(work_dir / "train.h5"), mask_degenerate=mask_degenerate_bins)
+    get_stats_hdf5(str(work_dir / "train.h5"), mask_degenerate=spec.mask_degenerate_bins)
     stats_npz = work_dir / STATS_NPZ_FILENAME
     if not stats_npz.is_file():
         raise FileNotFoundError(
@@ -184,9 +182,6 @@ def main() -> None:
     cfg.paths.output_dir = str(_REPO_ROOT)
     cfg.paths.work_dir = str(_REPO_ROOT)
 
-    # Read before spec_from_cfg; ``_NON_SPEC_KEYS`` drops it from spec construction.
-    mask_degenerate_bins = bool(cfg.get("mask_degenerate_bins", False))
-
     spec = spec_from_cfg(cfg)
     r2_io.ensure_r2_env_loaded()
 
@@ -198,9 +193,9 @@ def main() -> None:
     with tempfile.TemporaryDirectory() as raw_work_dir:
         work_dir = Path(raw_work_dir)
         if spec.output_format == "wds":
-            finalize_wds(spec, work_dir, mask_degenerate_bins=mask_degenerate_bins)
+            finalize_wds(spec, work_dir)
         elif spec.output_format == "hdf5":
-            finalize_hdf5(spec, work_dir, mask_degenerate_bins=mask_degenerate_bins)
+            finalize_hdf5(spec, work_dir)
         else:
             raise ValueError(f"unsupported output_format: {spec.output_format!r}")
         marker_local = work_dir / DATASET_COMPLETE_FILENAME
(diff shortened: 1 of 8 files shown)


ktinubu:
## Verification Results

Behavioral verifications for the `mask_degenerate_bins` wire and the degenerate-bin contract. The CI smoke matrix can only run after this PR merges into `feat/finalize-dataset-ci-presubmit`, so the end-to-end Level 1 check for the matrix itself is intentionally left unchecked with a reproducer.

### Explicit promises

- [x] **Smoke configs resolve `mask_degenerate_bins == True`; production-shape config resolves to `False`** *(Level 1 — exercises the real Hydra subsystem the CLI reads at runtime; this IS the system under test for the cfg wiring)*
  \`\`\`bash
  $ python3 -c \"
  from hydra import compose, initialize_config_dir
  from pathlib import Path
  config_dir = Path('configs').resolve()
  with initialize_config_dir(version_base='1.3', config_dir=str(config_dir)):
      cfg_smoke_wds = compose(config_name='dataset', overrides=['experiment=generate_dataset/smoke-shard-wds'])
  with initialize_config_dir(version_base='1.3', config_dir=str(config_dir)):
      cfg_smoke_hdf5 = compose(config_name='dataset', overrides=['experiment=generate_dataset/smoke-shard'])
  with initialize_config_dir(version_base='1.3', config_dir=str(config_dir)):
      cfg_prod = compose(config_name='dataset', overrides=['experiment=generate_dataset/ci-materialize-test-wds'])
  print('smoke-shard-wds  mask_degenerate_bins =', cfg_smoke_wds.get('mask_degenerate_bins'))
  print('smoke-shard      mask_degenerate_bins =', cfg_smoke_hdf5.get('mask_degenerate_bins'))
  print('ci-materialize-test-wds (production-shape) mask_degenerate_bins =', cfg_prod.get('mask_degenerate_bins'))
  \"
  smoke-shard-wds  mask_degenerate_bins = True
  smoke-shard      mask_degenerate_bins = True
  ci-materialize-test-wds (production-shape) mask_degenerate_bins = False
  \`\`\`

- [x] **`finalize_wds` / `finalize_hdf5` forward `mask_degenerate_bins` to the stats helpers on both polarities** *(Level 2 — invokes the wire-pinning tests directly; promise is kwarg propagation, not end-to-end)*
  \`\`\`bash
  $ python -m pytest tests/pipeline/test_entrypoints/test_finalize_dataset.py::test_finalize_wds_forwards_mask_degenerate_bins_to_stream_stats tests/pipeline/test_entrypoints/test_finalize_dataset.py::test_finalize_hdf5_forwards_mask_degenerate_bins_to_get_stats -v 2>&1 | tail -6
  tests/pipeline/test_entrypoints/test_finalize_dataset.py::test_finalize_wds_forwards_mask_degenerate_bins_to_stream_stats[True] PASSED   [ 25%]
  tests/pipeline/test_entrypoints/test_finalize_dataset.py::test_finalize_wds_forwards_mask_degenerate_bins_to_stream_stats[False] PASSED  [ 50%]
  tests/pipeline/test_entrypoints/test_finalize_dataset.py::test_finalize_hdf5_forwards_mask_degenerate_bins_to_get_stats[True] PASSED     [ 75%]
  tests/pipeline/test_entrypoints/test_finalize_dataset.py::test_finalize_hdf5_forwards_mask_degenerate_bins_to_get_stats[False] PASSED    [100%]
  4 passed in 1.34s
  \`\`\`

- [x] **Welford strict path still raises on a degenerate dataset; `mask_degenerate=True` substitutes `std=1.0`** *(Level 2 — exercises `finalize()` directly against a synthetic input that has a constant bin; this is the exact contract the CI failure tripped)*
  \`\`\`bash
  $ python3 -c \"
  import numpy as np
  from synth_setter.pipeline.data.stats import update, finalize
  existing = (0, np.array([0.0, 0.0]), np.array([0.0, 0.0]))
  for x in [np.array([1.0, 5.0]), np.array([2.0, 5.0]), np.array([3.0, 5.0])]:
      existing = update(existing, x)
  try:
      finalize(existing, mask_degenerate=False)
      print('FAIL: expected ValueError')
  except ValueError as e:
      print('strict raised:', str(e).split('.')[0])
  mean, std = finalize(existing, mask_degenerate=True)
  print('masked mean=', mean, 'std=', std)
  assert std[1] == 1.0
  print('PASS: mask substitutes std=1.0 at the degenerate bin')
  \"
  strict raised: Found 1 mel bin(s) with zero variance across the dataset (std shape (2,); indices [1])
  Masking 1 degenerate mel bin(s) with std=1.0 (std shape (2,); indices [1]).
  masked mean= [2. 5.] std= [0.81649658 1.        ]
  PASS: mask substitutes std=1.0 at the degenerate bin
  \`\`\`

- [ ] **`test-dataset-finalization.yml` wds matrix cell finalizes without raising** *(Level 1 — SKIP: the smoke matrix only fires on the parent PR #1214's branch via `on: pull_request`. This sub-PR targets `feat/finalize-dataset-ci-presubmit`; the smoke runs the next time #1214's HEAD moves after this PR is squashed into it.)*

  To verify manually after this PR lands on the parent branch:
  \`\`\`bash
  # Bump the parent PR's branch
  gh pr checks 1214 --repo tinaudio/synth-setter --watch
  # Inspect the wds matrix cell specifically
  gh run list --repo tinaudio/synth-setter -w 'Test Dataset Finalization' -L 1 \\
    --json databaseId,conclusion --jq '.[0]'
  # Confirm the 'Run finalize_dataset (skypilot-local / )' step under
  # 'Generate + finalize smoke (wds)' shows conclusion=success, not the
  # `ValueError: Found 2002 mel bin(s) with zero variance` from run 26189589640.
  \`\`\`

- [x] **`spec_from_cfg` filters `mask_degenerate_bins` out of `DatasetSpec` construction** *(Level 2 — invokes the real `spec_from_cfg` on a cfg that contains the key; promise is the `_NON_SPEC_KEYS` registration actually drops it before pydantic's strict-mode rejection fires)*
  \`\`\`bash
  $ python3 -c \"
  from hydra import compose, initialize_config_dir
  from pathlib import Path
  from synth_setter.cli.generate_dataset import spec_from_cfg
  with initialize_config_dir(version_base='1.3', config_dir=str(Path('configs').resolve())):
      cfg = compose(config_name='dataset', overrides=['experiment=generate_dataset/smoke-shard-wds'])
  cfg.paths.root_dir = str(Path('.').resolve())
  cfg.paths.output_dir = str(Path('.').resolve())
  cfg.paths.work_dir = str(Path('.').resolve())
  print('cfg contains mask_degenerate_bins:', 'mask_degenerate_bins' in cfg)
  spec = spec_from_cfg(cfg)
  print('spec built; spec.output_format =', spec.output_format)
  print('spec has no mask_degenerate_bins attr:', not hasattr(spec, 'mask_degenerate_bins'))
  \"
  cfg contains mask_degenerate_bins: True
  spec built; spec.output_format = wds
  spec has no mask_degenerate_bins attr: True
  \`\`\`
  Without the `_NON_SPEC_KEYS` registration, this would raise pydantic's `ValidationError: Extra inputs are not permitted` because `DatasetSpec` is `strict=True`.

- [x] **Missing-key fallback in `main()` honors the strict default** *(Level 2 — exercises the `bool(cfg.get('mask_degenerate_bins', False))` read with a cfg that lacks the key)*
  \`\`\`bash
  $ python3 -c \"
  cfg = OmegaConf.create({'paths': {}, 'output_format': 'wds'})
  val = bool(cfg.get('mask_degenerate_bins', False))
  print('missing-key default value =', val)
  assert val is False
  print('PASS: missing key falls back to False (strict)')
  \"
  missing-key default value = False
  PASS: missing key falls back to False (strict)
  \`\`\`
  Defends against an older operator-built cfg that pre-dates this PR — the CLI keeps strict behavior instead of defaulting silently to masking.

- [x] **Full existing finalize test suite still green under widened signatures** *(Level 2 — runs the full `test_finalize_dataset.py` module; promise is no regression in the 16 pre-existing tests after stub-signature updates)*
  \`\`\`bash
  $ python -m pytest tests/pipeline/test_entrypoints/test_finalize_dataset.py -q 2>&1 | tail -3
  (42 durations < 0.005s hidden.  Use -vv to show these durations.)
  ============================== 20 passed in 2.50s ==============================
  \`\`\`
  20 = 16 pre-existing + 4 new (2 polarities × wds wire + 2 polarities × hdf5 wire).

src/synth_setter/cli/finalize_dataset.py
```diff
@@ -64,7 +64,7 @@ def _download_train_shards_one_at_a_time(spec: DatasetSpec, work_dir: Path) -> I
             local.unlink(missing_ok=True)
 
 
-def finalize_wds(spec: DatasetSpec, work_dir: Path) -> None:
+def finalize_wds(spec: DatasetSpec, work_dir: Path, mask_degenerate_bins: bool = False) -> None:
```
ktinubu: Why can't we infer mask_degenerate_bins from the DatasetSpec?

ktinubu: Fixed in 4706c20b — added `mask_degenerate_bins: bool = False` to [`DatasetSpec`](https://github.com/tinaudio/synth-setter/blob/fix-finalize-wds-smoke/src/synth_setter/pipeline/schemas/spec.py), dropped the `_NON_SPEC_KEYS` carve-out, and removed the parallel kwarg from `finalize_wds`/`finalize_hdf5`/`main()`. Both finalize branches now read `spec.mask_degenerate_bins` directly. Wire tests updated to set the flag via the spec builder instead of via kwarg, and `_make_valid_spec` in `test_validate_spec.py` carries the new field (the validator's required-set derives from `model_fields`, so adding the field auto-tightens the structural check).

Issue #1250, opened 2026-05-24
feat(pipeline): default rank/world to (0, 1) when partition env absent


ktinubu:
## Context

`pipeline.partitioning.read_rank_world_from_env` currently raises if either `SYNTH_SETTER_WORKER_RANK` or `SYNTH_SETTER_NUM_WORKERS` is unset. The strict-fail behavior was put in place by [#763](https://github.com/tinaudio/synth-setter/issues/763) to prevent a misconfigured worker from silently duplicating every shard across every node.

The side effect: any local invocation of `synth-setter-generate-dataset` requires exporting both env vars first, even for a one-off single-process run that obviously can't duplicate work across nodes.

## Goal

Make `synth-setter-generate-dataset` runnable locally without partition env setup, while keeping the #763 invariant intact for any multi-node deployment.

## Scope

- `read_rank_world_from_env`: when BOTH vars are absent, return `(0, 1)` (local single-worker default).
- Keep raising for the partial case (only one var set) — that's the launcher-misconfig signal #763 was protecting against.
- Keep raising for malformed / out-of-bounds values.
- Update existing unit tests; add tests that catch regressions where the launcher drops env injection now that the worker has a default.
- Drop the now-redundant `SYNTH_SETTER_WORKER_RANK=0` / `SYNTH_SETTER_NUM_WORKERS=1` hardcode from `nightly-parallel-datagen.yml` so the workflow exercises the new default end-to-end.

## Out of scope

- Changes to the SkyPilot launcher's env injection (must continue to inject explicit values for every rank).
- Changes to `oci-image-bake.yaml` (separate orchestration path; can clean up later).

Issue #1247, opened 2026-05-22
Task 1.1: Restructure pyproject.toml + generate uv.lock


ktinubu:
Parent: #1244

## Scope

Apply the design in #1243 to `pyproject.toml` and produce the first committed `uv.lock`. This is the foundational PR — CI gating (Phase 2 / #1245) and docs (Phase 3 / #1246) depend on it landing first.

## Changes in this PR

- Move torch stack (`torch`, `torchvision`, `torchaudio`, `lightning`, `torchmetrics`) into regular `[project] dependencies`
- Replace the `torch`/`dev`/`all` optional-dependencies with a single `cpu` extra
- Move `dev` and `docs` to PEP 735 `[dependency-groups]` (with `dev` including `docs` so a bare `uv sync` covers both)
- Add `[[tool.uv.index]]` for `pytorch-cpu` and `pytorch-cu128` (explicit, marker-gated to Linux/Win)
- Add `[tool.uv.sources]` routing torch/torchvision/torchaudio per platform
- Add `[tool.uv]` `conflicts = [[{ extra = "cpu" }]]` so CPU and default-CUDA resolve independently into the same lockfile
- Tighten `requires-python = ">=3.10,<3.14"`
- Relax `scipy==1.14.1` → `scipy>=1.14,<1.15`
- Generate and commit `uv.lock`

## Out of scope (follow-up PRs)

- CI matrix and `uv sync --frozen` switchover — Phase 2 / #1245
- README/CONTRIBUTING + relock cadence docs — Phase 3 / #1246
- Skypilot bare-runner install update — Phase 2 / #1245

## Verification

- `uv lock --check` passes locally after commit
- `uv sync` resolves without `--extra` (mac MPS path or Linux CUDA path)
- `uv sync --extra cpu --no-default-groups --group dev` resolves the CPU wheel on Linux

ktinubu:
Closing as complete. Verified by multi-agent triage+audit+close-verify pass (tracking #1734).

## What was asked
Restructure `pyproject.toml` and generate the first committed `uv.lock` — the foundational PR for the lockfile epic (CI gating and docs depend on it).

## What landed
PR #1248 (commit 4d2c5c88) — `chore(deps): introduce uv.lock + hybrid PyTorch routing`
- `uv.lock` committed (1055253 bytes on disk as of HEAD)
[...]

Issue #763, opened 2026-05-02
internal-feat(pipeline): static range shard partitioning for SkyPilot multi-node workers


ktinubu:
## Context

`pipeline.entrypoints.generate_dataset.run(spec)` currently iterates over every shard in `spec.shards`. Once we move to a multi-node SkyPilot launch, that means every node renders every shard — duplicate work and duplicate R2 uploads.

## Goal

Each worker should render a deterministic, contiguous slice of the shard list based on its SkyPilot rank. The partition is static — computed purely from `(total_shards, rank, world)` with no leases, claim files, or liveness checks.

## Scope

- `pipeline.partitioning.get_my_shards(total_shards, rank, world) -> range` — pure function, no env reads.
- `pipeline.partitioning.validate_rank_world(rank, world)` — shared bounds check used by both the partitioner and the SkyPilot env verifier.
- `pipeline.entrypoints.verify_skypilot_env` — deployment guard that fails loudly when `SKYPILOT_NODE_RANK` / `SKYPILOT_NUM_NODES` are missing, run before `generate_dataset` in the SkyPilot YAML.
- Wire both into `generate_dataset.run` and the RunPod template's `run:` block.
- Multi-node smoke coverage in `.github/workflows/test-dataset-generation.yml` (3 workers × 3 shards).

## Out of scope

- Dynamic / work-stealing partitioning, leases, or claim-based ownership.
- Spec-upload deduplication across workers (every worker re-uploads to the same idempotent R2 key).
- Production-default `num_nodes` in the RunPod template (deployment concern; CI sets it via sed at job time).

PR #1251, opened 2026-05-24
feat(pipeline): default rank/world to (0, 1) when partition env absent

ktinubu:
## Summary

- [`read_rank_world_from_env`](https://github.com/tinaudio/synth-setter/blob/chore/local-default-rank-world/src/synth_setter/pipeline/partitioning.py) now defaults to `(0, 1)` when both `SYNTH_SETTER_WORKER_RANK` and `SYNTH_SETTER_NUM_WORKERS` are unset, so `synth-setter-generate-dataset` runs locally without exporting partition env. Partial config (only one set) still raises — that pattern almost always means a launcher dropped half its env injection, and silently coercing it would duplicate every shard across every node ([#763](https://github.com/tinaudio/synth-setter/issues/763) invariant preserved).
- Defense-in-depth tests pin the launcher/worker env-var contract so the relaxed worker default can't hide a launcher regression:
  - new launcher-side test asserts single-worker dispatch still injects explicit `WORKER_RANK=0` / `NUM_WORKERS=1`
  - new cross-platform integration test drives two ranks across the real subprocess boundary, wipes R2 state between runs, and asserts a disjoint + complete shard partition
- [`nightly-parallel-datagen.yml`](https://github.com/tinaudio/synth-setter/blob/chore/local-default-rank-world/.github/workflows/nightly-parallel-datagen.yml) drops the redundant hardcoded partition env so it exercises the new default end-to-end in CI — a regression that loses the fallback fails the workflow before a launcher-less developer hits it.
- Drive-by: rebound `os.sched_getaffinity` via `getattr` in `available_cpus` so pyright's narrowing works on macOS as well as Linux (the prior `hasattr` guard ran fine but didn't satisfy pyright on macOS).

> Note: PR base is `chore/1247-uv-lock-foundation` (the in-flight uv-lock branch this was cut from), not `main`. Retarget to `main` after #1247 lands, or rebase if #1247 stalls.

Fixes [#1250](https://github.com/tinaudio/synth-setter/issues/1250)

## Test plan

- [x] `uv run pytest tests/pipeline/test_partitioning.py tests/pipeline/test_entrypoints/test_generate_dataset.py tests/pipeline/test_entrypoints/test_skypilot_launch.py tests/integration/test_parallel_shard_dispatch.py` → 173 passed
- [x] `uv run pre-commit run --files <touched files>` → all hooks pass (ruff, pyright, docformatter, pydoclint, GHA validator)
- [x] Pre-PR multi-skill review (`/repo-review-full-no-comments`) — 0 BLOCK, ~14 WARN across 9 skills; 2 fixed in-PR (vacuous disjointness assertion, exact-order coupling), 12 deferred with documented rationale (sentinel: `.agent-reviews/repo-review-full-no-comments.6a09002df339c35e68eff0b1a625143c906cdfaa.md`)
- [ ] Nightly `nightly-parallel-datagen` succeeds on its next 05:00 UTC run — workflow-level proof the default works end-to-end (post-merge verification)

### DIFF
--- a/.github/workflows/nightly-parallel-datagen.yml
+++ b/.github/workflows/nightly-parallel-datagen.yml
@@ -112,8 +112,13 @@ jobs:
       #     keeps it null), so ``main()`` runs ``run(spec)`` inline rather than
       #     dispatching via SkyPilot — exercises ``_dispatch_shards_parallel``
       #     directly on this runner.
-      #   * ``SYNTH_SETTER_WORKER_RANK=0`` / ``SYNTH_SETTER_NUM_WORKERS=1`` so
-      #     the single in-process worker owns every shard.
+      #   * Partition env intentionally unset — exercises the
+      #     ``read_rank_world_from_env`` local default ``(rank=0, world=1)``
+      #     end-to-end. A regression that loses the default fails this
+      #     workflow before a launcher-less developer hits it. The #763
+      #     invariant (partial partition env still raises; only the
+      #     both-absent path defaults) is pinned by the partitioning unit
+      #     tests.
       #   * ``++render.parallel=true`` overrides the default ``parallel=false``
       #     even though the experiment YAML already sets it — the explicit
       #     override keeps the production-path coverage signal in the workflow
@@ -146,8 +151,6 @@ jobs:
             -e EXPERIMENT \
             -e NIGHTLY_TASK_NAME \
             -e HYDRA_FULL_ERROR=1 \
-            -e SYNTH_SETTER_WORKER_RANK=0 \
-            -e SYNTH_SETTER_NUM_WORKERS=1 \
             "$IMAGE" \
             bash -c '
               set -euo pipefail
--- a/src/synth_setter/pipeline/partitioning.py
+++ b/src/synth_setter/pipeline/partitioning.py
@@ -3,7 +3,8 @@
 Each worker owns a contiguous, deterministic slice of shard IDs computed from
 ``(total_shards, rank, world)``. Pure helpers (``get_my_shards``,
 ``validate_rank_world``) and an env-reading shell (``read_rank_world_from_env``)
-that fails loudly on missing partition env — see #763.
+that defaults to single-worker local mode when both partition env vars are
+absent, and rejects any partial / malformed config — see #763.
 
 Env vars are ``SYNTH_SETTER_``-prefixed to dodge SkyPilot's reserved
 ``SKYPILOT_NODE_RANK`` (clobbered to 0 per single-node cluster) and to avoid
@@ -27,8 +28,9 @@ def available_cpus() -> int:
 
     :returns: Number of CPUs usable by this process; always ``>= 1``.
     """
-    if hasattr(os, "sched_getaffinity"):
-        return len(os.sched_getaffinity(0))
+    sched_getaffinity = getattr(os, "sched_getaffinity", None)
+    if sched_getaffinity is not None:
+        return len(sched_getaffinity(0))
     return os.cpu_count() or 1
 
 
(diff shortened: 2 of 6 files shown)


ktinubu:
## Verification Results

All checks executed against PR head `6a09002d` on macOS Darwin 25.2.0 (local pre-merge verification). Pyright check uses `--pythonplatform Linux` to mirror CI.

- [x] **`read_rank_world_from_env` returns `(0, 1)` with no partition env exported** *(Level 2 — invoked the env-reading shell directly with the relevant vars stripped from `os.environ`; promise is "function returns the local default," not end-to-end pipeline behavior)*
  ```bash
  $ PYTHONPATH=src uv run python -c "
  import os
  os.environ.pop('SYNTH_SETTER_WORKER_RANK', None)
  os.environ.pop('SYNTH_SETTER_NUM_WORKERS', None)
  from synth_setter.pipeline.partitioning import read_rank_world_from_env
  print('result:', read_rank_world_from_env())
  "
  result: (0, 1)
  ```

- [x] **Partial env (only one of the two set) raises with both var names in the message** *(Level 2 — invoked the env-reading shell with each half-set permutation; promise is "loud error naming the missing piece," verified through both code paths)*
  ```bash
  # Case A: only world set
  $ PYTHONPATH=src uv run python -c "
  import os
  os.environ.pop('SYNTH_SETTER_WORKER_RANK', None)
  os.environ['SYNTH_SETTER_NUM_WORKERS'] = '2'
  from synth_setter.pipeline.partitioning import read_rank_world_from_env
  try:
      read_rank_world_from_env()
      print('FAIL — did not raise')
  except ValueError as e:
      print('only-world-set raised:', e)
  "
  only-world-set raised: partial partition env: SYNTH_SETTER_NUM_WORKERS set but SYNTH_SETTER_WORKER_RANK missing (set both or neither — neither defaults to single-worker local mode)

  # Case B: only rank set
  $ PYTHONPATH=src uv run python -c "
  import os
  os.environ.pop('SYNTH_SETTER_NUM_WORKERS', None)
  os.environ['SYNTH_SETTER_WORKER_RANK'] = '0'
  from synth_setter.pipeline.partitioning import read_rank_world_from_env
  try:
      read_rank_world_from_env()
      print('FAIL — did not raise')
  except ValueError as e:
      print('only-rank-set raised:', e)
  "
  only-rank-set raised: partial partition env: SYNTH_SETTER_WORKER_RANK set but SYNTH_SETTER_NUM_WORKERS missing (set both or neither — neither defaults to single-worker local mode)
  ```

- [x] **Two-rank integration test produces a disjoint + complete shard partition across the real subprocess boundary** *(Level 2 — runs the new integration test which crosses ThreadPoolExecutor → subprocess.check_call → fresh Python interpreter → real rclone copy → fake R2 remote. Highest level achievable without Surge VST + R2 creds; matches the level of the single-rank precedent test)*
  ```bash
  $ uv run pytest tests/integration/test_parallel_shard_dispatch.py::test_two_ranks_render_disjoint_complete_shard_partition -v --no-header 2>&1 | tail -3
  tests/integration/test_parallel_shard_dispatch.py::test_two_ranks_render_disjoint_complete_shard_partition PASSED [100%]
  ============================== 1 passed in 0.52s ===============================
  ```
  Note: the amended commit fixes a vacuous intersection check flagged by three reviewers — between rank-0 and rank-1 runs, the bucket prefix is now wiped (`shutil.rmtree`) so rank-1's landed set is observed independently rather than derived as a set difference. Without that, the disjointness assertion was mathematically guaranteed to pass regardless of what rank-1 actually did.

- [x] **SkyPilot launcher injects explicit rank/world env for every rank — single-worker and multi-worker** *(Level 2 — runs both the new defense-in-depth test and the existing fan-out test; promise is "launcher contract independent of the worker fallback")*
  ```bash
  $ uv run pytest tests/pipeline/test_entrypoints/test_skypilot_launch.py::TestDispatchViaSkypilot::test_single_worker_dispatch_still_injects_rank_world_env tests/pipeline/test_entrypoints/test_skypilot_launch.py::TestDispatchViaSkypilot::test_multi_worker_fans_out_one_task_per_rank -v --no-header 2>&1 | tail -3
  tests/pipeline/test_entrypoints/test_skypilot_launch.py::TestDispatchViaSkypilot::test_multi_worker_fans_out_one_task_per_rank PASSED [100%]
  2 passed in 1.30s
  ```

- [x] **Pyright clean on `partitioning.py` under CI platform (Linux)** *(Level 2 — invoked pyright with `--pythonplatform Linux` to match the CI runner; promise is "drive-by getattr rewrite resolves the macOS-vs-Linux divergence without breaking CI's narrow")*
  ```bash
  $ uv run pyright --pythonplatform Linux src/synth_setter/pipeline/partitioning.py 2>&1 | grep -E "error|warning|information"
  ```

- [x] **Malformed and out-of-bounds env vars still raise (adversarial — verifies the new default branch didn't accidentally swallow other error cases)** *(Level 2)*
  ```bash
  $ uv run pytest tests/pipeline/test_partitioning.py::TestReadRankWorldFromEnv::test_out_of_bounds_rank_raises tests/pipeline/test_partitioning.py::TestReadRankWorldFromEnv::test_non_integer_rank_raises tests/pipeline/test_partitioning.py::TestReadRankWorldFromEnv::test_non_integer_world_raises tests/pipeline/test_partitioning.py::TestReadRankWorldFromEnv::test_world_zero_raises -v --no-header 2>&1 | tail -3
  tests/pipeline/test_partitioning.py::TestReadRankWorldFromEnv::test_world_zero_raises PASSED [100%]
  4 passed in 0.10s
  ```

- [x] **`nightly-parallel-datagen.yml` no longer hardcodes the partition env vars** *(Level 6 — grep against the file; promise is "the workflow exercises the new default end-to-end." This check verifies the removal happened; the workflow-level proof that the end-to-end run still succeeds is the SKIP below)*
  ```bash
  $ grep -cn "SYNTH_SETTER_WORKER_RANK\|SYNTH_SETTER_NUM_WORKERS" .github/workflows/nightly-parallel-datagen.yml
  0
  ```

- [ ] **`nightly-parallel-datagen` workflow succeeds end-to-end without the env hardcode** *(Level 1 — SKIP: this PR's base is `chore/1247-uv-lock-foundation`, not `main`, so the workflow on this ref would also pull in unrelated #1247 changes. The workflow also requires R2 secrets and `dev-snapshot` Docker image access, and the schedule is `0 5 * * *`. The workflow-level proof will land naturally on the first scheduled run after this PR merges to main.)*
  To verify manually after merge to main:
  ```bash
  # 1. Confirm origin/main is up to date with this change merged
  # 2. Manually trigger:
  gh workflow run nightly-parallel-datagen.yml --ref main \
    -f image_tag=dev-snapshot
  # 3. Watch:
  gh run watch
  # 4. PASS = full job green (generate + validate steps both succeed without
  #    having to export SYNTH_SETTER_WORKER_RANK / SYNTH_SETTER_NUM_WORKERS)
  ```

PR #1257, opened 2026-05-24
fix(ci): force imagePullPolicy=Never in CI to avoid Docker Hub 429

ktinubu:
## Summary

- Inject `kubernetes.pod_config.spec.containers[0].imagePullPolicy: Never` into `~/.sky/config.yaml` from `_ensure_ci_sky_config()`, gated on `SYNTH_SETTER_CI_MODE=1`.
- SkyPilot merges this into every CI-kind pod (jobs controller + worker), so both pods now use the pre-loaded dev-snapshot image and never re-resolve the manifest from Docker Hub.
- Adds a test pinning the exact nested location of the override in the rendered config.

Refs #1255

## Why

The `wds` cell of the `generate-dataset-shards.yaml` matrix [timed out after 40 minutes](https://github.com/tinaudio/synth-setter/actions/runs/26371827439/job/77625225624) on [#1214](https://github.com/tinaudio/synth-setter/pull/1214) because the worker pod was stuck in `ImagePullBackOff` with `429 Too Many Requests` from `registry-1.docker.io`:

```
Failed to pull image "tinaudio/synth-setter:dev-snapshot": ...
  429 Too Many Requests - Server message: toomanyrequests:
  You have reached your unauthenticated pull rate limit.
```

The workflow already loads `dev-snapshot` into the kind node's containerd via `kind load docker-image`, so the image is physically present. Kubelet's `IfNotPresent` default still resolves the manifest from the registry on pod start though, and that resolution is what trips Docker Hub's anonymous limit. Forcing `imagePullPolicy: Never` makes kubelet use the local image without contacting the registry at all. The `hdf5` row earlier in the matrix made it through Docker Hub before the bucket emptied; `wds` was scheduled minutes later and lost the race.

`Never` is safe in CI because the workflow guarantees `kind load docker-image` runs to completion before the launcher starts; in production (no `SYNTH_SETTER_CI_MODE`) this code path is skipped so real-cloud workers continue to pull normally.

A companion follow-up will mirror the image to GHCR to remove the dependency on Docker Hub entirely; this PR is the immediate unblock — hence `Refs` rather than `Fixes`.

## Verification

- [x] `uv run pytest tests/pipeline/test_entrypoints/test_skypilot_launch.py -x -q` — 89 passed (88 existing + 1 new).
[...]

PR #1270, opened 2026-05-24
fix(ci): scope imagePullPolicy=Never to the worker task, not ~/.sky/config.yaml

### LINKED ISSUE #1255: fix(ci-automation): wds smoke times out on Docker Hub 429 — force imagePullPolicy=Never in CI
## Symptom

The `wds` cell of the `generate-dataset-shards.yaml` matrix (skypilot-local row) [timed out after 40 minutes](https://github.com/tinaudio/synth-setter/actions/runs/26371827439/job/77625225624) on PR [#1214](https://github.com/tinaudio/synth-setter/pull/1214). The captured kind cluster events (`finalize-presubmit-wds` artifact, `k8s_state/events.txt`) show the SkyPilot-managed worker pod stuck in `ImagePullBackOff` for the whole window:

```
Failed to pull image "tinaudio/synth-setter:dev-snapshot": failed to pull and unpack image
  "docker.io/tinaudio/synth-setter:dev-snapshot": ...
  429 Too Many Requests - Server message: toomanyrequests:
  You have reached your unauthenticated pull rate limit.
```

The `hdf5` row ran a few minutes earlier and made it through Docker Hub before the anonymous bucket emptied; the `wds` row was scheduled minutes later and got rate-limited for the whole 40-min window.

## Why this happens despite `kind load docker-image`

`generate-dataset-shards.yaml` already pre-loads the dev-snapshot image into the kind node's containerd via `kind load docker-image` before the launcher starts, so the image is physically present on the node. Kubelet's default `imagePullPolicy` for non-`:latest` tags is `IfNotPresent` *in theory*, but containerd still resolves the manifest from the registry on pod start — and that's the request hitting 429.

## Fix in this PR (immediate)

Inject `kubernetes.pod_config.spec.containers[*].imagePullPolicy: Never` into `~/.sky/config.yaml` from `_ensure_ci_sky_config()` in `src/synth_setter/pipeline/skypilot_launch.py`, gated on `SYNTH_SETTER_CI_MODE=1` (same gate as the existing controller-resource shrink). SkyPilot's `combine_pod_config_fields` merges this into every Kubernetes pod it provisions on the kind cluster — both the jobs controller pod (which got lucky on this run) and the worker pod (which did not). Production runs against real clouds never set `SYNTH_SETTER_CI_MODE`, so this code path is CI-only.

`Never` is safe here because the workflow guarantees the image is loaded into kind before the launcher runs, so kubelet always finds it locally.

## Companion follow-up: GHCR mirror

A companion PR will add a longer-term fix — mirror `tinaudio/synth-setter:dev-snapshot` to GHCR so the kind workflow pulls from a registry without anonymous rate limits. That makes the `imagePullPolicy: Never` belt-and-braces and removes the dependency on `kind load` happening successfully.

## Refs

- Failed run: https://github.com/tinaudio/synth-setter/actions/runs/26371827439/job/77625225624
- PR triggering the failure: [#1214](https://github.com/tinaudio/synth-setter/pull/1214)
- Related earlier work on CI sky config shrink: [#876](https://github.com/tinaudio/synth-setter/pull/876)

ktinubu:
## Summary

- Moves the `imagePullPolicy: Never` override out of `_ensure_ci_sky_config`'s global `~/.sky/config.yaml` write and into the top-level `config:` block of `src/synth_setter/configs/compute/local-template.yaml`. SkyPilot reads that field as the task's `Resources._cluster_config_overrides`, so the merge fires only when provisioning the user worker — never the jobs controller.
- Removes the now-redundant `kubernetes.pod_config` block from `_CI_SKY_CONFIG_YAML`; the controller-resource shrink (`jobs.controller.resources`) stays.
- Updates `tests/pipeline/test_entrypoints/test_skypilot_launch.py` to assert the new shape: `~/.sky/config.yaml` carries no `kubernetes.pod_config` key (the inverse of the old assertion), and `local-template.yaml` does carry the per-task override.

## Why

Run [26374103900](https://github.com/tinaudio/synth-setter/actions/runs/26374103900/job/77631256867) — the `Test Dataset Finalization` workflow on main, exercising the fix that landed in #1257 — timed out at 40 min in the launcher step for **both** matrix rows. The launcher tee showed kubelet rejecting the SkyPilot jobs controller pod 35+ times:

```
sky.provision.kubernetes.config.KubernetesError: Failed to create container while launching the node.
Error details: Container image "us-docker.pkg.dev/sky-dev-465/skypilotk8s/skypilot:20250909"
               is not present with pull policy of Never.
```

The image referenced is the **SkyPilot jobs-controller** image (resolved from SkyPilot's remote catalog: `skypilot:custom-cpu-ubuntu-2004` → `us-docker.pkg.dev/sky-dev-465/skypilotk8s/skypilot:20250909`). It lives on Google Artifact Registry (no anonymous rate limit) and is **not** loaded into kind by `kind load docker-image` — only `tinaudio/synth-setter:dev-snapshot` is. PR #1257's global `kubernetes.pod_config.spec.containers[0].imagePullPolicy: Never` merged into every kind pod SkyPilot provisions, including the controller, so kubelet refused to pull the controller image and the cluster never came up.

The original #1255 symptom — anonymous Docker Hub 429 on the worker's `tinaudio/synth-setter:dev-snapshot` pull — still gets `Never` because the per-task scope reaches the worker pod (where `kind load` already placed the image). Kubelet uses the loaded copy without re-resolving the manifest from Docker Hub, so the 429 path stays closed. The controller pulls normally from GAR.

I verified SkyPilot's task loader accepts the new YAML and that the override lands on `Resources._cluster_config_overrides` (task-scoped):

```python
>>> sky.Task.from_yaml_config(yaml.safe_load(open('local-template.yaml'))).resources[0].cluster_config_overrides
{'kubernetes': {'pod_config': {'spec': {'containers': [{'imagePullPolicy': 'Never'}]}}}}
```

## Test plan

- [ ] Local: `pytest tests/pipeline/test_entrypoints/test_skypilot_launch.py` — 91 passed in 2.14s on `4e0593a2`.
- [ ] CI: rerun `Test Dataset Finalization` (workflow_dispatch on main with this branch cherry-picked) and confirm both matrix rows finish the launcher step in <10 min and the verify-artifacts job passes. The previous run on main timed out the 40-min step cap.
- [ ] Operator local-dev: a kind cluster with the synth-setter image pre-loaded (per the local-template.yaml comment) still works; without the pre-load, the worker now fails fast with `not present with pull policy of Never` (intentional — matches the documented invariant on the YAML).

Fixes #1255.

### DIFF
--- a/src/synth_setter/configs/compute/local-template.yaml
+++ b/src/synth_setter/configs/compute/local-template.yaml
@@ -16,12 +16,10 @@
 # cluster via `kind load docker-image` ahead of launch. `image_id` is set
 # per-launch by the launcher from skypilot_launch.worker_image_tag.
 
-# Task-scoped pod_config override: `imagePullPolicy: Never` makes kubelet
-# use the kind-loaded synth-setter image and never resolve its manifest
-# from Docker Hub, avoiding the anonymous-pull 429 that timed out #1255.
-# Scoped to this task (not ~/.sky/config.yaml) so the SkyPilot jobs
-# controller — whose image lives on GAR and is not `kind load`-ed — still
-# pulls normally; a global override blocks the controller from starting.
+# Task-scoped (not ~/.sky/config.yaml): a global override blocks the
+# SkyPilot jobs controller, whose image lives on GAR and is not `kind
+# load`-ed. Name-less container merges into the pod's first container
+# via SkyPilot's patch-merge legacy. See #1255.
 config:
   kubernetes:
     pod_config:
--- a/src/synth_setter/configs/compute/local-template.yaml
+++ b/src/synth_setter/configs/compute/local-template.yaml
@@ -16,6 +16,19 @@
 # cluster via `kind load docker-image` ahead of launch. `image_id` is set
 # per-launch by the launcher from skypilot_launch.worker_image_tag.
 
+# Task-scoped pod_config override: `imagePullPolicy: Never` makes kubelet
+# use the kind-loaded synth-setter image and never resolve its manifest
+# from Docker Hub, avoiding the anonymous-pull 429 that timed out #1255.
+# Scoped to this task (not ~/.sky/config.yaml) so the SkyPilot jobs
+# controller — whose image lives on GAR and is not `kind load`-ed — still
+# pulls normally; a global override blocks the controller from starting.
+config:
+  kubernetes:
+    pod_config:
+      spec:
+        containers:
+          - imagePullPolicy: Never
+
 resources:
   cloud: kubernetes
   # Resource floor: shrunk from `cpus: 2+` to fit kind's allocatable headroom
--- a/src/synth_setter/pipeline/skypilot_launch.py
+++ b/src/synth_setter/pipeline/skypilot_launch.py
@@ -171,41 +171,21 @@ _CI_MODE_TRUTHY_VALUES = frozenset({"1", "true", "yes", "on"})
 
 # SkyPilot's default (cpus: 4+, memory: 4x) doesn't fit in GHA-kind's
 # ~1950m allocatable CPU after kube-system. See PR #876.
-#
-# `kubernetes.pod_config` forces every CI-kind pod (jobs controller + worker)
-# to use the locally-loaded image and never resolve the manifest from Docker
-# Hub. Without this, kind/containerd still contacts registry-1.docker.io and
-# trips its anonymous-pull 429, even though `kind load docker-image` already
-# placed the image on the node. Container `name`-less entry merges into the
-# pod's first container (SkyPilot patch-merge legacy). See #1255.
 _CI_SKY_CONFIG_YAML = """jobs:
   controller:
     resources:
       cpus: 1+
       memory: 1+
-kubernetes:
-  pod_config:
-    spec:
-      containers:
-        - imagePullPolicy: Never
 """
 
 
 def _ensure_ci_sky_config() -> None:
-    """Write ``~/.sky/config.yaml`` with CI-only SkyPilot overrides when CI mode is truthy.
-
-    Currently emits two overrides: a managed-jobs controller resource shrink so
-    the controller pod fits a GHA-kind cluster (#876), and a
-    ``kubernetes.pod_config.spec.containers[0].imagePullPolicy: Never`` so
-    every CI-kind pod uses the locally-loaded dev-snapshot image instead of
-    racing Docker Hub's anonymous 429 limit (#1255).
+    """Write ``~/.sky/config.yaml`` with the controller shrink when CI mode is truthy.
 
     Truthy = ``SYNTH_SETTER_CI_MODE`` ∈ {1, true, yes, on} (case-insensitive).
     Any other value (including ``0``, ``false``, unset) is a no-op, so an
     operator who exports ``SYNTH_SETTER_CI_MODE=0`` doesn't clobber a local
-    config, and production runs against real clouds never apply
-    ``imagePullPolicy: Never`` (which would prevent the worker from pulling
-    the image at all).
+    config.
     """
     if os.environ.get(_CI_MODE_ENV, "").strip().lower() not in _CI_MODE_TRUTHY_VALUES:
         return
--- a/tests/pipeline/test_entrypoints/test_skypilot_launch.py
+++ b/tests/pipeline/test_entrypoints/test_skypilot_launch.py
@@ -299,16 +299,16 @@ class TestEnsureCiSkyConfig:
         assert "memory: 1+" in body
         assert "controller:" in body
 
-    def test_writes_image_pull_policy_never_under_kubernetes_pod_config(
+    def test_global_config_does_not_carry_pod_config(
         self, tmp_path: Path, monkeypatch: pytest.MonkeyPatch
     ) -> None:
-        """SYNTH_SETTER_CI_MODE=1 → pod_config forces ``imagePullPolicy: Never``.
+        """``~/.sky/config.yaml`` must not set ``kubernetes.pod_config`` globally.
 
-        Pinning the exact nested location is what makes the override actually
-        reach kubelet — SkyPilot's ``combine_pod_config_fields`` merges this
-        into every kind pod (jobs controller + worker), letting the
-        pre-loaded dev-snapshot image satisfy starts without re-resolving the
-        manifest from Docker Hub (#1255).
+        A global ``imagePullPolicy: Never`` blocks the SkyPilot jobs
+        controller — whose image lives on GAR and is not ``kind load``-ed —
+        from starting (#1255 → run 26374103900). The pod-config override
+        belongs in the per-task ``local-template.yaml`` instead, where it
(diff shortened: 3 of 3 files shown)


ktinubu:
## Verification Results

Verified on HEAD `4e0593a206d5c04e28eebed0415bb43368e5f021` (worktree `/home/build/synth-setter/.claude/worktrees/fix-26374103900`).

- [x] **`_ensure_ci_sky_config()` writes only the controller shrink — no global `kubernetes.pod_config`** *(Level 2 — invoked the function and inspected the file it wrote; promise is the function's output shape, not end-to-end)*
  ```bash
  $ python3 -c "
  import os
  os.environ['SYNTH_SETTER_CI_MODE'] = '1'
  os.environ['HOME'] = '/tmp/verify-sky'
  import shutil
  shutil.rmtree('/tmp/verify-sky', ignore_errors=True)
  os.makedirs('/tmp/verify-sky')
  from synth_setter.pipeline.skypilot_launch import _ensure_ci_sky_config
  _ensure_ci_sky_config()
  import yaml
  doc = yaml.safe_load(open('/tmp/verify-sky/.sky/config.yaml').read())
  print('FULL written config:', doc)
  print('---')
  print('PASS check 1 (controller shrink present):', doc.get('jobs', {}).get('controller', {}).get('resources') == {'cpus': '1+', 'memory': '1+'})
  print('PASS check 2 (kubernetes.pod_config NOT present):', 'pod_config' not in doc.get('kubernetes', {}))
  "
  FULL written config: {'jobs': {'controller': {'resources': {'cpus': '1+', 'memory': '1+'}}}}
  ---
  PASS check 1 (controller shrink present): True
  PASS check 2 (kubernetes.pod_config NOT present): True
  ```

- [x] **SkyPilot's loader interprets `local-template.yaml`'s top-level `config:` as a per-task override (lands on `Resources._cluster_config_overrides`, not global)** *(Level 2 — invoked `sky.Task.from_yaml_config` against the real template; this is the load-bearing claim — that SkyPilot reads the field as task-scoped, not just that the YAML has the right shape. If SkyPilot ever renames or moves the top-level `config:` key, this check fails loudly.)*
  ```bash
  $ python3 -c "
  import yaml
  import sky
  with open('src/synth_setter/configs/compute/local-template.yaml') as f:
      doc = yaml.safe_load(f)
  task = sky.Task.from_yaml_config(doc)
  res = list(task.resources)[0]
  print('Task cluster_config_overrides:')
  print(res.cluster_config_overrides)
  print()
  expected = {'kubernetes': {'pod_config': {'spec': {'containers': [{'imagePullPolicy': 'Never'}]}}}}
  print('PASS' if res.cluster_config_overrides == expected else 'FAIL', '— override is task-scoped on Resources._cluster_config_overrides')
  "
  Task cluster_config_overrides:
  {'kubernetes': {'pod_config': {'spec': {'containers': [{'imagePullPolicy': 'Never'}]}}}}

  PASS — override is task-scoped on Resources._cluster_config_overrides
  ```

- [x] **Unit tests covering the new shape pass on this branch** *(Level 2 — runs the same test file that pins both the global-config negative assertion and the per-task positive assertion. Standalone `make test` is CI's job; this is a targeted exercise tied to the PR's promise.)*
  ```bash
  $ python3 -m pytest tests/pipeline/test_entrypoints/test_skypilot_launch.py -q 2>&1 | tail -1
  91 passed in 2.43s
  ```

- [ ] **`Test Dataset Finalization` workflow (the workflow whose failure motivated this PR) completes both matrix rows on this branch** *(Level 1 — SKIP: auto-mode classifier blocked `gh workflow run test-dataset-finalization.yml` as an unauthorized shared-infra action. This is the original failing workflow ([run 26374103900](https://github.com/tinaudio/synth-setter/actions/runs/26374103900)) and the load-bearing end-to-end check.)*

  To verify manually:
  ```bash
  gh workflow run test-dataset-finalization.yml --ref fix/scope-imagepullpolicy-never-to-task
  gh run list -w test-dataset-finalization.yml --branch fix/scope-imagepullpolicy-never-to-task --limit 1
  gh run watch <run-id>
  ```
  Expected: both `Generate + finalize smoke (hdf5)` and `Generate + finalize smoke (wds)` rows finish the `Run synth-setter-generate-dataset against local-template (skypilot-local row)` step in well under the 40-min timeout, and `Verify finalize artifacts` passes for both. The captured kind events should show **no** `Container image ... is not present with pull policy of Never` error on the `sky-jobs-controller-*` pod (the failure mode from run 26374103900).

- [ ] **`Test Dataset Finalization` is also exercised on PR open via the path-trigger** *(Level 1 — SKIP: workflow doesn't trigger on `pull_request` for this PR's diff. The path filter in `test-dataset-finalization.yml` (lines 28-38) lists `src/synth_setter/cli/finalize_dataset.py`, `src/synth_setter/pipeline/data/reshard.py`, etc., but **not** `src/synth_setter/pipeline/skypilot_launch.py` or `src/synth_setter/configs/compute/local-template.yaml`, so the path filter doesn't trip on this diff. Same dispatch instructions as the previous check apply.)*

### Adversarial / implicit-contract checks

- [x] **`_ensure_ci_sky_config()` remains a no-op when `SYNTH_SETTER_CI_MODE` is unset (no global side effects in operator-local-dev)** *(Level 2 — covered by the unaltered `test_no_op_when_env_var_unset` test in the file run above; explicitly part of the 91-passed suite. Verifies that an operator who doesn't opt in to CI mode gets no clobbering of `~/.sky/config.yaml`.)*

- [x] **The override targets the worker container correctly via SkyPilot's name-less patch-merge legacy** *(Level 5 — the merge logic in `sky/utils/config_utils.py::merge_k8s_configs` line ~255-260 reads: "Backward compatibility for containers: if no name is specified, merge into the first container (index 0)". The user-task pod's first container is `ray-node` (per `sky/templates/kubernetes-ray.yml.j2:510`), and `kind load docker-image tinaudio/synth-setter:dev-snapshot` was already loaded into kind via the workflow on line 352 of `.github/workflows/generate-dataset-shards.yaml`. So `imagePullPolicy: Never` reaches `ray-node` and kubelet uses the loaded copy. Cannot promote to Level 1 here without dispatching the workflow.)*

### Residual carry-over from the pre-PR review (non-blocking)

These came out of `/repo-review-full-no-comments` at cd54675 and remain open at 4e0593a2:

- `_CI_SKY_CONFIG_YAML` constant could be renamed to `_CI_CONTROLLER_SHRINK_YAML` now that it carries only one concern (cosmetic — not a behavior change).
- `test_image_pull_policy_never_is_scoped_to_task` could ascend from a YAML-shape assertion to a `sky.Task.from_yaml_config(...)` round-trip — note that the verification above DID run that round-trip, so the contract is exercised here; folding it back into the test is a follow-up tightening.

### Summary

3 / 5 checks PASS via local exercise. The 2 SKIPs are both gated on the same workflow dispatch — recommend running it before merge. The Level 2 SkyPilot loader check is the strongest local proxy: it confirms the override lands as task-scoped (not global), which is the actual contract the PR changes.

PR #1302, opened 2026-05-27
feat(eval): surface audio metrics to wandb + callback_metrics

### LINKED ISSUE #96: feat(eval): W&B metrics logging from compute_audio_metrics.py
## Summary

Optionally log evaluation metrics (MSS, wMFCC, SOT, RMS) to a W&B run for cross-run comparison dashboards.

## Deliverables

- [ ] Add `--wandb-run` CLI flag to `compute_audio_metrics.py`
- [ ] When set, log `aggregated_metrics.csv` values to W&B as summary metrics
- [ ] Optionally log per-sample metrics as a W&B Table
- [ ] Use existing W&B config from `configs/logger/wandb.yaml`

## Acceptance Criteria

- `python scripts/compute_audio_metrics.py output/audio/ output/metrics/ --wandb-run flow_simple_eval` logs to W&B
- Without `--wandb-run`, behavior unchanged (CSV only)

ktinubu:
## Summary

Follow-on to #1291. After ``compute_audio_metrics`` runs inside ``_run_predict_postprocessing`` (``cli/eval.py``), parse ``aggregated_metrics.csv`` and forward the flat ``audio/<name>_{mean,std}`` dict to (a) the active wandb run via ``wandb.run.log`` and (b) the dict returned by ``evaluate()`` alongside ``trainer.callback_metrics``. So the same wandb run that already carries ``test/param_mse`` now also carries the audio metrics.

This is option (A) from the design discussion: the smallest viable change that closes the subprocess → wandb gap. Logging happens in the parent process where the wandb run handle already lives — no ``--wandb-run`` flag needs to be threaded into the subprocess. Logging is gated on a real ``wandb.run`` so ``mode: predict`` with ``logger=null`` is unaffected.

Refs #96 — the original ``--wandb-run`` CLI-flag design on ``compute_audio_metrics.py`` is superseded by routing through the parent process; the maintainer can decide whether to close #96 outright or keep it open for the still-unbuilt per-sample W&B Table follow-up.

## Implementation notes

- New pure helper ``_load_audio_metrics(metrics_dir) -> dict[str, float]`` reads the CSV with ``index_col=0``, pins the expected stat columns to ``{"mean", "std"}`` (raising a directed ``ValueError`` when one is missing), and emits ``audio/<metric>_<stat>``. Raises ``FileNotFoundError`` when the CSV is absent — the producing subprocess just ran, so a missing file is a real bug, not a soft no-op.
- New helper ``_log_audio_metrics_to_wandb`` no-ops when ``wandb.run is None``, otherwise calls ``wandb.run.log(metrics)`` inside a ``try/except`` that logs a warning so a wandb-side failure does not lose the already-computed metrics.
- ``_run_predict_postprocessing`` now returns ``dict[str, float]`` (``{}`` when ``compute_metrics`` is off; the loaded dict when it ran).
- ``evaluate()`` initialises ``audio_metrics = {}``, fills it inside the rank-zero predict branch, then ``metric_dict.update(audio_metrics)`` before returning. ``:return:`` docstring documents the heterogeneous value types (``Tensor`` from Lightning, ``float`` from audio metrics).
- Design doc updated (``docs/design/eval-pipeline.md`` §5.1) to mention the wandb forwarding + ``callback_metrics`` merge.

## Test plan

8 new tests in ``tests/test_eval.py``; all 19 tests in the file pass locally.

- [x] ``test_load_audio_metrics_flattens_mean_and_std`` — CSV → flat ``audio/<name>_<stat>`` mapping
- [x] ``test_load_audio_metrics_returns_python_floats`` — values are plain ``float``, not numpy scalars
- [x] ``test_load_audio_metrics_missing_csv_raises`` — directed ``FileNotFoundError`` naming the producing subprocess
- [x] ``test_load_audio_metrics_missing_stat_column_raises`` — directed ``ValueError`` when the CSV lacks a required column
- [x] ``test_postprocessing_returns_empty_when_compute_metrics_off`` — ``{}`` so callers can unconditionally merge
- [x] ``test_postprocessing_returns_loaded_audio_metrics`` — values reach the caller through the helper return
- [x] ``test_postprocessing_logs_audio_metrics_to_active_wandb_run`` — state-checks against a fake ``wandb.run``
- [x] ``test_postprocessing_skips_wandb_when_no_run`` — no-op when no run is active
- [x] ``pre-commit run --files src/synth_setter/cli/eval.py tests/test_eval.py docs/design/eval-pipeline.md`` — all hooks pass
- [ ] CI smoke + slow ``requires_vst`` legs green

## Pre-PR review

Multi-skill dry-run review (``/repo-review-full-no-comments``) ran across 6 skills (code-health, python-style, tdd-implementation, comment-hygiene, synth-setter-project-standards, ml-test). Findings: **0 BLOCK, 35 WARN**. Sentinel file written to ``.agent-reviews/repo-review-full-no-comments.619ef60bef97dc1d6dab071d3ddb566c0e5632c3.md``. Substantive WARNs (CSV-column pinning, ``wandb.log`` error handling, magic argv index, float-equality fragility, hoisted wandb patch target, new failure-mode test) addressed in commit 619ef60b. Remaining nits (direct ``wandb.run.log`` vs. ``trainer.logger.log_metrics``, mixed ``Tensor``/``float`` dict values) acknowledged in code/docs; a deeper refactor would expand scope beyond option (A).

### DIFF
--- a/docs/design/eval-pipeline.md
+++ b/docs/design/eval-pipeline.md
@@ -251,6 +251,8 @@ When `cfg.mode == "predict"`, `cli/eval.py` invokes `_run_predict_postprocessing
 
 On Linux the render subprocess is prefixed with the headless wrapper materialised via `synth_setter.resources.vst_headless_wrapper()` so the VST3 plugin sees an Xvfb display before pedalboard imports it; the metrics subprocess is CPU-only and runs unwrapped. Both default-off so `mode: test` and `mode: validate` paths are unchanged.
 
+When `evaluation.compute_metrics` runs, the aggregated values from `aggregated_metrics.csv` are also surfaced to the active wandb run (as `audio/<name>_{mean,std}`) and merged into the dict returned by `evaluate()` alongside Lightning's `trainer.callback_metrics`, so the same wandb run that holds `test/param_mse` can carry the audio metrics too.
+
 ### 5.2 Render
 
 | Property     | Value                                                                                                                    |
--- a/docs/reference/wandb-integration.md
+++ b/docs/reference/wandb-integration.md
@@ -143,6 +143,23 @@ Written via `wandb.config.update(..., allow_val_change=True)`.
 | Git SHA + diff | W&B auto-captures if in a git repo             |
 | Python deps    | `pyproject.toml` / `pip freeze` snapshot       |
 
+### 2i. Eval-mode Audio Metrics (predict mode only)
+
+When `synth-setter-eval mode=predict evaluation.compute_metrics=true` runs and a W&B run is active, `_log_audio_metrics_to_wandb` (`src/synth_setter/cli/eval.py`) forwards the values from `aggregated_metrics.csv` directly to `wandb.run.log`, so they land in the same run as `test/param_mse`:
+
+| Key                | What                                                        |
+| ------------------ | ----------------------------------------------------------- |
+| `audio/mss_mean`   | Multi-scale log-mel spectrogram distance, mean over samples |
+| `audio/mss_std`    | Same, standard deviation                                    |
+| `audio/wmfcc_mean` | DTW-aligned MFCC distance, mean                             |
+| `audio/wmfcc_std`  | Same, standard deviation                                    |
+| `audio/sot_mean`   | Spectral optimal-transport distance, mean                   |
+| `audio/sot_std`    | Same, standard deviation                                    |
+| `audio/rms_mean`   | RMS envelope cosine similarity, mean                        |
+| `audio/rms_std`    | Same, standard deviation                                    |
+
+The same dict is also merged into the dict returned by `evaluate()` alongside Lightning's `trainer.callback_metrics`. See [eval-pipeline.md §5.1](../design/eval-pipeline.md) for the surrounding subprocess chain.
+
 ______________________________________________________________________
 
 ## 3. Artifacts
@@ -156,10 +173,10 @@ ______________________________________________________________________
 
 ## 4. Entry Points
 
-| Entry point                     | W&B usage                                                                                       | File                            |
-| ------------------------------- | ----------------------------------------------------------------------------------------------- | ------------------------------- |
-| `src/synth_setter/cli/train.py` | Full: logger init → hparams → provenance → train metrics → test metrics → teardown              | `src/synth_setter/cli/train.py` |
-| `src/synth_setter/cli/eval.py`  | Full: logger init → hparams → provenance → test/val metrics (+ optional predictions) → teardown | `src/synth_setter/cli/eval.py`  |
+| Entry point                     | W&B usage                                                                                                                                                                          | File                            |
+| ------------------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ------------------------------- |
+| `src/synth_setter/cli/train.py` | Full: logger init → hparams → provenance → train metrics → test metrics → teardown                                                                                                 | `src/synth_setter/cli/train.py` |
+| `src/synth_setter/cli/eval.py`  | Full: logger init → hparams → provenance → test/val metrics (+ optional predictions) → predict-mode `audio/<metric>_{mean,std}` keys from `_log_audio_metrics_to_wandb` → teardown | `src/synth_setter/cli/eval.py`  |
 
 Both use `@task_wrapper` which ensures `wandb.finish()` runs even on exception.
 
--- a/src/synth_setter/cli/eval.py
+++ b/src/synth_setter/cli/eval.py
@@ -7,6 +7,8 @@ from pathlib import Path
 from typing import Any
 
 import hydra
+import pandas as pd
+import wandb
 from lightning import Callback, LightningDataModule, LightningModule, Trainer
 from lightning.pytorch.loggers import Logger
 from omegaconf import DictConfig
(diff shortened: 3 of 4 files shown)


ktinubu:
## Verification Results

Verified against worktree HEAD `9cd2a386` (post doc-drift fix). All four behavioral checks ran end-to-end.

- [x] **All 19 eval unit tests pass against the real helpers** *(Level 2 — invokes `_load_audio_metrics`, `_log_audio_metrics_to_wandb`, `_run_predict_postprocessing` directly via pytest)*

  ```
  $ pytest tests/test_eval.py -q --no-header
  ======================== 19 passed, 2 warnings in 4.16s ========================
  ```

  The 8 new tests (`test_load_audio_metrics_flattens_mean_and_std`, `..._returns_python_floats`, `..._missing_csv_raises`, `..._missing_stat_column_raises`, `test_postprocessing_returns_empty_when_compute_metrics_off`, `..._returns_loaded_audio_metrics`, `..._logs_audio_metrics_to_active_wandb_run`, `..._skips_wandb_when_no_run`) all pass alongside the 11 pre-existing ones.

- [x] **`_log_audio_metrics_to_wandb` swallows wandb-side exceptions and emits a directed warning** *(Level 2 — adversarial: not covered by existing tests; the try/except added in commit 619ef60 had no direct behavioral pin)*

  ```
  $ PYTHONPATH=src python3 - <<'PY'
  import logging
  from synth_setter.cli import eval as eval_mod
  records: list[logging.LogRecord] = []
  class _Capture(logging.Handler):
      def emit(self, record): records.append(record)
  eval_mod.log.logger.addHandler(_Capture())
  eval_mod.log.logger.setLevel(logging.DEBUG)
  class _RaisingRun:
      def log(self, payload):
          raise RuntimeError("simulated wandb backend outage")
  eval_mod.wandb.run = _RaisingRun()
  try:
      eval_mod._log_audio_metrics_to_wandb({"audio/mss_mean": 0.5})
      print("RESULT: swallowed (no exception propagated)")
  except Exception as exc:
      print(f"RESULT: did NOT swallow — {type(exc).__name__}: {exc}")
  warn_records = [r for r in records if r.levelno == logging.WARNING]
  print(f"WARNING records emitted: {len(warn_records)}")
  for r in warn_records:
      print(f"  message: {r.getMessage()}")
  eval_mod.wandb.run = None
  PY
  RESULT: swallowed (no exception propagated)
  WARNING records emitted: 1
    message: [rank: 0] wandb.run.log raised RuntimeError: simulated wandb backend outage; metrics still returned.
  ```

- [x] **Real `compute_audio_metrics` CSV round-trips through `_load_audio_metrics`** *(Level 2 — runs the actual producer subprocess against a synthesized 1s stereo `pred.wav`/`target.wav` pair, then feeds the CSV it writes into the loader; pins the contract between the producing and consuming modules without relying on the hand-crafted fixture CSV)*

  ```
  $ PYTHONPATH=src timeout 90 python3 - <<'PY' [...full script in run, omitted for brevity...]
  Running compute_audio_metrics...
  exit=0
  CSV exists: True
  CSV contents:
  ,mean,std
  mss,1.3912883,
  wmfcc,12.190260355472564,
  sot,0.00031039512,
  rms,0.9999966,

  _load_audio_metrics keys: ['audio/mss_mean', 'audio/mss_std', 'audio/rms_mean', 'audio/rms_std', 'audio/sot_mean', 'audio/sot_std', 'audio/wmfcc_mean', 'audio/wmfcc_std']
  All values are float? True
  ```

  Side note surfaced by this check: a single-sample CSV writes empty `std` cells (pandas serializes the `NaN` std as blank), and `_load_audio_metrics` loads them as `float('nan')`. The schema-guard still passes because the `std` column header is present; the NaN flows straight to wandb, which is the established behavior for single-sample evals.

- [x] **Full predict-mode pipeline (train → evaluate → render → metrics → merge) succeeds end-to-end** *(Level 1 — promise is end-to-end; lower levels would not exercise the full contract that audio metrics actually reach the dict `evaluate()` returns)*

  ```
  $ PYTHONPATH=src timeout 480 pytest 'tests/test_train.py::test_train_eval_surge_xt[cpu-surge/fake_oracle]' -m 'requires_vst and slow' -q --no-header
  tests/test_train.py::test_train_eval_surge_xt[cpu-surge/fake_oracle] PASSED  [100%]
  =================== 1 passed, 1 warning in 74.13s (0:01:14) ====================
  ```

  This exercises the real `evaluate(mode=predict, compute_metrics=true)` path: real VST3 render under the headless wrapper, real `compute_audio_metrics` subprocess, real `_load_audio_metrics` → `_log_audio_metrics_to_wandb` → merge into the returned dict. The `[cpu-surge/ffn_full]` parametrization is not run here (~equivalent code path, same gate; skipped to keep verification time bounded — run it locally as a second sanity check before merging if desired).

- [ ] **MPS/GPU parametrizations of the e2e test** *(SKIP — hardware-gated; `mps` legs only run on macOS, `gpu` legs require CUDA; this environment is Linux/CPU. To verify manually on a Mac with MPS: `pytest 'tests/test_train.py::test_train_eval_surge_xt[mps-surge/fake_oracle]' -m 'requires_vst and slow'` after `WANDB_MODE=offline` if needed.)*

### Adversarial gaps discovered during verification

1. **Empty `std` from single-sample eval** → propagates `NaN` into wandb. Not a regression vs. pre-PR (CSV was already written with empty cells), and the merge into `callback_metrics` accepts floats including NaN. Worth a follow-up: emit a `log.warning` when any value is non-finite so a debugger doesn't have to scroll wandb to notice.

2. **Extra CSV columns are silently dropped.** `_load_audio_metrics` iterates `_AGGREGATED_METRICS_STATS = ("mean", "std")`, so an upstream addition (e.g. `median`) wouldn't surface as a new key. The pin protects against silent leakage of unexpected stats, but means downstream additions need a coordinated change in both modules.

3. **No multi-rank/DDP exercise.** The rank-zero gate in `evaluate()` is exercised only by the single-process e2e here; a multi-rank test would need a CI-level DDP harness. Pre-PR code had the same gate, so this is not a regression — flagging only because the e2e test does not prove the gate.

### Doc-drift sweep

Stop hook flagged `docs/reference/wandb-integration.md` was missing the new `audio/*_{mean,std}` keys. Fixed in commit `9cd2a386`: extended §4 row, added §2i listing every key with what it measures and a back-reference to `eval-pipeline.md` §5.1.

PR #1307, opened 2026-05-27
ci: fix by hoist r2 fields out of f-string in nightly-parallel-datagen

### LINKED ISSUE #1237: bug(ci-automation): nightly-parallel-datagen scheduled failure
The nightly-parallel-datagen workflow failed on its scheduled cron. Auto-opened by .github/workflows/nightly-parallel-datagen.yml so the regression in the parallel-dispatch path (or its Xvfb / R2 dependencies) does not go silently stale.

- Failed run: https://github.com/tinaudio/synth-setter/actions/runs/26209263766
- Refs #1189 (parallel feature), #1197 (this workflow)

ktinubu:
## Summary

- The nightly cron's `Resolve spec_uri from local materialized spec` step has been failing every night (latest: [run 26494852896, job 78020533537](https://github.com/tinaudio/synth-setter/actions/runs/26494852896/job/78020533537), tracked in #1237) with:

  ```
  File "<string>", line 5
      print(f"r2://{r2[\"bucket\"]}/{r2[\"prefix\"]}input_spec.json")
  SyntaxError: f-string expression part cannot include a backslash
  ```

- The runner is `ubuntu-22.04` (Python 3.10); the `\"` inside the f-string expression part is rejected before 3.12. The script is wrapped in a single-quoted shell heredoc, so inner `'` isn't an option either.

- Read `bucket` / `prefix` into locals before the f-string so the expression parts only reference bare names. No behavior change. Regression came from #1303 (the step was introduced there to replace the old stdout-sentinel routing).

## Test plan

- [x] Reproduce the failure locally on Python 3.10 with the original snippet — `SyntaxError` as observed in CI.
- [x] Run the fixed snippet locally on Python 3.10 — prints the expected `r2://<bucket>/<prefix>input_spec.json` URI.
- [ ] Next scheduled nightly run (or manual `workflow_dispatch`) succeeds end-to-end through the validate step.

Fixes #1237.

### DIFF
--- a/.github/workflows/nightly-parallel-datagen.yml
+++ b/.github/workflows/nightly-parallel-datagen.yml
@@ -169,6 +169,8 @@ jobs:
       # URI isn't reproducible from ``(experiment, run_id)`` alone. Read it from
       # the locally-written spec under
       # ``<workspace>/data/<task_name>/<run_id>/metadata/input_spec.json``.
+      # Hoist ``r2["bucket"]`` / ``r2["prefix"]`` to locals: Python <3.12 rejects
+      # ``\"`` inside f-string parts, and the single-quoted heredoc forbids ``'``.
       - name: Resolve spec_uri from local materialized spec
         id: spec_uri
         env:
@@ -185,9 +187,9 @@ jobs:
           SPEC_PATH="${specs[0]}"
           SPEC_URI=$(python3 -c '
           import json, sys
-          spec = json.load(open(sys.argv[1]))
-          r2 = spec["r2"]
-          print(f"r2://{r2[\"bucket\"]}/{r2[\"prefix\"]}input_spec.json")
+          r2 = json.load(open(sys.argv[1]))["r2"]
+          bucket, prefix = r2["bucket"], r2["prefix"]
+          print(f"r2://{bucket}/{prefix}input_spec.json")
           ' "${SPEC_PATH}")
           R2_PREFIX=$(python3 -c '
           import json, sys


ktinubu:
## Verification Results

Verifying the promise: **the `Resolve spec_uri from local materialized spec` step's `python3 -c` snippet parses on the runner's Python 3.10 (ubuntu-22.04) and emits the canonical `r2://<bucket>/<prefix>input_spec.json` URI for a real `DatasetSpec`-shaped input.**

- [x] **New snippet parses and emits the expected URI on Python 3.10** *(Level 2 — runner uses ubuntu-22.04 / Python 3.10.12; promise is "the inline Python actually runs," so invoking the interpreter with the same source is the right level)*
  ```bash
  $ cat > /tmp/spec.json <<'JSON'
  {"r2": {"bucket": "intermediate-data", "prefix": "data/nightly-parallel-26494852896-1-3622bf49/run-1/"}}
  JSON
  $ python3 -c '
  import json, sys
  r2 = json.load(open(sys.argv[1]))["r2"]
  bucket, prefix = r2["bucket"], r2["prefix"]
  print(f"r2://{bucket}/{prefix}input_spec.json")
  ' /tmp/spec.json
  r2://intermediate-data/data/nightly-parallel-26494852896-1-3622bf49/run-1/input_spec.json
  ```

- [x] **The pre-fix snippet still reproduces the exact CI SyntaxError on Python 3.10** *(Level 2 adversarial — pins the regression so a partial revert breaks loudly; same Python 3.10.12 interpreter as above)*
  ```bash
  $ python3 -c '
  import json, sys
  spec = json.load(open("/tmp/spec.json"))
  r2 = spec["r2"]
  print(f"r2://{r2[\"bucket\"]}/{r2[\"prefix\"]}input_spec.json")
  '
    File "<string>", line 5
      print(f"r2://{r2[\"bucket\"]}/{r2[\"prefix\"]}input_spec.json")
                                                                    ^
  SyntaxError: f-string expression part cannot include a backslash
  ```
  Matches the CI log from [job 78020533537](https://github.com/tinaudio/synth-setter/actions/runs/26494852896/job/78020533537) byte-for-byte.

- [x] **Snippet output matches `DatasetSpec.r2.input_spec_uri()` against a Hydra-composed real spec** *(Level 2 — cross-checks the inline Python against the canonical `synth-setter-spec-uri` derivation that `generate-dataset-shards.yaml` uses; ensures both call sites stay in agreement)*
  ```bash
  $ uv run --no-project python3 -c "
  import json, sys, tempfile, pathlib, subprocess
  sys.path.insert(0, 'src')
      cfg = compose(config_name='dataset', overrides=['experiment=generate_dataset/smoke-shard', '+run_id=verify-1307'])
  cfg.paths.root_dir = cfg.paths.output_dir = cfg.paths.work_dir = '.'
  raw = OmegaConf.to_container(cfg, resolve=True)
  NON_SPEC = {'data', 'paths', 'hydra', 'run_name', 'skypilot_launch'}
  spec = DatasetSpec(**{k: v for k, v in raw.items() if k not in NON_SPEC})
  p = pathlib.Path(tempfile.mkstemp(suffix='_input_spec.json')[1])
  p.write_text(spec.model_dump_json())
  out = subprocess.check_output(['python3', '-c', '''
  import json, sys
  r2 = json.load(open(sys.argv[1]))[\"r2\"]
  bucket, prefix = r2[\"bucket\"], r2[\"prefix\"]
  print(f\"r2://{bucket}/{prefix}input_spec.json\")
  ''', str(p)], text=True).strip()
  print('canonical URI:', spec.r2.input_spec_uri())
  print('snippet URI:  ', out)
  print('match:', out == spec.r2.input_spec_uri())
  "
  canonical URI: r2://intermediate-data/data/smoke-shard/verify-1307/input_spec.json
  snippet URI:   r2://intermediate-data/data/smoke-shard/verify-1307/input_spec.json
  match: True
  ```

- [ ] **End-to-end: nightly-parallel-datagen succeeds when triggered via `workflow_dispatch` on this branch** *(Level 1 — SKIP: triggering this shared production workflow goes beyond the PR-open ask and was blocked by the auto-mode classifier.)*
  To verify manually before merge:
  ```bash
  gh workflow run nightly-parallel-datagen.yml --ref fix/nightly-parallel-fstring-backslash
  gh run watch  # ~10–30 min; succeeds end-to-end through the validate step
  ```
  This is the only check that exercises the full step (including the materialized-spec glob and `GITHUB_OUTPUT`/`GITHUB_ENV` writes) in the real runner environment with R2 credentials. The three Level 2 checks above prove the Python source is valid and the URI extraction is correct against the real `DatasetSpec` schema, but only the workflow trigger proves the surrounding bash + step wiring still works.

Issue #1328, opened 2026-05-28
chore(devcontainer): restore claude/gh tmux panes on rebuild instead of bare bash


ktinubu:
## Problem

The devcontainer runs tmux with tmux-resurrect + tmux-continuum to auto-save/restore sessions across container rebuilds (added in #1103, root variant in #1121). After a rebuild, continuum restores sessions and working directories correctly, but every pane that was running an interactive `claude` (Claude Code CLI) or `gh` process comes back as a bare `bash` shell.

Root cause: tmux-resurrect only restores a fixed default program list (`vi vim nvim emacs man less more tail top htop irssi weechat mutt`) unless extra programs are opted in via the `@resurrect-processes` option. `claude` and `gh` are not in the default list, so they are dropped to bash on restore. We lost 8 running Claude sessions on a recent rebuild.

## Fix

Add to `.devcontainer/tmux.conf`:

    set -g @resurrect-processes '"~claude" "~gh"'

The `~` prefix is resurrect's loose-match (matches the program regardless of its arguments, e.g. `claude --resume <uuid>`). Per the plugin source (`scripts/process_restore_helpers.sh:_restore_list`), `@resurrect-processes` **appends** to the default list rather than replacing it, so the defaults remain intact.

## Limitation

This restores the program fresh — it relaunches `claude`/`gh` in the right cwd. It does NOT restore Claude's in-app conversation state; the user must `/resume` inside the relaunched claude.

PR #1329, opened 2026-05-28
chore(devcontainer): restore claude/gh tmux panes on rebuild

ktinubu:
## Problem

The devcontainer runs tmux with tmux-resurrect + tmux-continuum to auto-save/restore sessions across container rebuilds. After a rebuild, continuum restores sessions and their working directories correctly, but every pane that was running an interactive `claude` (Claude Code CLI) or `gh` process comes back as a bare `bash` shell. We lost 8 running Claude sessions on a recent rebuild and had to relaunch them manually.

Root cause: tmux-resurrect only restores a fixed default program list (`vi vim nvim emacs man less more tail top htop irssi weechat mutt`) unless extra programs are opted in via `@resurrect-processes`. `claude` and `gh` aren't in that list, so they drop to bash on restore.

## Fix

Add one line to `.devcontainer/tmux.conf` (which `post-create.sh` installs verbatim to `~/.tmux.conf`):

    set -g @resurrect-processes '"~claude" "~gh"'

The `~` prefix is resurrect's loose-match — it matches the program regardless of arguments (e.g. `claude --resume <uuid>`). Verified against the installed plugin source (`scripts/process_restore_helpers.sh` `_restore_list`): `@resurrect-processes` **appends** to the default list rather than replacing it, so all the built-in programs stay restored. The edited config was sanity-checked with `tmux -L <throwaway> source-file` (parses clean).

## Limitation (be honest)

This restores the **program** fresh — it relaunches `claude`/`gh` in the correct working directory. It does **NOT** restore Claude's in-app conversation state. After a rebuild the user still has to run `/resume` inside the relaunched `claude` to pick up the prior conversation.

Refs #1328

### DIFF
--- a/.devcontainer/tmux.conf
+++ b/.devcontainer/tmux.conf
@@ -49,6 +49,9 @@ set -g @continuum-save-interval '5'
 # Pin resurrect state to a fixed path so a named volume in devcontainer.json
 # can mount it and survive container rebuilds.
 set -g @resurrect-dir '~/.local/share/tmux/resurrect'
+# Revive claude/gh panes as themselves on rebuild; resurrect drops non-default
+# programs to bare bash otherwise. Relaunches fresh (no in-app state).
+set -g @resurrect-processes '"~claude" "~gh"'
 
 # Must be the last line. Bootstraps TPM and loads the plugins above.
 run '~/.tmux/plugins/tpm/tpm'
--- a/docs/doc-map.yaml
+++ b/docs/doc-map.yaml
@@ -226,7 +226,7 @@ docs:
       - pattern: ".devcontainer/root_gpu/devcontainer.json"
         covers: "Root-user GPU dev container variant — remoteUser defaults to root, docker-in-docker + act features, --gpus all, VS Code terminal profile defaulting to tmux"
       - pattern: ".devcontainer/tmux.conf"
-        covers: "tmux config installed to ~/.tmux.conf by post-create.sh — mouse mode, true color, 1-indexed windows, larger scrollback, vi copy mode, intuitive split keys, TPM plugin manager + tmux-resurrect + tmux-continuum (auto-restore on, 5-min save interval, resurrect state pinned to ~/.local/share/tmux/resurrect so the named volume can persist it)"
+        covers: "tmux config installed to ~/.tmux.conf by post-create.sh — mouse mode, true color, 1-indexed windows, larger scrollback, vi copy mode, intuitive split keys, TPM plugin manager + tmux-resurrect + tmux-continuum (auto-restore on, 5-min save interval, resurrect state pinned to ~/.local/share/tmux/resurrect so the named volume can persist it, claude/gh panes revived as themselves on restore)"
       - pattern: ".devcontainer/initialize.sh"
         covers: "Host-side initializeCommand — refuses pointer-file `.git` (worktree/submodule/--separate-git-dir), ensures .env exists"
       - pattern: ".devcontainer/post-create.sh"


ktinubu:
## Verification Results

Verified against the installed `tmux-resurrect` plugin (`~/.tmux/plugins/tmux-resurrect/scripts/`) on the PR branch.

- [x] **claude/gh panes are selected for restore (incl. with arguments); defaults still restore; unrelated programs drop to bash** *(Level 2 — invokes resurrect's real restore-decision function `_process_on_the_restore_list` with the PR's `@resurrect-processes` value loaded into a live tmux server, rather than parsing the config. Lower than Level 1 because a true end-to-end check requires a container rebuild + saved state, which isn't reproducible in the PR context.)*
  ```bash
  $ SOCK="verify_$$"
  $ tmux -L "$SOCK" new-session -d -s keep 'sleep 300'
  $ tmux -L "$SOCK" set -g @resurrect-processes "$(grep '@resurrect-processes' .devcontainer/tmux.conf | sed "s/.*@resurrect-processes '//; s/'$//")"
  $ bash -c '
      RES=$HOME/.tmux/plugins/tmux-resurrect/scripts
      source "$RES/variables.sh"; source "$RES/helpers.sh"; source "$RES/process_restore_helpers.sh"
      get_tmux_option(){ v=$(tmux -L "'"$SOCK"'" show -gqv "$1"); [ -z "$v" ] && echo "$2" || echo "$v"; }
      expect(){ w=$1; shift; _process_on_the_restore_list "$1" && g=RESTORE || g=drop; [ "$g" = "$w" ] && echo "ok ($g) <= $1" || { echo "FAIL want=$w got=$g <= $1"; fail=1; }; }
      expect RESTORE "claude"
      expect RESTORE "claude --resume 1a2b3c4d-5e6f-7890-abcd-ef0123456789"
      expect RESTORE "node /usr/lib/node_modules/@anthropic/claude/cli.js --resume xyz"
      expect RESTORE "gh"; expect RESTORE "gh pr checks 1329 --watch"
      expect RESTORE "vim foo.py"; expect RESTORE "htop"
      expect drop "python train.py"; expect drop "bash"; expect drop "make test-fast"
      echo "RESULT: $([ ${fail:-0} -eq 0 ] && echo ALL-PASS || echo SOME-FAILED)"'
  ok (RESTORE) <= claude
  ok (RESTORE) <= claude --resume 1a2b3c4d-5e6f-7890-abcd-ef0123456789
  ok (RESTORE) <= node /usr/lib/node_modules/@anthropic/claude/cli.js --resume xyz
  ok (RESTORE) <= gh
  ok (RESTORE) <= gh pr checks 1329 --watch
  ok (RESTORE) <= vim foo.py
  ok (RESTORE) <= htop
  ok (drop) <= python train.py
  ok (drop) <= bash
  ok (drop) <= make test-fast
  RESULT: ALL-PASS
  ```
  The `vim`/`htop` rows prove the option **appends** to resurrect's default program list rather than replacing it (matches `_restore_list`: `echo "$default_processes $user_processes"`).

- [x] **The edited `tmux.conf` parses without errors on a live tmux server** *(Level 2 — sources the full file via `source-file`, which is what tmux does at startup)*
  ```bash
  $ tmux -L parse_$$ new-session -d -s s 'sleep 5' \; source-file "$PWD/.devcontainer/tmux.conf" \; show -gv @resurrect-processes
  "~claude" "~gh"
  $ echo "parse-exit=$?"
  parse-exit=0
  ```

- [x] **`post-create.sh` installs `tmux.conf` byte-for-byte, so editing the source is the only change required** *(Level 2 — runs the exact `install` command the script uses and byte-compares)*
  ```bash
  $ grep -n "tmux.conf" .devcontainer/post-create.sh | head -1
  16:install -m 0644 "$_devc_dir/tmux.conf" "$HOME/.tmux.conf"
  $ TMP=$(mktemp); install -m 0644 .devcontainer/tmux.conf "$TMP"; cmp -s .devcontainer/tmux.conf "$TMP" && echo IDENTICAL || echo DIFFERENT
  IDENTICAL
  ```

- [ ] **End-to-end: a `claude`/`gh` pane survives an actual container rebuild as itself** *(Level 1 — SKIP: requires rebuilding the devcontainer with the resurrect named volume populated, which can't be reproduced in the PR CI context.)*
  To verify manually:
  ```bash
  # 1. In a devcontainer using this branch's ~/.tmux.conf, open a tmux pane and run `claude`.
  # 2. Wait for continuum's autosave (<=5 min) or force: tmux run-shell ~/.tmux/plugins/tmux-resurrect/scripts/save.sh
  # 3. Rebuild the container (Dev Containers: Rebuild Container).
  # 4. Re-attach to tmux; confirm the pane comes back running `claude` (not bash).
  #    Note: `claude` relaunches fresh — run /resume inside it to restore the conversation.
  ```

Issue #91, opened 2026-03-19
internal-feat(storage): R2 dataset download in Hydra config


ktinubu:
## Summary

Add a `data.r2_path` option to Hydra data configs so `SurgeDataModule.prepare_data()` can sync datasets from R2 before training/eval.

## Current State

- `SurgeDataModule` expects data at a local `dataset_root` path
- No mechanism to download from R2 — user must manually sync data
- `prepare_data()` hook exists in Lightning but isn't used for remote fetch

## Deliverables

- [ ] Add `r2_path: Optional[str] = null` to `configs/data/surge*.yaml`
- [ ] In `SurgeDataModule.prepare_data()`: if `r2_path` is set and `dataset_root` is empty/missing, sync from R2
- [ ] Use rclone wrapper from the sibling issue (feat(storage): rclone wrapper utility)
- [ ] No-op if data already exists locally (skip download)
- [ ] `--checksum` always used
- [ ] Integration test: mock rclone, verify `prepare_data()` calls sync when data is missing

## Key Files

- `src/data/surge_datamodule.py` — `SurgeDataModule.prepare_data()` (currently empty/minimal)
- `configs/data/surge.yaml`, `surge_simple.yaml`, `surge_mini.yaml` — add `r2_path` field
- `src/data/storage.py` — rclone wrapper (from sibling issue)

## Example Usage

```bash
python src/eval.py \
    experiment=surge/flow_simple \
    data.r2_path=r2:synth-data/surge-simple/ \
    data.dataset_root=data/surge-simple/ \
    mode=predict \
    ckpt_path=checkpoints/best.ckpt
```

## Acceptance Criteria

- Data auto-downloaded from R2 on first run
- Subsequent runs skip download (data exists)
- Works with `make predict` and `make train`

ktinubu:
### Dependencies

**Blocks:** —
**Blocked by:** #90 (rclone wrapper), #94 (clean config paths — `data.r2_path` extends data configs)

PR #1341, opened 2026-06-02
internal-feat(eval): inline oracle-eval in the generate_dataset render sweep

ktinubu:
## Summary

Stacked on top of #1338 (`feat/eval-from-r2`). Makes the
`generate_dataset_cadence` sweep run end-to-end — generate → finalize →
inline oracle-eval (`mode=predict`) — logging `audio/{mss,rms,sot,wmfcc}_{mean,std}`
to wandb. The fake oracle echoes target params, so all audio metrics are `0`
(perfect round-trip), which is the expected pass signal.

**Datamodule owns its own data.** `SurgeDataModule` gains a
`download_dataset_root_uri` ctor arg and hydrates `dataset_root` from R2 in
`prepare_data()` — Lightning's rank-0-only hook, so it's DDP-safe and matches
the `kosc`/`ksin`/`fm` datamodules. `eval`/`train` now build the datamodule with
plain `hydra.utils.instantiate(cfg.data)`; the `download_dataset` /
`instantiate_datamodule` helpers are gone (the latter only existed to strip the
download key before instantiation). A `null` URI is a no-op.

**Supporting fixes:**
- `data/*.yaml`: anchor `dataset_root` via the `${hydra:runtime.output_dir}`
  resolver (the dot-lookup form can't resolve under `@hydra.main`).
- `surge_datamodule`: `predict_file` defaults to `<dataset_root>/test.h5`; the
  predict split is always built and `teardown` closes its h5 handle too.
- `spec.py`: mask non-spec keys before resolving so bare `compose()` works.
- `test-mps-fake-oracle`: `render: null` (not MISSING) so `extras()` resolves
  the config tree in `mode=test`.

## Commits

1. `feat(eval): inline oracle-eval …` — the feature + initial datamodule self-download.
2. `refactor(data): download in prepare_data; address review` — moves the
   download to `prepare_data`, removes a dead `predict_dataset` branch + h5
   handle leak, drops the unused `DatasetSpec.file_extension` field and a broken
   unreferenced `sweeps/eval_render.yaml`, and fixes two stale comments.

## Test plan

- `tests/data/test_surge_datamodule.py` — `test_prepare_data_hydrates_dataset_root_from_r2_when_uri_set`
  and `test_prepare_data_no_download_when_uri_none` pass against a real rclone
  local-backed remote; `test_teardown_closes_open_h5_handles` now pins the
  predict handle too.
- `tests/test_eval.py`, `tests/test_train.py` (incl. `fast_dev_run` through the
  trainer), `tests/test_configs.py`, `tests/pipeline/test_configs/test_experiment_yamls.py`,
  `tests/pipeline/test_entrypoints/test_generate_dataset.py`,
  `tests/pipeline/test_schemas/test_dataset_spec.py` — green (`-m "not slow"`).
- `pre-commit` clean on all changed files.

Refs #91


ktinubu:
## Verification Results

Branch `feat/eval-render-sweep` @ `5d01cc39`. Behavioral checks against the real CLI/datamodule, not diff-grep. Run from a fresh worktree (which required recreating the gitignored `plugins/Surge XT.vst3 -> /usr/lib/vst3/Surge XT.vst3` symlink that `make`/CI set up — see note on the last item).

- [x] **R2 dataset hydration runs in `prepare_data()` through the real `eval` CLI** *(Level 1 — full entrypoint: instantiate → prepare_data download → setup → trainer.test → oracle → metrics.json)*

  ```bash
  $ uv run pytest "tests/test_eval.py::test_eval_cli_downloads_dataset_from_r2_then_scores_oracle" -m "" -p no:randomly -q
  tests/test_eval.py::test_eval_cli_downloads_dataset_from_r2_then_scores_oracle PASSED [100%]
  1 passed in 41.97s
  ```

  The CLI builds `SurgeDataModule` via plain `hydra.utils.instantiate(cfg.data)`, `prepare_data()` copies the dataset from `r2://intermediate-data/dataset` into an initially-absent `dataset_root` (real `rclone`, local-backed remote), then the fake oracle's exact-zero `test/param_mse` reaches `metrics.json`.

- [x] **`prepare_data()` is opt-in: hydrates when a URI is set, no-op when `null`** *(Level 2/3 — real `rclone` local remote; promise is the hook's behavior)*

  ```bash
  $ uv run pytest tests/data/test_surge_datamodule.py -m "" -p no:randomly -q \
      -k "prepare_data or teardown_closes_open or without_predict_file_defaults"
  ...test_prepare_data_hydrates_dataset_root_from_r2_when_uri_set PASSED [ 25%]
  ...test_prepare_data_no_download_when_uri_none                  PASSED [ 50%]
  ...test_setup_without_predict_file_defaults_to_test_split       PASSED [ 75%]
  ...test_teardown_closes_open_h5_handles                          PASSED [100%]
  4 passed, 63 deselected
  ```

- [x] **Download is deferred out of `__init__` (the DDP-safety point of moving to `prepare_data`)** *(Level 2 — behavioral: construction must not perform I/O)*

  ```bash
  $ uv run python -c "
  import os
  for k in list(os.environ):
      if k.startswith('RCLONE_CONFIG_R2'): del os.environ[k]
  from synth_setter.data.surge_datamodule import SurgeDataModule
  m = SurgeDataModule(dataset_root='/tmp/nonexistent-ds', download_dataset_root_uri='r2://bogus/prefix')
  print('__init__ with URI did NOT download/raise -> deferred:', m.download_dataset_root_uri)
  assert hasattr(m, 'prepare_data')
  print('prepare_data hook present:', m.prepare_data.__qualname__)
  "
  __init__ with URI did NOT download/raise -> deferred: r2://bogus/prefix
  prepare_data hook present: SurgeDataModule.prepare_data
  ```

  With no R2 env set, a download in `__init__` would have raised in `ensure_r2_env_loaded`; construction succeeding proves the I/O moved to the rank-0 `prepare_data` hook.

- [x] **Predict split is always built and `teardown` closes its h5 handle** *(Level 2 — covered by `test_teardown_closes_open_h5_handles` above, which now also asserts `not module.predict_dataset.dataset_file`, and `test_setup_without_predict_file_defaults_to_test_split`)*

- [x] **`download_dataset` / `instantiate_datamodule` helpers removed; `eval`/`train` use plain instantiate** *(Level 2 — invokes the import + the helpers are exercised away by the Level-1 CLI test above)*

  ```bash
  $ uv run python -c "import synth_setter.utils as u; assert not hasattr(u,'download_dataset'); assert not hasattr(u,'instantiate_datamodule'); print('download_dataset/instantiate_datamodule: REMOVED')"
  download_dataset/instantiate_datamodule: REMOVED
  ```

- [ ] **Inline oracle-eval logs `audio/*` through the full generate→finalize→eval CLI** *(Level 1 — SKIP/partial: the CLI runs end-to-end and exits 0, but the wrapping integration test fails on an unrelated assertion)*

  ```bash
  $ uv run pytest "tests/integration/test_generate_dataset_cli_wandb_e2e.py::test_oracle_eval_inline_resumes_generate_wandb_run" -m "" -p no:randomly -q
  # _assert_cli_succeeded (line 399) PASSED -> the generate→finalize→inline-oracle-eval
  # CLI exited 0: datamodule instantiate, prepare_data, setup, trainer.predict, and the
  # VST headless re-render all succeed.
  # Then FAILS at line 405: expected oracle_eval dir under SYNTH_SETTER_WORKSPACE, but
  # generate_dataset.main() writes it to cfg.paths.output_dir/oracle_eval/<run_id> (line 791).
  ```

  This remaining failure is in untouched code (`generate_dataset.main()` oracle_eval dir placement vs the test's `operator_workspace()` assumption) and is **independent of this PR's datamodule changes** — the CLI exit-0 confirms the datamodule path. Flagging it as a pre-existing branch test/code mismatch worth a separate look. Note: a fresh worktree must recreate the gitignored `plugins/Surge XT.vst3` symlink (`ln -sfn '/usr/lib/vst3/Surge XT.vst3' 'plugins/Surge XT.vst3'`) before any VST re-render runs locally.

**Not included as checkboxes** (CI's job): full `pytest` / `pre-commit` runs — both green locally on the changed files.

PR #1350, opened 2026-06-02
test(eval): align inline oracle-eval e2e with current behavior

ktinubu:
## Summary

Fixes the pre-existing failure in `test_oracle_eval_inline_resumes_generate_wandb_run` (surfaced while verifying #1341). The test was added in #1335 and is gated `slow`+`r2`, so it never ran in CI and drifted from the code on two points:

1. **oracle_eval dir location.** The test looked for `oracle_eval/<run_id>/` under `operator_workspace()` / `$SYNTH_SETTER_WORKSPACE`, but `generate_dataset.main()` writes it under `cfg.paths.output_dir` (= `${hydra:runtime.output_dir}`, the per-run hydra dir, which is gitignored). The code is correct — `operator_workspace()` defaults to the checkout root, which would write `oracle_eval/` into the repo. The test now asserts against the launcher's hydra dir (and the local is renamed `eval_output_dir`).
2. **eval metrics namespace.** The test asserted `test/*` history rows from Lightning's `trainer.test`, but the inline eval now runs `mode=predict` and logs `audio/*` metrics via `wandb.run.log` (eval.py:83). The test now asserts `audio/*`.

Also corrects the matching stale `test/*` / `test_step` references in `_run_oracle_eval_subprocess`'s docstring and a comment (doc-side of the same drift). Comment- and test-only — no behavior change.

## Test plan

```
$ uv run pytest "tests/integration/test_generate_dataset_cli_wandb_e2e.py::test_oracle_eval_inline_resumes_generate_wandb_run" -m "" -p no:randomly -q
1 passed in 202.46s
```

End-to-end: one CLI invocation runs generate → finalize → inline oracle-eval (`mode=predict`), the eval resumes the generate wandb run, and `audio/*` rows land in the resumed run's history. (Requires the gitignored `plugins/Surge XT.vst3` symlink that `make`/CI set up.)

Refs #91

### DIFF
--- a/docs/doc-map.yaml
+++ b/docs/doc-map.yaml
@@ -56,7 +56,7 @@ docs:
       - pattern: "src/synth_setter/cli/eval.py"
         covers: "Checkpoint resolution from W&B, evaluation logging"
       - pattern: "src/synth_setter/cli/generate_dataset.py"
-        covers: "WandbLogger lifecycle for dataset generation: spec → hyperparams + dataset-spec artifact, per-shard + summary history rows, run id pinned to spec.run_id, inline oracle eval resuming the run to append test/* rows"
+        covers: "WandbLogger lifecycle for dataset generation: spec → hyperparams + dataset-spec artifact, per-shard + summary history rows, run id pinned to spec.run_id, inline oracle eval resuming the run to append audio/* rows"
       - pattern: "src/synth_setter/configs/dataset.yaml"
         covers: "Default `logger: wandb` for dataset generation; entity/project sourced from logger/wandb.yaml"
       - pattern: "sweeps/**"
--- a/docs/reference/wandb-integration.md
+++ b/docs/reference/wandb-integration.md
@@ -260,11 +260,11 @@ Each trial subprocess opens its own wandb run with `id = spec.run_id`; the
 When `oracle_eval_inline=true`, the local-run path shells out to
 `synth_setter.cli.eval` after `generate(...)` has closed its run.
 `_run_oracle_eval_subprocess` (`src/synth_setter/cli/generate_dataset.py`)
-re-opens the same run via `+logger.wandb.id=<spec.run_id> +logger.wandb.resume=must`, so Lightning's `trainer.test` deposits `test/*`
-metrics (e.g. `test/param_mse` from the surge fake-oracle) onto the generate
-run — a `wandb sync` then merges both phases under one run id. The eval
-subprocess inherits `WANDB_MODE` from the launcher, so its offline/online
-posture follows the parent's.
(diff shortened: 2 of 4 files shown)


ktinubu:
## Verification Results

Branch `fix/oracle-eval-workspace-dir` @ `5d28e84b`. The promise is narrow: the previously-red, slow+r2-gated e2e test now passes against the current code, asserting the *correct* behavior (oracle_eval under `cfg.paths.output_dir`; `audio/*` history rows from `mode=predict`).

- [x] **The inline-oracle-eval e2e passes end-to-end on this branch** *(Level 1 — this test IS the full generate→finalize→inline-oracle-eval integration; the change is the test itself, so running it is the promise)*

  ```bash
  $ uv run pytest "tests/integration/test_generate_dataset_cli_wandb_e2e.py::test_oracle_eval_inline_resumes_generate_wandb_run" -m "" -p no:randomly -q
  tests/integration/test_generate_dataset_cli_wandb_e2e.py::test_oracle_eval_inline_resumes_generate_wandb_run PASSED [100%]
  1 passed in 148.54s
  ```

  One CLI invocation runs generate → finalize → inline oracle-eval; the test then reads the **real** offline wandb binaries (no mocks) and confirms: the eval's `oracle_eval/<run_id>/` dir exists under the launcher's hydra dir, its run-id slug matches generate's (the eval resumed the same run), generate logged `shard/bytes`+`shards/rendered`, and the eval logged at least one `audio/*` row.

- [x] **The two corrected assertions are what was actually red before** *(Level 1 — observed by running the same test against the base earlier this session)*
  Against `feat/eval-from-r2`, the same test failed first at the dir assertion (`expected eval workspace dir at .../oracle_eval/<run_id>` — main writes it under `cfg.paths.output_dir`, not `operator_workspace()`), and after that was corrected, at the metrics assertion (`eval dir has no test/* history rows` — `mode=predict` logs `audio/*`, never `test/*`). Both are now green, so the fix targets the real failures, not phantom ones.

- [x] **`mode=predict` genuinely logs `audio/*` (not `test/*`) to the resumed run** *(Level 1 — the passing `audio/*` assertion above reads the eval's offline wandb history binary; this is the observable contract behind eval.py's `wandb.run.log(audio_metrics)`)*

**Adversarial note (env):** a fresh worktree must recreate the gitignored `plugins/Surge XT.vst3` symlink (`ln -sfn '/usr/lib/vst3/Surge XT.vst3' 'plugins/Surge XT.vst3'`) before any VST re-render runs locally — CI/`make` already do this. Without it the eval re-render fails with a plugin-not-found error unrelated to this change.

**Not included as checkboxes** (CI's job): full `pytest`/`pre-commit` — both green locally on the two changed files. The production-side change (`generate_dataset.py` docstring/comment `test/*`→`audio/*`) is documentation only; its correctness is exercised by the `audio/*` assertion above.

PR #1361, opened 2026-06-02
refactor(pipeline): one DatasetSpec.from_hydra_cfg helper (mask-before-resolve)

ktinubu:
## Summary

Collapses the two parallel "Hydra-composed cfg → `DatasetSpec`" implementations onto a single allowlist-based helper, eliminating the duplication that let them drift (and caused the `test_spec_uri` regression fixed in the base branch).

- `DatasetSpec.from_hydra_cfg(cfg)` (new, in `pipeline/schemas/spec.py`): masks the cfg to the model's own fields (`k in cls.model_fields`) **before** `resolve=True`, so non-spec subtrees whose interpolations need `@hydra.main` (e.g. `data.dataset_root: ${...output_dir}/data`) are never evaluated. Raises `TypeError` on a non-mapping.
- `cli/generate_dataset.spec_from_cfg` is now a one-line wrapper; **`_NON_SPEC_KEYS` deleted**.
- `pipeline/ci/spec_uri.compute_spec_uri_from_hydra` is now `DatasetSpec.from_hydra_cfg(cfg).r2.input_spec_uri()`; **deleted** the `cfg.paths.* = "."` placeholder block, the resolve-then-filter, `_NON_SPEC_CFG_KEYS`, and the now-unused import.
- Removed the twin-constant **sync contract test** (single source of truth now). Added `TestFromHydraCfg` pinning the mask-before-resolve regression (a `${hydra:...}` interpolation under a non-spec group with no `HydraConfig` set must not raise) and non-mapping rejection.

The allowlist (`cls.model_fields`) is self-maintaining: add a spec field, it's picked up; add a non-spec config group, it's ignored — no constant to keep in sync. `DatasetSpec`'s top-level fields have no aliases, so the allowlist matches the composed cfg keys exactly.

This is the structural fix behind the `${paths.output_dir}` config swap that landed in the base PR (#1338); with the helper, the interpolation form no longer matters because the subtree is dropped before resolving.

## Test plan
- `tests/pipeline/test_ci/test_spec_uri.py`, `test_entrypoints/test_generate_dataset.py`, `test_configs/test_experiment_yamls.py`, `test_schemas/` → **347 passed** (incl. `TestComputeSpecUriFromHydra`, `TestSpecFromCfg`).
- `synth-setter-spec-uri --from-experiment generate_dataset/smoke-shard --run-id-override test` → `r2://intermediate-data/data/smoke-shard/test/input_spec.json`, exit 0.
- `pre-commit` green on all changed files.

Refs #91

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -179,3 +179,6 @@ thoughts/
 
 # Surge-XT mini-example local artifacts (e.g. learned-model presets)
 presets/*.fxp
+
+# wandb local run/sweep artifacts
+wandb/
--- a/docs/doc-map.yaml
+++ b/docs/doc-map.yaml
@@ -56,7 +56,7 @@ docs:
       - pattern: "src/synth_setter/cli/eval.py"
         covers: "Checkpoint resolution from W&B, evaluation logging"
       - pattern: "src/synth_setter/cli/generate_dataset.py"
-        covers: "WandbLogger lifecycle for dataset generation: spec → hyperparams + dataset-spec artifact, per-shard + summary history rows, run id pinned to spec.run_id, inline oracle eval resuming the run to append test/* rows"
+        covers: "WandbLogger lifecycle for dataset generation: spec → hyperparams + dataset-spec artifact, per-shard + summary history rows, run id pinned to spec.run_id, inline oracle eval resuming the run to append audio/* rows"
       - pattern: "src/synth_setter/configs/dataset.yaml"
         covers: "Default `logger: wandb` for dataset generation; entity/project sourced from logger/wandb.yaml"
       - pattern: "sweeps/**"
--- a/docs/reference/wandb-integration.md
+++ b/docs/reference/wandb-integration.md
@@ -260,11 +260,11 @@ Each trial subprocess opens its own wandb run with `id = spec.run_id`; the
 When `oracle_eval_inline=true`, the local-run path shells out to
 `synth_setter.cli.eval` after `generate(...)` has closed its run.
 `_run_oracle_eval_subprocess` (`src/synth_setter/cli/generate_dataset.py`)
-re-opens the same run via `+logger.wandb.id=<spec.run_id> +logger.wandb.resume=must`, so Lightning's `trainer.test` deposits `test/*`
-metrics (e.g. `test/param_mse` from the surge fake-oracle) onto the generate
-run — a `wandb sync` then merges both phases under one run id. The eval
-subprocess inherits `WANDB_MODE` from the launcher, so its offline/online
-posture follows the parent's.
+re-opens the same run via `logger.wandb.id=<spec.run_id> +logger.wandb.resume=must`, runs `mode=predict` with `render=surge_simple` to
+re-render the predicted params, and deposits `audio/*` audio-similarity
+metrics onto the generate run — a `wandb sync` then merges both phases under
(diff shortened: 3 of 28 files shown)


ktinubu:
## Verification Results

Branch `refactor/spec-from-hydra-cfg` @ `8404315`. Promise: the spec-uri `--from-experiment` path resolves a dataset cfg whose `data.dataset_root` uses a `@hydra.main`-only resolver **without `HydraConfig` set** (no `InterpolationResolutionError`), and both call sites still produce identical specs.

- [x] **`synth-setter-spec-uri --from-experiment` returns the canonical URI** *(Level 1 — the real CLI path: compose → `from_hydra_cfg` → `DatasetSpec` → URI, run end-to-end)*

  ```bash
  $ uv run synth-setter-spec-uri --from-experiment generate_dataset/smoke-shard --run-id-override test
  r2://intermediate-data/data/smoke-shard/test/input_spec.json
  ```

- [x] **The masking is load-bearing — old full-resolve raises, the helper succeeds** *(Level 1 — before/after on the same composed cfg; proves the regression is genuinely fixed, not masked)*

  ```bash
  $ uv run python -c "
      cfg = compose(config_name='dataset', overrides=['experiment=generate_dataset/smoke-shard', '+run_id=test'])
  try:
      OmegaConf.to_container(cfg, resolve=True)
      print('OLD resolve-everything: NO error (unexpected)')
  except Exception as e:
      print(f'OLD resolve-everything raises: {type(e).__name__} (this is why masking is needed)')
  spec = DatasetSpec.from_hydra_cfg(cfg)
  print(f'NEW from_hydra_cfg: OK -> task_name={spec.task_name}, uri={spec.r2.input_spec_uri()}')
  "
  OLD resolve-everything raises: InterpolationResolutionError (this is why masking is needed)
  NEW from_hydra_cfg: OK -> task_name=smoke-shard, uri=r2://intermediate-data/data/smoke-shard/test/input_spec.json
  ```

  This is the exact failure mode of the 5 red `test_spec_uri` tests (resolve-then-filter tripped on `data.dataset_root`); `from_hydra_cfg` masks the subtree away before resolving.

- [x] **Both call sites produce identical specs + the allowlist drops/keeps the right keys** *(Level 2 — the parity test asserts `compute_spec_uri_from_hydra` == launcher-side `spec_from_cfg`; `test_experiment_yamls` builds real specs from 14 composed experiment cfgs, so a dropped spec field or leaked non-spec key would fail there; `TestFromHydraCfg` pins the mask-before-resolve regression + non-mapping rejection)*

  ```bash
  $ uv run pytest tests/pipeline/test_ci/test_spec_uri.py tests/pipeline/test_entrypoints/test_generate_dataset.py tests/pipeline/test_configs/test_experiment_yamls.py tests/pipeline/test_schemas/ -p no:randomly -q
  347 passed in 8.56s
  ```

  Includes `TestComputeSpecUriFromHydra::test_matches_launcher_side_computation` (parity) and `TestSpecFromCfg` (the `spec_from_cfg` wrapper).

**Not included as checkboxes** (CI's job): full `pytest`/`pre-commit`. One unrelated pre-existing failure in the broader suite — `test_generate_dataset_renders_shards_to_r2` needs the real Surge XT VST3 binary (absent in sandbox); not touched by this diff.

Issue #1243, opened 2026-05-22
chore(deps): Migrate dependency management to `uv lock` + `uv sync` (hybrid PyTorch routing)


ktinubu:
# Migrate dependency management to `uv lock` + `uv sync` (hybrid PyTorch routing)

**Labels:** `infra`, `build`, `reproducibility`
**Type:** Tracking issue (multi-PR)

## Summary

We install dependencies imperatively today:

```bash
uv pip install --torch-backend ${TORCH_BACKEND} -r pyproject.toml --extra torch --extra dev
```

This produces **no lockfile**. Every install re-resolves the full dependency graph from our loose constraints, so transitive versions drift silently — between developer machines, between CI runs, and over time. For a project whose purpose is studying a specific, measurable effect, the failure mode is concrete: *we cannot reliably reconstruct the environment a published result was produced in.* A reviewer (or future us) asking "reproduce the numbers in Table 3" should not be blocked on "we don't know exactly which versions were installed that month."

This issue tracks three coupled changes:

1. Adopt a committed `uv.lock` as the single source of truth for installs.
2. Restructure `pyproject.toml` so PyTorch resolves correctly across all our hardware targets from that one lockfile.
3. Add a CI gate that verifies the lock stays valid on every platform we claim to support.

It also folds in two cheap dependency-hygiene fixes that naturally ride along with the relock.

---

## Background: why this is not a one-line change

The blocker is that **`--torch-backend` only works in the `uv pip` interface.** It does *not* work with `uv lock`, `uv sync`, or `uv run`. So we cannot keep our current auto-detect flow *and* gain a lockfile — adopting a lock requires moving the PyTorch backend routing into `pyproject.toml`.

### Is losing `--torch-backend=auto` a problem?

No, and arguably it's an improvement. `--torch-backend=auto` runs *only at install time*: it shells out to `nvidia-smi`, reads the CUDA driver version, and picks a matching wheel index. It has zero runtime presence — the installed `torch` behaves identically regardless of how the wheel was selected. No application code changes or breaks.

What we lose is the convenience of auto-detecting unknown/heterogeneous GPU drivers. What we *gain* is consistency: every GPU box gets the **same** pinned CUDA wheel from the lock, rather than a per-machine choice that varies with whatever driver each box happens to have. For reproducible research, a deliberate locked CUDA version is the correct direction, not a regression.

[...]

PR #1364, opened 2026-06-02
chore(ci): add uv-lock pre-commit hook

ktinubu:
## What

Adds [`astral-sh/uv-pre-commit`](https://github.com/astral-sh/uv-pre-commit) (rev `0.11.18`) and enables the `uv-lock` hook so `uv.lock` is kept in sync with `pyproject.toml` on every commit.

## Why

Without this hook, `uv.lock` can silently drift from `pyproject.toml`. Enabling it surfaced exactly that drift: the lockfile recorded the project at `8.13.0` while `pyproject.toml` declares `8.14.0`. This PR includes the resulting relock so the hook passes in CI.

## Changes

- `.pre-commit-config.yaml`: add the `uv-lock` hook (placed next to `validate-pyproject`).
- `uv.lock`: relock project version `8.13.0` → `8.14.0`.

Refs #1243

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -92,6 +92,12 @@ repos:
     hooks:
       - id: validate-pyproject
 
+  # uv: keep uv.lock in sync with pyproject.toml
+  - repo: https://github.com/astral-sh/uv-pre-commit
+    rev: 0.11.18
+    hooks:
+      - id: uv-lock
+
   # gitlint: lints commit messages for conventional commits compliance
   - repo: https://github.com/jorisroovers/gitlint
     rev: v0.19.1
--- a/docs/doc-map.yaml
+++ b/docs/doc-map.yaml
@@ -21,6 +21,8 @@ docs:
         covers: "Committed universal lockfile (macOS + Linux + Windows resolution); regenerate with `uv lock` after pyproject edits and review the diff in PRs"
       - pattern: ".github/workflows/uv-lock-check.yml"
         covers: "macOS + Linux matrix running `uv lock --check` on every PR that touches pyproject.toml or uv.lock; catches stale-lock drift"
+      - pattern: ".pre-commit-config.yaml"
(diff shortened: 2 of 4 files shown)


ktinubu:
## Verification Results

Verified on branch `chore/uv-lock-pre-commit` (HEAD `67c8074`). The promise: the `uv-lock` hook is correctly wired and keeps `uv.lock` in sync with `pyproject.toml` — passing when synced, failing/relocking on drift.

- [x] **Pre-commit config is valid and the uv-lock hook is registered** *(Level 2 — invoked `pre-commit validate-config`; promise is that pre-commit accepts the new repo/hook entry)*
  ```bash
  $ pre-commit validate-config .pre-commit-config.yaml; echo "exit=$?"
  exit=0
  ```

- [x] **uv-lock hook passes on this tree (lock in sync with pyproject)** *(Level 2 — runs the actual hook via pre-commit, not a YAML parse)*
  ```bash
  $ pre-commit run uv-lock --all-files
  uv-lock..................................................................Passed
  ```

- [x] **Hook actively detects pyproject↔lock drift** *(Level 2, adversarial — proves the hook enforces sync, not just that it is wired in. Temporarily bumped pyproject version, ran the hook, then restored.)*
  ```bash
  $ sed -i 's/^version = "8.14.0"/version = "8.14.1"/' pyproject.toml
  $ pre-commit run uv-lock --all-files
  - files were modified by this hook
  Resolved 327 packages in 580ms
  Updated synth-setter v8.14.0 -> v8.14.1
  $ git checkout -- pyproject.toml uv.lock   # restored; tree clean
  ```

- [x] **The included relock resolves the real drift this PR fixes** *(Level 2 — lock and pyproject now agree)*
  ```bash
  $ grep -m1 '^version' pyproject.toml
  version = "8.14.0"
  $ grep -A1 'name = "synth-setter"' uv.lock | grep version
  version = "8.14.0"
  ```

**Level 1 note:** the end-to-end firing of this hook on a real commit was already exercised when this branch was committed — the commit ran the full pre-commit suite and `uv-lock` reported `Passed`. CI runs the same hook suite on the PR.

Issue #1124, opened 2026-05-19
test(pipeline): migrate R2 tests from rclone subprocess.check_call mocks to a fake-local R2 fixture


ktinubu:
## Context

Today, every R2-using test in the repo (uploads, downloads, size probes) patches `subprocess.check_call` and asserts on the rclone args list. This couples tests to internal command shape (the exact \`copy\` vs \`copyto\` verb, flag ordering, etc.) rather than to the **observable behavior** they care about: \"the file lands at the destination with the right contents.\"

See [the top-level review on PR #1120](https://github.com/tinaudio/synth-setter/pull/1120#pullrequestreview-4319039621) for the motivation and the recommended shape:

```python
def test_upload_lands_file_at_destination(tmp_path, monkeypatch):
    monkeypatch.setenv("RCLONE_CONFIG_R2_TYPE", "local")
    src = tmp_path / "hello.txt"
    src.write_text("hello world")
    dest_dir = tmp_path / "bucket"
    upload_to_r2(str(src), f"r2:{dest_dir}/uploaded.txt")
    assert (dest_dir / "uploaded.txt").read_text() == "hello world"
```

This is a state-based test — it lets the real rclone binary do the work against a local-typed remote, and asserts on filesystem state at the end.

## Plan

1. **Introduce a `fake_r2_remote` fixture** (likely autouse-scoped, opt-in via marker) that:
   - Sets `RCLONE_CONFIG_R2_TYPE=local` and points the remote root at a `tmp_path` subdirectory.
   - Optionally yields the `tmp_path / \"bucket\"` directory so tests can assert on the materialized files.

2. **Migrate the existing R2 test cases** (drive with `/tdd-refactor` for safety — write the state-based test first, get it green, then delete the mock-based test):
   - `tests/pipeline/test_r2_io.py` — upload/download/size-probe tests.
   - `tests/pipeline/test_spec_io.py` — \`upload_spec\` rclone-flag assertions.
   - `tests/pipeline/test_entrypoints/test_generate_dataset.py` — the \`_stub_upload_spec\` and \`_rclone_copy\` mock patterns.
   - Any other call site that patches \`subprocess.check_call\` for R2.

3. **Keep one or two thin tests on the rclone-flag layer** (the \`--checksum / --contimeout / --timeout / --retries\` set is a real invariant — losing those flags would be a correctness regression). Those can stay as targeted unit tests on a single helper.

## Acceptance

[...]

PR #1377, opened 2026-06-03
test(r2-io): replace rclone argv mocks with behavioral state tests

### LINKED ISSUE #1375: test(r2-io): finish migrating test_r2_io.py argv mocks to fake_r2_remote state tests
## Context

Follow-up to #1124 (closed). That migration left three `subprocess.check_call` argv-pinning tests in `tests/pipeline/test_r2_io.py` that still assert on the rclone command/flag list rather than on observable filesystem state — the interaction-testing pattern #1124 set out to remove.

## Scope

- `download_dir_no_overwrite` — `test_command_carries_immutable_and_reliability_flags` → behavioral `--immutable` test (re-download over a locally-drifted file asserts the real `CalledProcessError`).
- `upload_to_uri` — `test_command_carries_rclone_reliability_flags` → re-upload-overwrites test (proves absence of `--immutable`); landing/`copyto` already covered by existing state tests.
- `upload` (R2→R2) — `test_r2_uri_source_uses_rclone_copyto_with_reliability_flags` → deleted as redundant with `test_r2_uri_source_copies_remote_to_remote`.

Reliability-flag (`--retries`/`--timeout`/`--contimeout`) argv assertions are dropped: config tuning, unobservable in a local fake remote, change-detector when pinned via argv.

Refs #1124.

ktinubu:
Finishes the migration #1124 started: removes the last three `subprocess.check_call` argv-pinning tests in `tests/pipeline/test_r2_io.py` in favor of state/behavioral assertions against the `fake_r2_remote` fixture.

## Changes

- **`download_dir_no_overwrite`** — `test_command_carries_immutable_and_reliability_flags` → `test_redownload_over_drifted_file_hard_fails`. Downloads a tree, drifts a local file, re-downloads, and asserts the real `subprocess.CalledProcessError` — exercising `--immutable` end-to-end instead of grepping argv. (Verified: without `--immutable` the re-download silently overwrites and exits 0.)
- **`upload_to_uri`** — `test_command_carries_rclone_reliability_flags` → `test_reupload_overwrites_remote_object`. Re-uploads changed bytes and asserts the remote object is overwritten (the publish-overwrite contract). The `copyto`/landing contract is already covered by the existing `test_lands_local_bytes_at_remote_uri` + `test_preserves_destination_filename_not_source_basename`.
- **`upload` (R2→R2)** — deleted `test_r2_uri_source_uses_rclone_copyto_with_reliability_flags` as redundant: `test_r2_uri_source_copies_remote_to_remote` already proves the remote copy and path translation from filesystem state (with `copy` instead of `copyto`, bytes would land nested and the assertion would fail).

## Intentionally unpinned flags

Reliability flags (`--retries`/`--timeout`/`--contimeout`/`-vv`) are dropped — they are config tuning, unobservable in a local fake remote, and pinning them via argv is change-detector testing. The `--checksum` invariant ("All rclone operations use --checksum") is also left unpinned here: it cannot be observed from filesystem state, and pinning it would re-introduce the exact `subprocess.check_call` argv-mock pattern this PR removes. It remains enforced in `r2_io` source and by review. (`tests/pipeline/test_spec_io.py` keeps a separate argv guard for the spec-upload path.)

All 52 tests in `tests/pipeline/test_r2_io.py` pass; pre-commit clean.

Closes #1375
Refs #1124

### DIFF
--- a/tests/pipeline/test_r2_io.py
+++ b/tests/pipeline/test_r2_io.py
@@ -3,9 +3,9 @@
 State-based tests use the ``fake_r2_remote`` fixture (see ``tests/pipeline/
 conftest.py``): rclone runs against a local-typed remote rooted at a tmp dir,
 so each test asserts on the filesystem state after the helper runs instead of
-on the rclone argv list. Two narrow argv-shape tests survive (the reliability-
-flag set on ``upload_to_uri`` and the ``lsf --format=s`` shape on
-``object_size``) to pin invariants state-based tests cannot observe.
+on the rclone argv list. A few narrow tests still inspect the argv (e.g. the
+``lsf --format=s`` shape on ``object_size``) or stub ``subprocess.run`` to
+simulate non-zero exits — invariants the filesystem state cannot reveal.
 """
 
 from __future__ import annotations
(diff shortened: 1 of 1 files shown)


ktinubu:
## Verification Results

Reviewed at: `e02c160aabfd32554c5be1395f9d5071e3502b23`

This PR converts three rclone argv-mock tests into state/behavioral tests. The meaningful verification is not "tests pass" (CI does that) but that the new behavioral tests **actually catch the regressions they claim to guard** — i.e. they are behavior tests, not change-detectors. Each guard was checked by mutating the production source and confirming the test fails.

- [x] **Migrated tests pass end-to-end against the real `rclone` binary** *(Level 1 — `fake_r2_remote` runs real rclone against a local-typed remote; no mock)*
  ```bash
  $ uv run pytest tests/pipeline/test_r2_io.py -p no:cacheprovider -q | tail -1
  52 passed in 0.82s
  ```

- [x] **`test_redownload_over_drifted_file_hard_fails` genuinely guards `--immutable`** *(Level 1 — mutate source, prove the test fails)*
  ```bash
  # Remove the first --immutable arg from r2_io.py (download_dir_no_overwrite), then run the test:
  $ sed -i '0,/        "--immutable",/{/        "--immutable",/d;}' src/synth_setter/pipeline/r2_io.py
  $ uv run pytest "tests/pipeline/test_r2_io.py::TestDownloadDirNoOverwrite::test_redownload_over_drifted_file_hard_fails" -p no:cacheprovider -q | tail -2
  FAILED ... - Failed: DID NOT RAISE <class 'subprocess.CalledProcessError'>
  1 failed in 0.40s
  $ git checkout -- src/synth_setter/pipeline/r2_io.py   # reverted
  ```
  Without `--immutable`, the re-download silently overwrites and exits 0 — so the test is a real behavioral guard, not a string check.

- [x] **Deleting the R2→R2 argv test lost no `copyto`-verb coverage** *(Level 1 — mutate source, prove the surviving state test fails)*
  ```bash
  # Change the R2->R2 copyto to copy (line 265), then run the surviving state test:
  $ sed -i '265s/"copyto"/"copy"/' src/synth_setter/pipeline/r2_io.py
  $ uv run pytest "tests/pipeline/test_r2_io.py::TestUpload::test_r2_uri_source_copies_remote_to_remote" -p no:cacheprovider -q | tail -2
  FAILED ... - IsADirectoryError: [Errno 21] Is a directory: ...
  1 failed in 0.40s
  $ git checkout -- src/synth_setter/pipeline/r2_io.py   # reverted
  ```
  With `copy`, bytes land nested under `dst/key.json/key.json` and the state assertion fails — so `test_r2_uri_source_copies_remote_to_remote` already pins the verb the deleted argv test asserted.

- [ ] **`--checksum` / reliability-flag presence** *(SKIP — intentionally unpinned; see PR body)*
  These flags are unobservable from filesystem state in a local fake remote, and pinning them requires the exact `subprocess.check_call` argv-mock pattern this PR removes. To guard `--checksum` manually if desired:
  ```bash
  grep -n '--checksum' src/synth_setter/pipeline/r2_io.py   # present on all rclone invocations
  ```

### Note found during verification
`test_reupload_overwrites_remote_object` proves the **publish-overwrite** contract (a second push replaces the remote object), but it does **not** guard absence of `--immutable`: `rclone copyto` (file→file) ignores `--immutable` entirely regardless of size — only `rclone copy` (the download dir-sync path) honors it. The test docstring was corrected to claim only the overwrite contract it actually verifies.

PR #1384, opened 2026-06-03
automation: give headless review-agent runs an --allowedTools allowlist so gh works

### LINKED ISSUE #1382: fix(ci-automation): headless pr-review-resolver/doc-drift agents can't run gh (auto-denied, no allowlist)
## Bug

The `pr-review-resolver` and `doc-drift` PostToolUse hooks (`agent/hooks/_lib.sh` → `run_agent_prompt`) launch `claude -p` headlessly in a detached worktree. With no TTY, the interactive permission prompt can't be answered, so any tool that isn't pre-approved is **auto-denied**.

`pr-review-resolver` needs `gh` to fetch and reply to PR review comments — every `gh` call (even `gh --version`) returns "This command requires approval" and the agent bails without resolving anything. Observed on PR #1374, where the resolver produced a report saying it was blocked on `gh`.

Root cause: the detached worktree contains only tracked files, so it lacks the primary checkout's untracked `.claude/settings.local.json` (which is what grants `gh` in interactive sessions), and the committed `.claude/settings.json` has no `permissions` block.

## Fix

Pass an explicit `--allowedTools` allowlist to the headless `claude -p` invocation, covering what the agents need (`gh`, `git`, file tools, and the `make`/`python`/`uv`/`pre-commit` commands the resolver uses to validate fixes). Overridable via `AGENT_ALLOWED_TOOLS`. PreToolUse safety hooks still fire on top of the allowlist.

## Done when

- The headless `claude -p` command carries `--allowedTools` including `Bash(gh:*)`.
- `AGENT_ALLOWED_TOOLS` overrides the default list.
- Covered by `agent/hooks/test.sh`.

Part of #372.

ktinubu:
## What

The `pr-review-resolver` and `doc-drift` PostToolUse hooks launch `claude -p` headlessly via `run_agent_prompt` (`agent/hooks/_lib.sh`). Pass an explicit `--allowedTools` allowlist to that invocation so the agents can actually run their tools, overridable via `AGENT_ALLOWED_TOOLS`.

## Why

`claude -p` has no TTY, so the interactive permission prompt can't be answered and any not-preapproved tool is **auto-denied**. `pr-review-resolver` needs `gh` to fetch and reply to PR review comments — on PR #1374 it bailed with "This command requires approval" on every `gh` call and resolved nothing. The detached worktree also lacks the primary checkout's untracked `.claude/settings.local.json` (which grants `gh` interactively), and the committed `.claude/settings.json` has no `permissions` block.

## How

- Default allowlist: `gh`, `git`, `jq`, first-class `Read/Edit/Write/Grep/Glob/LS`, and `make/python/uv/pre-commit` for validating fixes. Shell-text utilities (`sed`, `cat`, …) are intentionally **excluded** — they're redundant with the first-class tools and `sed -i` could edit files past the `Edit|Write` PreToolUse guard hooks.
- Prompt is ordered before the variadic `--allowedTools` so it isn't swallowed.
- `AGENT_ALLOWED_TOOLS` (space-separated) replaces the default; empty/unset falls back via an array-length guard (also avoids a `set -u` empty-array error on stock macOS bash).
- PreToolUse safety hooks (trailer-check, baseline-freeze, worktree-guard, yaml-run) fire on top of the allowlist regardless, so commits stay guarded.

## Tests

Two new cases in `agent/hooks/test.sh` (full suite 83/0): the headless invocation carries `--allowedTools` incl. `Bash(gh:*)`, and `AGENT_ALLOWED_TOOLS` overrides the default.

## Notes for reviewers

- Pre-PR review (`/repo-review-full-no-comments`, 4 passes): 0 BLOCK, 6 WARN — all addressed in-commit except the deliberate `Bash(git:*)` breadth (guarded by PreToolUse hooks + detached-worktree isolation). Detail in the rendered report.

Fixes #1382

### DIFF
--- a/agent/hooks/_lib.sh
+++ b/agent/hooks/_lib.sh
@@ -129,9 +129,27 @@ run_agent_prompt() {
   else
     log "no GNU timeout/gtimeout on PATH; running agent without timeout enforcement"
   fi
+  # No TTY in a detached worktree, so the permission prompt can't be answered
+  # and any not-preapproved tool is auto-denied — pass an explicit allowlist,
+  # shared by the resolver and doc-drift runners. AGENT_ALLOWED_TOOLS replaces
+  # it (space-separated; empty/unset keeps the default). File inspection goes
+  # through first-class Read/Grep/Glob, not Bash(sed/cat/…), so edits can't slip
+  # past the Edit|Write guard hooks.
+  local -a allowed
+  read -r -a allowed <<< "${AGENT_ALLOWED_TOOLS:-}"
+  if [[ ${#allowed[@]} -eq 0 ]]; then
+    allowed=(
+      "Bash(gh:*)" "Bash(git:*)" "Bash(jq:*)"
+      Read Edit Write Grep Glob LS
+      "Bash(make:*)" "Bash(python:*)" "Bash(python3:*)" "Bash(uv:*)" "Bash(pre-commit:*)"
+    )
+  fi
+  # The prompt precedes --allowedTools: the option is variadic, so anything
+  # after it is swallowed into the tool list instead of read as the prompt.
   local -a cmd
   case "$cli" in
-    claude) cmd=(claude -p "$prompt") ;;
+    claude) cmd=(claude -p "$prompt" --allowedTools "${allowed[@]}") ;;
+    # codex uses its own approval/sandbox model; it has no --allowedTools.
     codex)  cmd=(codex exec "$prompt") ;;
     *)
       printf 'No supported headless agent CLI found; install claude or codex (AGENT_HEADLESS=%s).\n' \
--- a/agent/hooks/test.sh
+++ b/agent/hooks/test.sh
@@ -425,6 +425,50 @@ T_resolver_report_lands_in_main_repo() {
 }
 it "pr-review-resolver: report file lands in main repo .agent-reviews/, not the worktree" T_resolver_report_lands_in_main_repo
 
+T_resolver_headless_agent_gets_allowedtools_for_gh() {
+  local report
+  export RESOLVER_SLEEP_SECS=1 GH_STUB_PR=99
+  unset CLAUDE_STUB_PWD_FILE CLAUDE_STUB_HEAD_FILE AGENT_ALLOWED_TOOLS
+  resolver_setup_feature_branch "feature-allowedtools"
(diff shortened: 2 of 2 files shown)


ktinubu:
## Verification Results

The promise is causal: *with* `--allowedTools "Bash(gh:*)"` a headless `claude -p` can run `gh`; *without* it, `gh` is auto-denied (the bug). Verified end-to-end by exercising the actual changed code path (`run_agent_prompt`) against the real `claude` CLI, plus a flagless negative control that reproduces the original failure.

- [x] **The headless agent runs `gh` through the changed code path** *(Level 1 — invoked the real `run_agent_prompt` from `agent/hooks/_lib.sh` with the real `claude` CLI, not a stub)*
  ```bash
  $ cd .../worktrees/headless-gh-perms
  $ export HOOK_NAME=verify AGENT_HEADLESS=claude AGENT_TIMEOUT_SECS=150
  $ source agent/hooks/_lib.sh
  $ run_agent_prompt 'Run exactly this shell command: gh --version — then reply with ONLY the raw output of that command, nothing else.'
  gh version 2.92.0 (2026-04-28)
  https://github.com/cli/cli/releases/tag/v2.92.0
  # exit 0
  ```
  The agent actually executed `gh` and returned its output — the resolver's blocking failure mode is gone.

- [x] **Negative control: without `--allowedTools`, `gh` is auto-denied (reproduces the bug)** *(Level 1 — same headless prompt, flag removed, run from a worktree with no `settings.local.json` — i.e. the detached-hook-worktree condition)*
  ```bash
  $ timeout --kill-after=10 150 claude -p 'Run exactly this shell command: gh --version — then reply with ONLY the raw output of that command, nothing else.'
  The `gh --version` command requires approval and I wasn't able to run it. Please approve the command (or grant permission for `gh`) and I'll return its raw output.
  # exit 0 — gh never ran
  ```
  This is verbatim the failure the `pr-review-resolver` advisory hit on PR #1374 ("every `gh` invocation … returning 'This command requires approval'"). The only difference between the two runs is the flag this PR adds — so the fix is causal.

- [x] **The argv contract is pinned by unit tests** *(Level 2 — `agent/hooks/test.sh`, 83/0)*
  `T_resolver_headless_agent_gets_allowedtools_for_gh` asserts the headless command carries `--allowedTools` incl. `Bash(gh:*)`; `T_resolver_agent_allowed_tools_env_overrides_default` asserts `AGENT_ALLOWED_TOOLS` overrides it. These guard against the flag silently disappearing in a future refactor.

PR #1389, opened 2026-06-03
test(testing): harden offline-wandb history read against async flush race

### LINKED ISSUE #1378: bug(testing): offline-wandb history flush race flakes test_generate_logs_per_shard_and_summary_metrics_offline (conda)
## Symptom

`tests/test_generate_dataset_wandb.py::test_generate_logs_per_shard_and_summary_metrics_offline` fails intermittently on the **conda** test job only:

```
AssertionError: expected 2 per-shard history rows, got 0: []
assert 0 == 2
```

First observed: [run 26850084587](https://github.com/tinaudio/synth-setter/actions/runs/26850084587/job/79179729041) (PR #1363, a config-only change unrelated to wandb). The same commit passed `run_tests_conda` on `main`, and the identical test passed on `run_tests` (MPS), `run_tests_ubuntu` (3.10/3.11), and `run_tests_macos` in the same CI run — so it is non-deterministic, not a logic regression.

## Root cause

`tests/helpers/wandb_offline.read_history_rows` decodes the offline `run-*.wandb` datastore binary in a single pass immediately after `generate()` returns. The wandb offline writer flushes history records **asynchronously**, so on a slower/conda-resolved writer the scan can race ahead of the records and observe zero history rows. The all-or-nothing "0 rows" shape (rather than dropped specific keys) is the tell of a flush race rather than a decode bug. `wandb` is unpinned in `pyproject.toml`, so the conda env can resolve a build with different flush timing than the pip jobs.

## Fix

Make `read_history_rows` flush-aware: add an optional `until=<predicate>` that re-scans the binary until the predicate holds or a timeout elapses, returning the last scan on timeout so the caller's own assertion still reports the shortfall. Wire the predicate into all four call sites.

## Acceptance

- `read_history_rows` polls until the expected history rows materialize.
- All offline-wandb consumers pass the predicate matching what they assert.
- No regression / slowdown on the happy path (rows already flushed → returns on first scan).

ktinubu:
## What & why

`tests/test_generate_dataset_wandb.py::test_generate_logs_per_shard_and_summary_metrics_offline` flaked on the **conda** CI job ([run 26850084587](https://github.com/tinaudio/synth-setter/actions/runs/26850084587/job/79179729041)) with `expected 2 per-shard history rows, got 0`. The same commit passed `run_tests_conda` on `main` and the identical test passed on MPS, ubuntu (3.10/3.11), and macOS in the same run — a non-deterministic flake, not a regression. The triggering PR (#1363) was a config-only change unrelated to wandb.

**Root cause:** `tests/helpers/wandb_offline.read_history_rows` decoded the offline `run-*.wandb` datastore in a single pass right after `generate()` returned. The wandb offline writer flushes history records asynchronously, so on a slower (conda-resolved) writer the scan races ahead of the records and sees zero rows. The all-or-nothing "0 rows" shape is the tell of a flush race rather than a decode bug.

## Change

- `read_history_rows` gains an optional `until=<predicate>`: when passed, it re-scans until the predicate holds or `timeout_s` (default 10s, 50ms poll) elapses, **returning the last scan on timeout** so the caller's own assertion still reports the real shortfall. `until=None` keeps the exact single-shot behavior.
- Split the decode into `_scan_history_rows` (the unchanged single pass) and close the datastore handle each scan so the polling loop can't leak a file handle per re-scan.
- Wired the predicate into all four offline-wandb consumers (the originally-failing test + 3 e2e call sites), each matching what it then asserts.
- New `tests/test_wandb_offline_helper.py` pins the retry / deadline-timeout / single-shot paths deterministically (stubbed scan + fake clock, asserting scan and sleep counts).

## Verification

- `test_generate_dataset_wandb.py` (both offline tests) + 3 new helper tests: pass.
- The originally-failing test run 100× under xdist: 100/100 before and after locally — this Linux/non-conda env (matching the CI jobs that passed) cannot reproduce the conda-specific race, so the deterministic proof is in the unit tests, which simulate the flush lag directly. No regression / no slowdown (happy path returns on the first scan).
- pre-commit (ruff, pyright, pydoclint, docformatter, interrogate, codespell) clean.

Fixes #1378

### DIFF
--- a/tests/helpers/wandb_offline.py
+++ b/tests/helpers/wandb_offline.py
@@ -14,6 +14,8 @@ updating one site.
 
 from __future__ import annotations
 
+import time
+from collections.abc import Callable
 from pathlib import Path
 
 import pytest
@@ -33,31 +35,70 @@ except ImportError as exc:  # pragma: no cover — guards a future wandb upgrade
     )
 
 
-def read_history_rows(wandb_binary: Path) -> list[dict[str, str]]:
+_FLUSH_TIMEOUT_S = 10.0
+_FLUSH_POLL_S = 0.05
+
+
+def read_history_rows(
+    wandb_binary: Path,
+    *,
+    until: Callable[[list[dict[str, str]]], bool] | None = None,
+    timeout_s: float = _FLUSH_TIMEOUT_S,
+) -> list[dict[str, str]]:
     """Decode history records in a wandb offline ``run-*.wandb`` binary.
 
     Slash-paths arrive as ``nested_key`` (e.g. ``['shard', 'bytes']``); the
     rejoiner reconstructs the keys callers passed to ``log_metrics`` so the
     caller's assertions read like the production payload.
 
+    The offline writer flushes history asynchronously, so a single scan can
+    race ahead of the records (a conda-only 0-rows flake). Pass ``until`` to
+    re-scan until the predicate holds; on timeout the last scan is returned so
+    the caller's own assertion reports the shortfall.
+
     :param wandb_binary: Path to the offline ``run-*.wandb`` file.
+    :param until: Predicate over the decoded rows; when omitted the binary is
+        scanned exactly once (no polling).
+    :param timeout_s: Upper bound on flush polling; ignored when ``until`` is
+        ``None``.
     :returns: One dict per history record; values are JSON-encoded strings
         (as the datastore stores them).
     """
+    if until is None:
+        return _scan_history_rows(wandb_binary)
+    deadline = time.monotonic() + timeout_s
+    while True:
+        rows = _scan_history_rows(wandb_binary)
+        if until(rows) or time.monotonic() >= deadline:
+            return rows
+        time.sleep(_FLUSH_POLL_S)
+
+
+def _scan_history_rows(wandb_binary: Path) -> list[dict[str, str]]:
+    """Single-pass decode of the datastore binary's history records.
+
+    :param wandb_binary: Path to the offline ``run-*.wandb`` file.
+    :returns: One dict per history record found in this scan.
+    """
     ds = wandb_datastore.DataStore()
     ds.open_for_scan(str(wandb_binary))
-    rows: list[dict[str, str]] = []
-    while True:
-        data = ds.scan_data()
-        if data is None:
-            break
-        rec = wandb_pb.Record()  # pyright: ignore[reportAttributeAccessIssue]
-        rec.ParseFromString(data)
-        if not rec.HasField("history"):
-            continue
-        row: dict[str, str] = {}
-        for item in rec.history.item:
-            key = item.key if item.key else "/".join(item.nested_key)
-            row[key] = item.value_json
-        rows.append(row)
-    return rows
+    # ``open_for_scan`` opens a file handle; close it each pass so the polling
+    # loop in ``read_history_rows`` can't leak one handle per re-scan.
+    try:
+        rows: list[dict[str, str]] = []
+        while True:
+            data = ds.scan_data()
+            if data is None:
+                break
+            rec = wandb_pb.Record()  # pyright: ignore[reportAttributeAccessIssue]
+            rec.ParseFromString(data)
+            if not rec.HasField("history"):
+                continue
+            row: dict[str, str] = {}
+            for item in rec.history.item:
+                key = item.key if item.key else "/".join(item.nested_key)
+                row[key] = item.value_json
+            rows.append(row)
+        return rows
+    finally:
+        ds.close()
--- a/tests/integration/test_generate_dataset_cli_wandb_e2e.py
+++ b/tests/integration/test_generate_dataset_cli_wandb_e2e.py
@@ -250,7 +250,13 @@ def test_phase_1_and_2_cli_emits_run_id_plus_shard_and_summary_history(
     assert len(wandb_binaries) == 1, (
         f"expected exactly one .wandb binary under {run_dir}; got {wandb_binaries}"
     )
-    rows = read_history_rows(wandb_binaries[0])
+    rows = read_history_rows(
+        wandb_binaries[0],
+        until=lambda scanned: (
+            sum("shard/bytes" in r for r in scanned) >= _SMOKE_NUM_SHARDS
+            and any("shards/rendered" in r for r in scanned)
+        ),
+    )
 
     shard_rows = [r for r in rows if "shard/bytes" in r]
     assert len(shard_rows) == _SMOKE_NUM_SHARDS, (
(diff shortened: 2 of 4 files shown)


ktinubu:
## Verification Results

Verified on branch `fix/wandb-offline-flush-race` @ `0596d157`. The promise: `read_history_rows` waits out the async offline-writer flush instead of single-shot reading, **without masking a genuine shortfall**, and the new polling loop leaks no file handles.

- [x] **The flush-race contract holds: poll-until-materialized, return-last-on-deadline, single-shot when no predicate** *(Level 2 — invokes the helper through its public API with a stubbed scan + fake clock; this is the deterministic simulation of the conda flush lag that CI cannot reproduce on Linux)*
  ```bash
  $ uv run pytest tests/test_wandb_offline_helper.py -p no:randomly -v
  test_until_predicate_retries_until_rows_materialize PASSED   [ 33%]
  test_until_predicate_polls_to_deadline_then_returns_last_scan PASSED [ 66%]
  test_without_predicate_reads_once PASSED                     [100%]
  3 passed in 1.07s
  ```
  Retry test asserts `scan_count == 3` and `sleep_count == 2`; deadline test drives a fake monotonic clock so the loop exits on the timeout branch after multiple polls (`scan_count > 1`); single-shot test fails if `sleep` is ever called.

- [x] **The originally-failing test passes end-to-end (generate → real offline WandbLogger → decoder)** *(Level 1 — exercises the full offline path the conda job flaked on; this env matches the CI jobs that passed, so it cannot reproduce the race, but it confirms no regression on the real system)*
  ```bash
  $ uv run pytest "tests/test_generate_dataset_wandb.py::test_generate_logs_per_shard_and_summary_metrics_offline" -p no:randomly -v
  test_generate_logs_per_shard_and_summary_metrics_offline PASSED [100%]
  1 passed in 2.05s
  ```

- [x] **The polling loop leaks no file handles (adversarial — review flagged that `open_for_scan` opens an fd and the loop now scans repeatedly)** *(Level 2 — produces a real offline `.wandb` binary, then forces ~10 re-scans via an always-false predicate and measures open fds before/after)*
  ```bash
  $ PYTHONPATH=. uv run python /tmp/fdleak_check.py
  forced multi-scan polling; open fds before=8 after=8
  PASS: polling loop leaked no file handles
  ```
  Stable at 8 across ~10 forced scans. Without the added `ds.close()` in `_scan_history_rows`, this would climb to ~18.

### Not separately checked
- `make test` / pre-commit — CI handles those (pre-commit was clean locally: ruff, pyright, pydoclint, docformatter, interrogate, codespell).
- 100× xdist loop on the integration test: 100/100 both before and after the change locally — informative for no-regression but **cannot distinguish before/after** because this Linux/non-conda env doesn't trigger the async-flush race. The deterministic proof is Check A above.

ktinubu:
## Verification Results

**Before/after flake-rate comparison** for `tests/test_generate_dataset_wandb.py::test_generate_logs_per_shard_and_summary_metrics_offline`, run from two isolated worktrees + venvs:

- **before** = `main` @ `81ce972a` (no fix)
- **after** = `fix/wandb-offline-flush-race` @ `0596d157`
- **env** = Linux, non-conda, 32 cores (matches the CI jobs that *passed* — not the conda job that flaked)

______________________________________________________________________

### TL;DR

The wild flake-rate comparison in this environment is **0/300 → 0/300** — it does **not** demonstrate the fix, because this Linux/non-conda env cannot reproduce the conda-specific async-flush race (consistent with the PR's own "100/100 before and after" note). The fix is instead demonstrated **deterministically** by simulating the flush lag against each branch's real decode path: **before → 0 rows (the flake), after → records recovered via the `until=` retry.**

______________________________________________________________________

- [x] **PR branch is not flaky under repeated runs** *(Level 1 — runs the exact failing test end-to-end, 300×)*

  ```bash
  $ uv run pytest "tests/test_generate_dataset_wandb.py::test_generate_logs_per_shard_and_summary_metrics_offline" \
      --count=300 -n 16 -p no:randomly -q
  ============================= 300 passed in 23.36s =============================
  ```

  300/300 on the PR branch.

- [ ] **The flake reproduces on `main` in this environment** *(Level 1 — SKIP: conda-specific race; this Linux/non-conda env cannot reproduce it)*

  ```bash
  $ uv run pytest "tests/test_generate_dataset_wandb.py::test_generate_logs_per_shard_and_summary_metrics_offline" \
      --count=300 -n 16 -p no:randomly -q   # on main @ 81ce972a
  ============================= 300 passed in 23.61s =============================
  ```

  300/300 on `main` too — i.e. **0/300 before, 0/300 after**. A raw repeat-run here proves nothing about the fix because the failure is a slow-writer (conda-resolved) flush race. To reproduce the real flake you'd need the conda CI job ([run 26850084587](https://github.com/tinaudio/synth-setter/actions/runs/26850084587/job/79179729041)). The deterministic check below stands in for it.

- [x] **Mechanism before/after under a simulated flush lag** *(Level 2 — injects a fake `DataStore` whose history records only flush on the 3rd scan, then drives each branch's real `read_history_rows` decode path; the live env can't reproduce the race so this simulates it directly)*

  ```bash
  ########## BEFORE (main @ 81ce972a) ##########
  read_history_rows signature: (wandb_binary: 'Path') -> 'list[dict[str, str]]'
  supports until= retry param: False
  single-shot read under flush lag -> 0 shard row(s)

  ########## AFTER (PR branch @ 0596d157) ##########
  read_history_rows signature: (wandb_binary: 'Path', *, until: '...' = None, timeout_s: 'float' = 10.0) -> 'list[dict[str, str]]'
  supports until= retry param: True
  single-shot read under flush lag -> 0 shard row(s)
  until= retry read under flush lag -> 1 shard row(s)
  ```

  This is the real before/after:

  - **before**: a single-shot read landing mid-flush returns **0 shard rows** — exactly the `expected 2 per-shard history rows, got 0` failure.
  - **after**: the default single-shot path is **unchanged** (still 0 — no behavior change when `until` is omitted), and the `until=` retry **recovers the record once flushed (1 row)**.

- [x] **Committed deterministic proof passes on the branch** *(Level 2 — runs the new unit tests that pin the retry / deadline-timeout / single-shot contract)*

  ```bash
  $ uv run pytest tests/test_wandb_offline_helper.py -p no:randomly -q
  tests/test_wandb_offline_helper.py::test_until_predicate_retries_until_rows_materialize PASSED
  tests/test_wandb_offline_helper.py::test_until_predicate_polls_to_deadline_then_returns_last_scan PASSED
  tests/test_wandb_offline_helper.py::test_without_predicate_reads_once PASSED
  ============================== 3 passed in 1.06s ===============================
  ```

  This file is **absent on `main`** (`git cat-file -e origin/main:tests/test_wandb_offline_helper.py` → not found) — it's the after-side proof added by this PR.

______________________________________________________________________

### Verdict

The fix behaves as advertised: under a simulated async-flush lag, the unpatched decoder returns 0 rows (the flake) while the patched decoder polls until the records materialize, with the no-`until` default path byte-for-byte unchanged. The raw 0/300 → 0/300 repeat-run is reported for completeness but is **not** evidence either way — the conda-only race is out of reach in this environment, and the deterministic simulation + committed unit tests are the load-bearing proof.

Issue #1139, opened 2026-05-19
Task: split [project.dependencies] into runtime / dev / ci-lite groups so CI doesn't need --no-deps hacks


ktinubu:
### Parent Issue

#603

### Domain Label

ci-automation

### Milestone

data-pipeline v1.0.0

### Problem

Lightweight CI jobs (chiefly `Validate spec structure`, which runs
`python -m synth_setter.pipeline.ci.validate_spec` on the runner host) must
import `synth_setter.pipeline.*` **without** installing the full heavy runtime
(`torch`, `skypilot`, `librosa`, `h5py`, `dask`, `pedalboard`, …).

Because the heavy runtime lives in `[project.dependencies]`, installing the
project always drags it in — so today these jobs use a fragile hack
(`validate-dataset-shards.yaml`, `spec-materialization.yml`,
`test-spec-materialization.yml`):

```bash
pip install --no-deps -e .
pip install "pydantic>=2,<3" python-dotenv   # hand-picked, in three places
```

The hand-picked list is a guess at the import closure of `validate_spec` +
`r2_io`. **Any new top-level import on that path breaks the jobs silently** until
CI fails — e.g. #1120 (`dotenv`) and #1390 (`omegaconf`, unblocked by the
interim lazy-import fix in #1395).

### Solution

Move the heavy runtime out of `[project.dependencies]` into PEP 735
`[dependency-groups]`, leaving the base as just the lite import closure
(`pydantic`, `python-dotenv`). Aggregate the heavy groups under a `runtime`
meta-group, fold it into `dev`, and set `default-groups = ["dev"]` so a bare
`uv sync` still installs everything.

Lite jobs then run `uv sync --frozen --no-default-groups` — project importable,
lite base only, **no `--no-deps`, no hand-picked list, single source of truth**.
The full-runtime install sites already pass `--group dev` (⊇ `runtime`) and are
unchanged. The `cpu`/`cu128` torch backend extras stay; their `[tool.uv.sources]`
index routing is preserved across the move.

Full rationale, dependency partition, torch-routing migration, alternatives, and
risks: **[docs/design/ci-lite-dependency-groups.md](https://github.com/tinaudio/synth-setter/blob/main/docs/design/ci-lite-dependency-groups.md)**.

### Acceptance criteria

- [ ] `[project.dependencies]` trimmed to the lite closure; heavy runtime moved
      into `[dependency-groups]` with a `runtime` meta-group; `default-groups =
[...]

PR #1395, opened 2026-06-03
ci(pipeline): fix. defer omegaconf import so the validate_spec minimal-env CI job imports cleanly

ktinubu:
## Problem

The scheduled *Test Dataset Generation* run failed in the `Validate spec structure` job (`validate-dataset-shards.yaml`) with:

```
ModuleNotFoundError: No module named 'omegaconf'
  File ".../pipeline/ci/validate_spec.py", line 14, in <module>
    from synth_setter.pipeline.schemas.spec import (...)
  File ".../pipeline/schemas/spec.py", line 20, in <module>
    from omegaconf import DictConfig, OmegaConf
```

That job deliberately installs a minimal env (`pip install --no-deps -e .` + `pydantic` + `python-dotenv`, no omegaconf) so the structural check stays fast. #1390 added a **module-level** `from omegaconf import DictConfig, OmegaConf` to `spec.py` for the new `DatasetSpec.from_hydra_cfg`, which runs on every import of the module — including `validate_spec`'s import chain — breaking the minimal-env contract.

## Fix

- Move `OmegaConf` to a **lazy import inside `from_hydra_cfg`** (its only runtime user).
- Move `DictConfig` to a `TYPE_CHECKING` block — the `cfg: DictConfig` annotation is already a string under `from __future__ import annotations`, so no runtime symbol is needed.
- Add `test_bare_spec_import_does_not_pull_omegaconf`, a subprocess import-purity test mirroring the existing launcher-pure checks, so a future re-promotion of the import fails locally before CI breaks.

This is the interim unblock; the long-term `pipeline-ci` dependency group that removes the `--no-deps` hack class is tracked separately and being implemented in parallel.

Refs #1139

## Test plan

- `uv run pytest tests/pipeline/schemas/test_dataset_spec.py tests/pipeline/ci_validate/test_validate_spec.py` — 104 passed.
- New test confirmed RED before the fix, GREEN after.
- `make format` clean (ruff, pyright, pydoclint, …).
- Pre-PR `/repo-review-full-no-comments`: 0 BLOCK, 1 stylistic WARN across 7 skills.

### DIFF
--- a/src/synth_setter/pipeline/schemas/spec.py
+++ b/src/synth_setter/pipeline/schemas/spec.py
@@ -15,9 +15,8 @@ import subprocess
 import sys
 from datetime import datetime, timedelta, timezone
 from functools import cached_property
-from typing import Any, Literal
+from typing import TYPE_CHECKING, Any, Literal
 
-from omegaconf import DictConfig, OmegaConf
 from pydantic import (
     BaseModel,
     ConfigDict,
@@ -34,6 +33,9 @@ from synth_setter.pipeline.schemas.prefix import (
     make_r2_prefix,
 )
 from synth_setter.pipeline.schemas.r2_location import R2Location
+
+if TYPE_CHECKING:
+    from omegaconf import DictConfig
 
 __all__ = [
     "EXTENSION_TO_OUTPUT_FORMAT",
@@ -608,6 +610,10 @@ class DatasetSpec(BaseModel):
             which ``masked_copy`` rejects with ``ValueError``) or the masked cfg
             did not resolve to a mapping — both normalized to one stable type.
         """
+        # Lazy import: omegaconf is absent from the minimal-env CI install that
+        # runs `validate_spec`, which imports this module but never calls this.
+        from omegaconf import OmegaConf
+
         spec_keys = [k for k in cfg if isinstance(k, str) and k in cls.model_fields]
         try:
             masked = OmegaConf.masked_copy(cfg, spec_keys)
--- a/tests/pipeline/schemas/test_dataset_spec.py
+++ b/tests/pipeline/schemas/test_dataset_spec.py
@@ -816,6 +816,33 @@ class TestSpecImportStaysLauncherPure:
             f"stdout={result.stdout}\nstderr={result.stderr}"
         )
 
+    def test_bare_spec_import_does_not_pull_omegaconf(self) -> None:
+        """Bare spec import must not load ``omegaconf`` at module level.
+
+        ``omegaconf`` is only needed by ``DatasetSpec.from_hydra_cfg`` and runs
+        lazily inside it. The CI ``validate_spec`` job installs a minimal env
+        (pydantic + python-dotenv, no omegaconf); promoting the import back to
+        module level fails this test before that job breaks at runtime.
+        """
+        script = (
+            "import sys\n"
+            "import synth_setter.pipeline.schemas.spec  # noqa: F401\n"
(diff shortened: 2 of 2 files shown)


ktinubu:
## Verification Results

**Promise under test:** the `validate_spec` CI step runs in a deliberately minimal env (`pip install --no-deps -e .` + `pydantic` + `python-dotenv`, **no omegaconf**); after this change the `synth_setter.pipeline.ci.validate_spec` import chain must load and run without `ModuleNotFoundError: No module named 'omegaconf'`.

I reconstructed that exact env (`uv venv` + `--no-deps -e .` + `pydantic>=2,<3` + `python-dotenv`) and confirmed `omegaconf`, `torch`, and `hydra` are absent — matching the CI install.

- [x] **`validate_spec` imports + runs in the minimal env (the failing job's contract)** *(Level 1 — reproduces the exact CI job env and runs the real `python -m` entrypoint end-to-end)*
  ```bash
  $ /tmp/minenv/bin/python -c "import importlib.util as u; print('omegaconf:', u.find_spec('omegaconf') is not None, '| torch:', u.find_spec('torch') is not None, '| hydra:', u.find_spec('hydra') is not None, '| pydantic:', u.find_spec('pydantic') is not None, '| dotenv:', u.find_spec('dotenv') is not None)"
  omegaconf: False | torch: False | hydra: False | pydantic: True | dotenv: True

  $ /tmp/minenv/bin/python -c "import synth_setter.pipeline.ci.validate_spec as m; print('validate_spec import OK; main callable:', callable(m.main))"
  validate_spec import OK; main callable: True

  $ /tmp/minenv/bin/python -m synth_setter.pipeline.ci.validate_spec /tmp/does-not-exist.json
  ...
  FileNotFoundError: [Errno 2] No such file or directory: '/tmp/does-not-exist.json'
  # exit=1 — fails on the missing spec file, i.e. it ran PAST the import that previously broke CI
  ```

- [x] **Negative control: the pre-fix eager import still breaks the same env** *(Level 1 — temporarily re-injected the module-level `from omegaconf import ...`, ran the same import, then restored the file; `git status` clean after)*
  ```bash
  # eager import re-injected into spec.py, then:
  $ /tmp/minenv/bin/python -c "import synth_setter.pipeline.ci.validate_spec"
    File ".../pipeline/schemas/spec.py", line 20, in <module>
      from omegaconf import DictConfig, OmegaConf  # BASELINE-SIM
  ModuleNotFoundError: No module named 'omegaconf'
  # file restored; git status --porcelain → clean
  ```
  Before → after in one env: **fails** with the CI error pre-fix, **imports clean** post-fix.

- [x] **`from_hydra_cfg` still works when omegaconf IS available (lazy import resolves)** *(Level 2 — runs the behavioral suite that exercises the lazy import's only caller; promise is method behavior, not end-to-end)*
  ```bash
  $ uv run pytest tests/pipeline/schemas/test_dataset_spec.py::TestFromHydraCfg -q
  ...
  4 passed
  ```

- [x] **Regression guard fails before the fix, passes after** *(Level 2 — the new subprocess import-purity test; confirmed RED on the eager import, GREEN on the lazy form)*
  ```bash
  $ uv run pytest tests/pipeline/schemas/test_dataset_spec.py -k omegaconf -q
  1 passed
  ```

All checks executed and observed. The lazy import keeps the minimal-env CI contract intact while preserving `from_hydra_cfg` behavior; a future re-promotion of the import now fails the new test locally before reaching CI.

PR #1397, opened 2026-06-03
fix(eval): inline oracle-eval fetches whole dataset root so VDS split shards resolve

### LINKED ISSUE #1396: fix(eval): inline oracle-eval downloads only splits, not the shards their VDS reference
## Summary

The inline oracle-eval (`generate_dataset.py`, added in #1338) crashes during `mode=predict` reading the `audio` dataset:

```
OSError: Can't synchronously read data (insufficient elements in destination selection)
  at surge_datamodule.py:106  ds[start_idx:end_idx]   # the "audio" dataset
```

## Root cause

The finalized `{train,val,test}.h5` splits are **HDF5 virtual datasets** (`reshard.py`) that reference the source `shard-*.h5` files **by relative basename**, so they only resolve when the shards are co-located. `_download_finalized_splits` downloaded only the three split files + `stats.npz` into a fresh `oracle_eval/<run_id>/` directory — **without the shards**. The dangling virtual mappings read back fill-value zeros for the first row and raise `insufficient elements` for the rest.

Confirmed empirically: the run-dir `test.h5` (shards beside it) reads real audio; the relocated `oracle_eval/` copy fails on `audio[1:2]`. Not a datamodule bug — the data simply isn't co-located.

## Fix

Download the **whole dataset-root prefix** (shards + splits + stats) via `r2_io.download_dir_no_overwrite`, so the virtual mappings resolve. Adds `R2Location.dataset_root_uri()` for the prefix-level URI.

Refs #1338, #1331.

ktinubu:
## Problem

The inline oracle-eval (added in #1338) crashes during `mode=predict` reading the `audio` dataset:

```
OSError: Can't synchronously read data (insufficient elements in destination selection)
  surge_datamodule.py:106  ds[start_idx:end_idx]   # the "audio" dataset
```

It surfaces when running the `generate_dataset_cadence` sweep with `oracle_eval_inline=true`.

## Root cause

The finalized `{train,val,test}.h5` splits are **HDF5 virtual datasets** (`reshard.py`) that reference the source `shard-*.h5` files **by relative basename** — they hold no data themselves (`storage_size=0`) and only resolve when the shards are co-located. `_download_finalized_splits` downloaded only the three split files + `stats.npz` into a fresh `oracle_eval/<run_id>/` directory, **without the shards**. The dangling virtual mappings read back fill-value zeros for the first row and raise `insufficient elements` for the rest.

Verified empirically: the run-dir `test.h5` (shards beside it) reads real audio; the relocated `oracle_eval/` copy fails on `audio[1:2]`. The datamodule reads correctly — the data simply wasn't co-located.

## Fix

Download the **whole dataset-root prefix** (shards + splits + stats) via `r2_io.download_dir_no_overwrite`, so the virtual mappings resolve. Adds `R2Location.dataset_root_uri()` for the prefix-level URI and renames the helper `_download_finalized_splits` → `_download_dataset_root` to match its new behavior.

## Tests

- `test_dataset_root_uri` pins the bare-prefix URI (trailing slash preserved).
- `test_download_dataset_root_fetches_shards_beside_splits` — state-based via the real-rclone `fake_r2_remote` fixture; asserts the shards land beside the splits. Verified to FAIL against the old splits-only implementation and PASS against the fix. `stats.npz` coverage preserved.
- `make format` clean (ruff, pyright, pydoclint, …); 182 tests pass across the touched suites.

Fixes #1396

### DIFF
--- a/src/synth_setter/cli/generate_dataset.py
+++ b/src/synth_setter/cli/generate_dataset.py
@@ -32,7 +32,7 @@ from omegaconf import DictConfig, OmegaConf
 from synth_setter.cli.finalize_dataset import finalize_from_spec
 from synth_setter.data.vst.core import extract_renderer_version
 from synth_setter.pipeline import r2_io
-from synth_setter.pipeline.constants import STATS_NPZ_FILENAME, WORKER_SPEC_URI_ENV
+from synth_setter.pipeline.constants import WORKER_SPEC_URI_ENV
 from synth_setter.pipeline.partitioning import (
     available_cpus,
     get_my_shards,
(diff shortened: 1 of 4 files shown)


ktinubu:
## Verification Results

Verified the fix's promise behaviorally — that the inline oracle-eval download brings the shards along so a finalized **virtual-dataset** split resolves. Reproduced the exact production failure mode (`audio[1:2]`) as a black-box check against `_download_dataset_root(spec, dest)` exercised through the **real rclone binary** (`RCLONE_CONFIG_R2_TYPE=local`).

- [x] **Whole-root download → VDS split resolves; the slice that crashed in prod reads real data** *(Level 1 — end-to-end: real rclone copy + real HDF5 virtual dataset + the failing read)*
  Built a real VDS `test.h5` whose `audio` is virtual and references `shard-000000.h5` **by basename** (as `reshard.py` does), staged the shard + split + `stats.npz` under the dataset-root prefix, ran the actual helper, then read `audio[1:2]`:
  ```text
  dataset_root_uri(): r2://intermediate-data/data/test-dataset/test-dataset-20260328T120000000Z/
  rclone copy ... r2:.../test-dataset-...Z/  ->  oracle_eval/
    Copied (new): stats.npz, test.h5, shard-000000.h5
  downloaded into dest: ['shard-000000.h5', 'stats.npz', 'test.h5']
  audio[1:2] shape: (1, 2, 10) sum: 610.0
  CHECK 1 PASS: whole-root download → VDS resolves, audio[1:2] reads real data
  ```
  The shard lands beside the split and the virtual mapping resolves to the real data (sum 610).

- [x] **Falsifiable: without the shards the split reads WRONG data** *(Level 1 — same read on a splits-only directory)*
  Copied only `test.h5` (no shard) into a fresh dir and read the same slice:
  ```text
  CHECK 2b: splits-only audio[1:2] sum = 0.0  (real shard sum would be 610)
  ```
  Confirms the check is falsifiable and pins the bug: a splits-only download leaves the VDS dangling (fill-value zeros here; `OSError: insufficient elements` on the real compressed production shard). Either way the eval is wrong without the shards.

- [x] **Helper creates the destination dir itself (no manual `mkdir`)** *(Level 1 — adversarial; the removed `dest.mkdir` is now rclone's job)*
  `oracle_eval/` did not pre-exist; rclone created it during the copy (`CHECK 3: dest dir auto-created`).

- [x] **`dataset_root_uri()` and the download path are unit/integration covered** *(Level 2 — ran the suites, not CI noise)*
  `test_dataset_root_uri`, `test_download_dataset_root_fetches_shards_beside_splits` (real-rclone state test), and the dispatch tests pass — 182 passed across the touched suites. The download test was independently confirmed to FAIL against the old splits-only implementation.

Script: `/tmp/verify_pr1397.py` (run via `uv run python`).

PR #1399, opened 2026-06-03
docs: design CI-lite dependency-group split (#1139)

ktinubu:
## What

Adds `docs/design/ci-lite-dependency-groups.md` — the design for the long-term #1139 fix: stop the fragile `--no-deps` + hand-picked install in the lightweight CI jobs.

## Approach

Move the heavy runtime out of `[project.dependencies]` into PEP 735 `[dependency-groups]`, keeping the base at the **union lite closure** (`pydantic` + `python-dotenv` + `pyyaml`). Aggregate heavy deps under a `runtime` meta-group folded into `dev`, and set `default-groups = ["dev"]` so a bare `uv sync` is unchanged. Because `pip`/`uv pip` never install PEP 735 groups, the 5 lite sites simply **drop `--no-deps`** — no hand-picked list, single source of truth.

The doc covers: all five lite sites and their three closures, the torch `cpu`/`cu128` index-routing migration (extras stay; all three source blocks move together; backend flag is load-bearing), the per-callsite migration plan, the drift-guard test, and why extras / `--only-group` don't work here.

## Context

- Replaces the original (infeasible) issue AC — PEP 621 extras are additive, so `-e .[pipeline-ci]` can't subset the heavy base.
- Interim unblock for today's omegaconf break already shipped in #1395 (lazy import); independent of this.
- Issue #1139 has been rewritten to just the problem + this solution.

Implementation lands as a separate PR once the design is approved.

Refs #1139

### DIFF
--- a/docs/design/ci-lite-dependency-groups.md
+++ b/docs/design/ci-lite-dependency-groups.md
@@ -1,0 +1,241 @@
+# Design Doc: CI-Lite Dependency Groups
+
+> **Status**: Draft
+> **Author**: ktinubu@
+> **Last Updated**: 2026-06-03
+> **Tracking**: #1139
+
+______________________________________________________________________
+
+### Index
+
+| §   | Section                                               | What it covers                                            |
+| --- | ----------------------------------------------------- | --------------------------------------------------------- |
+| 1   | [Context & Problem](#1-context--problem)              | Why lightweight CI jobs break, and why the hack is forced |
+| 2   | [Goals & Non-Goals](#2-goals--non-goals)              | What this change must and must not do                     |
+| 3   | [Solution](#3-solution)                               | Light base + PEP 735 groups + `default-groups`            |
+| 4   | [Torch Index Routing](#4-torch-index-routing)         | Migrating the `cpu`/`cu128` routing off `[project]`       |
+| 5   | [Migration Plan](#5-migration-plan)                   | Per-callsite changes and ordering                         |
+| 6   | [Drift Guard](#6-drift-guard)                         | The test that keeps the base honest                       |
+| 7   | [Alternatives Considered](#7-alternatives-considered) | Why extras and `--only-group` don't work here             |
+| 8   | [Risks & Rollout](#8-risks--rollout)                  | Failure modes and validation                              |
+
+______________________________________________________________________
+
+## 1. Context & Problem
+
+A handful of CI jobs import `synth_setter.pipeline.*` to do cheap structural
+work — chiefly `Validate spec structure`, which runs
+`python -m synth_setter.pipeline.ci.validate_spec` on the runner host (outside
+the Docker image). These jobs must **not** install the full runtime
+(`torch`, `skypilot`, `librosa`, `h5py`, `dask`, `pedalboard`, …) — that is
+minutes of wheel downloads for a job that only needs to parse a Pydantic model.
+
+The current workaround (three workflows: `validate-dataset-shards.yaml`,
+`spec-materialization.yml`, `test-spec-materialization.yml`):
+
+```bash
+pip install --no-deps -e .
+pip install "pydantic>=2,<3" python-dotenv   # hand-picked, by hand
+```
+
+This is fragile. The hand-picked list is a guess at the import closure of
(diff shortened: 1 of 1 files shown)


ktinubu:
## Verification Results

This is a **docs-only** PR (a design doc); there is no runtime behavior to exercise. Verification covers (1) the doc integrates into the published site, and (2) the design's load-bearing technical claims are actually true — if a claim is wrong, the design is wrong.

- [x] **Doc builds into the site with no broken links/nav** *(Level 1 — runs the exact docs CI gate `mkdocs build --strict`; for a doc this is the end-to-end integration)*
  ```bash
  $ uv run mkdocs build --strict 2>&1 | tail -4
  INFO    -  The following pages exist in the docs directory, but are not included in the "nav" configuration:
    - pr-readiness-loop.md
    - testing/mutmut.md
  INFO    -  Documentation built in 0.73 seconds
  # exit 0; the only "not in nav" warnings are pre-existing files, NOT the new design doc
  # (design/ auto-nav picked it up) — strict mode would have failed otherwise.
  ```

- [x] **The union lite base {pydantic, python-dotenv, pyyaml} satisfies all 3 lite entrypoints, with zero heavy deps** *(Level 2 — runs the actual imports the lite CI jobs run, in a clean env that mirrors the proposed `pip install -e .` on a light base)*
  ```bash
  $ uv venv /tmp/vunion
  $ VIRTUAL_ENV=/tmp/vunion uv pip install --no-deps -e .        # project editable
  $ VIRTUAL_ENV=/tmp/vunion uv pip install pydantic python-dotenv pyyaml   # union base + transitives
  $ /tmp/vunion/bin/python -c "
  from synth_setter.pipeline.ci.validate_spec import main
  from synth_setter.pipeline.r2_io import ensure_r2_env_loaded
  from synth_setter.pipeline.ci.load_image_config import main as lic_main
  print('ALL THREE LITE ENTRYPOINTS IMPORT OK on union base')"
  ALL THREE LITE ENTRYPOINTS IMPORT OK on union base
  $ /tmp/vunion/bin/python -c "
  import importlib.util as u
  print('heavy present:', [m for m in ('torch','hydra','skypilot','librosa','h5py','omegaconf') if u.find_spec(m)] or 'NONE')"
  heavy present: NONE
  ```
  Confirms §2's closure table and that the lite env stays free of torch/hydra/skypilot/librosa/omegaconf.

- [x] **`default-groups=["dev"]` + heavy-in-group means bare `uv sync` is unchanged, while `uv sync --no-default-groups` yields project + light base only** *(Level 2 — runs `uv` against a scratch project with the exact proposed shape)*
  ```bash
  # scratch pyproject: dependencies=["python-dotenv"]; [dependency-groups] heavy=["numpy"],
  # runtime=[{include-group=heavy}], dev=[{include-group=runtime}]; [tool.uv] default-groups=["dev"]
  $ uv sync                      # => project + python-dotenv + numpy (everything)
  $ python -c "import uvtest2, numpy"        # OK
  $ uv sync --no-default-groups  # => uninstalls numpy
   - numpy==2.2.6
  $ python -c "import uvtest2"    # OK (project still importable)
  $ python -c "import numpy"      # ModuleNotFoundError: No module named 'numpy'
  ```
  Confirms §4's core mechanic and the install matrix.

- [x] **`--only-group` omits the project (why it can't be used here), and group-conditioned `[tool.uv.sources]` routing works (§5 torch migration)** *(Level 2 — runs `uv` directly)*
  - `uv sync --only-group <g>` installed only the group's deps; `import <project>` → `ModuleNotFoundError` (project omitted). Documented in §8 alternatives.
  - `[tool.uv.sources]` with `{ index = "...", group = "..." }` resolved and routed the dep to the named index on uv 0.11.2 (lockfile `source = { registry = "test.pypi.org" }`). Confirms torch routing can follow torch into the `torch` group.

- [ ] **Full implementation behavior (lockfile resolves identically across Linux-cpu/cu128/macOS; the 5 lite workflows go green on the light base)** *(SKIP — out of scope for this design-doc PR; this is the acceptance gate for the implementation PR per §6/§9. To verify then: regenerate `uv.lock`, diff the resolved set per platform, and confirm the existing `test`, `cpu-slow`, `test-mps`, `docker-build-validation` jobs pass.)*

Issue #353, opened 2026-03-28
fix(ci): pipeline workflow consistency fixes


ktinubu:
## Context

During the dataset generation workflow PR (#344), a comparison of `docker-build-validation.yml` and `dataset-generation.yml` found 4 consistency issues across the two pipeline workflows.

## Tasks

- [ ] **1. Add R2 credential pre-check to dataset-generation workflow**
  The image workflow has a "Verify Docker Hub push access" step (lines 76-91 of `docker-build-validation.yml`) that checks token scope before attempting a push. The dataset workflow has no equivalent — if R2 credentials are bad, the workflow runs the full generation (potentially hours), then fails at the upload step. Add a pre-flight `rclone lsd r2:intermediate-data` check.

- [ ] **2. Add `set -o pipefail` to image workflow smoke tests**
  Lines 133 and 141 of `docker-build-validation.yml` run `docker run` commands with piped output but no `set -o pipefail`. Same bug we fixed in the dataset workflow (e53994f). If the docker run fails but the pipe succeeds, the step reports success.

- [ ] **3. Add artifact capture to image workflow smoke tests**
  The dataset workflow uploads a manifest artifact (config + manifest JSON + logs). The image workflow runs smoke tests but saves no output as artifacts. If a smoke test fails intermittently, there's no downloadable log to inspect after the Actions UI log expires.

- [ ] **4. Standardize Pydantic config style in `pipeline/schemas/`**
  `ImageConfig` uses old Pydantic syntax (`class ImageConfig(BaseModel, strict=True)`) while `DatasetConfig` uses the Pydantic v2 `ConfigDict` style. Both now live in `pipeline/schemas/` — should be consistent. Standardize on `ConfigDict`.

## Acceptance Criteria

- All 4 items addressed in a single PR
- Both workflow files follow the same patterns for credential checks, error handling, and artifact capture
- Pydantic models in `pipeline/schemas/` use consistent style

Refs #344

PR #1412, opened 2026-06-04
ci(pipeline): emit full Hydra tracebacks from datagen launcher CI workflows

ktinubu:
## Summary

The datagen launcher entrypoint `synth-setter-generate-dataset` (`src/synth_setter/cli/generate_dataset.py`) is wrapped in `@hydra.main`. When the launcher fails (spec materialize, R2 upload, or SkyPilot dispatch), Hydra's decorator suppresses the traceback and prints only a one-line summary unless `HYDRA_FULL_ERROR=1` is in the environment. Several CI workflows that drive this entrypoint did not set it, so launcher-side failures surfaced as truncated, hard-to-debug errors.

## Changes

Export `HYDRA_FULL_ERROR=1` at every CI call site that invokes the `@hydra.main` entrypoint:

- **`generate-dataset-shards.yaml`** — `gen_local` (host) and `gen_docker` (in-container, forwarded via `-e HYDRA_FULL_ERROR`). This reusable also covers `test-dataset-generation.yml`, which dispatches through it.
- **`test-skypilot-debug.yml`** — `launcher-runner` and `launcher-docker` modes.

## Intentionally out of scope

`spec-materialization.yml`, `test-spec-materialization.yml`, and `validate-dataset-shards.yaml` call `materialize_spec` / `validate_spec` / `validate_shard` — none is a `@hydra.main`. `materialize_spec` uses the Hydra compose API (and already catches `HydraException`); the others aren't Hydra entrypoints. `HYDRA_FULL_ERROR` would be inert there, so they are left untouched.

## Testing

- `pytest tests/infra/test_generate_dataset_shards_hydra_overrides_validation.py` — 36 passed
- `pre-commit run` on both workflows — all hooks pass (incl. actionlint "Validate GitHub Workflows")
- `/repo-review-full-no-comments` — 0 BLOCK across 5 skills (code-health, synth-setter-project-standards, gha-workflow-validator, comment-hygiene, shell-style); 1 non-actionable advisory WARN (per-step env duplication, which cannot be hoisted without leaking into non-Hydra docker steps).

Refs #353

### DIFF
--- a/.github/workflows/generate-dataset-shards.yaml
+++ b/.github/workflows/generate-dataset-shards.yaml
@@ -429,6 +429,7 @@ jobs:
           INPUT_HYDRA_OVERRIDES: ${{ inputs.hydra_overrides }}
           INPUT_RUN_ID: ${{ inputs.run_id || format('{0}-{1}', github.run_id, github.run_attempt) }}
           SYNTH_SETTER_CI_MODE: "1"
+          HYDRA_FULL_ERROR: "1"
         run: |
           set -euo pipefail
           OVERRIDES=(
@@ -553,6 +554,7 @@ jobs:
           LOCAL_OVERRIDE: ${{ inputs.local && 'skypilot_launch.local=true' || '' }}
           HYDRA_OVERRIDES_EXTRA: ${{ inputs.hydra_overrides }}
           INPUT_RUN_ID: ${{ inputs.run_id || format('{0}-{1}', github.run_id, github.run_attempt) }}
+          HYDRA_FULL_ERROR: "1"
         run: |
           set -euo pipefail
           docker run --rm \
@@ -581,6 +583,7 @@ jobs:
             -e LOCAL_OVERRIDE \
             -e HYDRA_OVERRIDES_EXTRA \
             -e INPUT_RUN_ID \
+            -e HYDRA_FULL_ERROR \
             "${IMAGE}" \
             bash -c '
               set -euo pipefail
--- a/.github/workflows/test-skypilot-debug.yml
+++ b/.github/workflows/test-skypilot-debug.yml
@@ -388,6 +388,7 @@ jobs:
           CONFIG_PATH: ${{ matrix.config }}
           TEMPLATE: ${{ matrix.template }}
           LOG_PATH: /tmp/debug-pure-sky-${{ matrix.cluster_suffix }}.log
+          HYDRA_FULL_ERROR: "1"
         run: |
           set -euo pipefail
           uv run synth-setter-generate-dataset \
@@ -408,6 +409,7 @@ jobs:
           CONFIG_PATH: ${{ matrix.config }}
           TEMPLATE: ${{ matrix.template }}
           LOG_PATH: /tmp/debug-pure-sky-${{ matrix.cluster_suffix }}.log
+          HYDRA_FULL_ERROR: "1"
         run: |
           set -euo pipefail
           docker pull "$IMAGE"
@@ -422,6 +424,7 @@ jobs:
             -e RCLONE_CONFIG_R2_SECRET_ACCESS_KEY \
             -e RCLONE_CONFIG_R2_ENDPOINT \
             -e R2_ACCOUNT_ID \
+            -e HYDRA_FULL_ERROR \
             "$IMAGE" \
             bash -c '
               set -euo pipefail


ktinubu:
## Verification Results

The promise this PR makes: **when the `@hydra.main` datagen launcher entrypoint (`synth-setter-generate-dataset`) fails in CI, `HYDRA_FULL_ERROR=1` makes it emit a complete Python traceback instead of Hydra's truncated one-line summary.** Verified against the real entrypoint installed in the worktree venv.

- [x] **The real entrypoint emits a full traceback iff `HYDRA_FULL_ERROR=1`** *(Level 2 — invoked the actual `synth-setter-generate-dataset` `@hydra.main` entrypoint; the promise is tool behavior, not end-to-end. This is exactly the behavior the workflow env change relies on.)*

  Forced a genuine launcher-side failure (missing `Surge XT.vst3` plugin — the same class of failure the PR targets) and compared the two environments. Falsifiable signal: Hydra prints `Set the environment variable HYDRA_FULL_ERROR=1 for a complete stack trace.` only when the var is unset, and includes its internal frames only when it is set.

  ```bash
  $ env -u HYDRA_FULL_ERROR WANDB_MODE=offline uv run synth-setter-generate-dataset \
      experiment=generate_dataset/smoke-shard 2>&1 \
      | grep -c "Set the environment variable HYDRA_FULL_ERROR=1"
  1
  $ HYDRA_FULL_ERROR=1 WANDB_MODE=offline uv run synth-setter-generate-dataset \
      experiment=generate_dataset/smoke-shard 2>&1 \
      | grep -c "Set the environment variable HYDRA_FULL_ERROR=1"
  0
  ```

  With `HYDRA_FULL_ERROR=1` the trace additionally includes the Hydra framework frames (`hydra/main.py:94`, `_internal/utils.py:394/457/223`, `run_and_report`, `core/utils.py:260/186`) down to the app frame `cli/generate_dataset.py:763 → generate() → core.py:55`. Without it, only the app frames show and the hint line is appended.

- [x] **Every entrypoint call site sets the var, and every in-container site forwards it** *(Level 5 — parses the two workflow YAMLs and asserts the env wiring. Justification: the promise here is structural completeness — "no `synth-setter-generate-dataset` call site was missed." No tool exercises a workflow's internal step `env:`/`-e` wiring without triggering the paid workflow, which is the SKIP below.)*

  ```bash
  $ python3 parse_check.py   # parses jobs→steps; finds every step whose run invokes the
                             # entrypoint as a command word; asserts HYDRA_FULL_ERROR=1 in env
                             # and (for docker-run steps) "-e HYDRA_FULL_ERROR" in the body
  [OK] generate-dataset-shards.yaml :: 'Run synth-setter-generate-dataset ... (skypilot-local row)'  env_set=True container=False forwards=False
  [OK] generate-dataset-shards.yaml :: 'Run synth-setter-generate-dataset (runpod / oci row; in container)' env_set=True container=True forwards=True
  [OK] test-skypilot-debug.yml :: 'Run launcher (launcher-runner mode)'  env_set=True container=False forwards=False
  [OK] test-skypilot-debug.yml :: 'Run launcher (launcher-docker mode)'  env_set=True container=True forwards=True
  ALL ENTRYPOINT STEPS COVERED
  ```

  (Host steps need only the `env:` key; the two `docker run` steps additionally forward it via bare `-e HYDRA_FULL_ERROR`, which propagates the runner-shell value into the container.)

- [ ] **In-container `-e HYDRA_FULL_ERROR` forwarding propagates the value into the worker container** *(Level 2/3 — SKIP: the Docker daemon is down in this environment (`docker info` → daemon DOWN), so the container run can't execute.)*
  To verify manually:
  ```bash
  # With a running Docker daemon:
  HYDRA_FULL_ERROR=1 docker run --rm -e HYDRA_FULL_ERROR alpine sh -c 'echo "in-container: $HYDRA_FULL_ERROR"'
  # Expect: in-container: 1
  ```

- [ ] **Full end-to-end workflow run on GitHub** *(Level 1 — SKIP: `generate-dataset-shards.yaml` and `test-skypilot-debug.yml` require R2 credentials and paid RunPod/OCI/SkyPilot compute, which the project forbids spending without explicit approval. Triggering the dispatch would launch real clusters.)*
  To verify manually (costs money — operator decision):
  ```bash
  gh workflow run test-skypilot-debug.yml --ref ci/hydra-full-error-datagen-entrypoints
  gh run watch
  # Then induce a launcher failure and confirm the worker log shows the full traceback,
  # with no "Set the environment variable HYDRA_FULL_ERROR=1" hint line.
  ```

### Summary

The load-bearing behavior (Hydra emits a complete traceback under `HYDRA_FULL_ERROR=1`) is confirmed Level-2 against the real entrypoint, and the workflow wiring is confirmed complete. The two unchecked items require infrastructure not available/authorized here (Docker daemon down; paid CI compute) and are documented for manual run.

PR #1418, opened 2026-06-04
ci: strip inherited packagecloud git-lfs apt source so act-runner image builds

### LINKED ISSUE #1417: fix(ci): act-runner image build fails on packagecloud git-lfs 403
## Summary

The `Verify act setup` workflow (`.github/workflows/test-act.yaml`) fails at the **"Build act-runner-gh image from docs Dockerfile"** step. The build's `apt-get update` returns **403 Forbidden** while fetching the external **packagecloud.io git-lfs** apt repository.

## Root cause

`docs/operations/act-runner.Dockerfile` does not install git-lfs — it only layers the `gh` CLI onto `catthehacker/ubuntu:act-latest`. But the catthehacker base image ships a pre-configured packagecloud.io git-lfs apt source in `/etc/apt/sources.list.d/`. Our `apt-get update` reads *every* configured source, so when packagecloud returns 403 for the git-lfs repo (rate-limited / decommissioned), `apt-get update` exits non-zero and `set -eux` fails the whole build.

The `--no-cache` build in `test-act.yaml` surfaces this on every relevant PR.

## Fix

Strip any packagecloud.io source from the inherited apt config before the first `apt-get update`. git-lfs is already present in the base image and the repo is never needed for the `gh` install, so the build no longer depends on it.

## Acceptance

- [ ] `Verify act setup` workflow's runner-image build step passes (green) with `--no-cache`.

ktinubu:
## What

`docs/operations/act-runner.Dockerfile` only layers the `gh` CLI onto `catthehacker/ubuntu:act-latest`, but that base image ships a pre-configured **packagecloud.io git-lfs** apt source. Our `apt-get update` reads every configured source, so when packagecloud returns **403 Forbidden** for the git-lfs repo, `apt-get update` exits non-zero and `set -eux` fails the whole runner-image build — reddening the `Verify act setup` workflow's `--no-cache` build step.

This strips any `packagecloud.io` source from the inherited apt config before the first `apt-get update`. git-lfs is already installed in the base image and the repo was never needed for the `gh` install, so the build no longer depends on it.

```dockerfile
{ grep -rlZ 'packagecloud\.io' '/etc/apt/sources.list.d/' 2>/dev/null || true; } | xargs -0 -r rm -f;
```

The `|| true` makes the no-match case explicit so correctness doesn't depend on `pipefail` being absent; `-Z`/`xargs -0` is NUL-safe and `xargs -r` no-ops on zero matches.

## Verification

The `Verify act setup` workflow (`.github/workflows/test-act.yaml`) builds this image with `--no-cache` on every relevant PR, so this PR's own CI exercises the fix end-to-end.

## Review

Pre-PR `repo-review-full-no-comments` gate: 0 BLOCK, 4 WARN (all advisory; the shell-style robustness WARNs were applied in the diff). comment-hygiene clean.

Fixes #1417

### DIFF
--- a/docs/operations/act-runner.Dockerfile
+++ b/docs/operations/act-runner.Dockerfile
@@ -3,6 +3,9 @@ FROM catthehacker/ubuntu:act-latest
 ENV DEBIAN_FRONTEND=noninteractive
 
 RUN set -eux; \
+    # catthehacker's base ships a packagecloud.io git-lfs apt source that now 403s;
+    # git-lfs is already installed and we only need gh, so drop it before apt-get update.
+    find /etc/apt/sources.list.d/ -type f -exec grep -q 'packagecloud\.io' {} \; -delete; \
     apt-get update; \
     apt-get install -y --no-install-recommends \
         ca-certificates \


ktinubu:
## Verification Results

**Promise under test:** the act-runner image build no longer fails when the inherited `packagecloud.io` git-lfs apt source returns 403 — i.e. the new step removes that source (and only that source) before `apt-get update`, surviving `set -eux`.

- [x] **Image builds clean with `--no-cache`** *(Level 1 — end-to-end. Verified green by the PR's own `Verify act setup` workflow → "Build act-runner-gh image from docs Dockerfile" step (run 26925900930, 3m29s, conclusion success). Not runnable locally — no Docker daemon — so the CI build is the authoritative end-to-end check.)*
  To run manually:
  ```bash
  DOCKER_BUILDKIT=0 docker build --no-cache -f docs/operations/act-runner.Dockerfile -t act-runner-gh:latest docs/operations
  docker run --rm act-runner-gh:latest gh --version
  ```

- [x] **The step removes only the packagecloud source, keeping other apt sources** *(Level 3 — exercises the exact `find … -exec grep -q … -delete` command against a real `sources.list.d/` fixture; black-box input→output)*
  ```bash
  $ bash --noprofile --norc -c '
  work=$(mktemp -d); d="$work/sources.list.d"; mkdir -p "$d"
  printf "deb https://packagecloud.io/github/git-lfs/ubuntu/ jammy main\n" > "$d/github_git-lfs.list"
  printf "deb https://archive.ubuntu.com/ubuntu jammy main\n" > "$d/ubuntu.list"
  printf "deb https://cli.github.com/packages stable main\n" > "$d/github-cli.list"
  echo "before:"; ls "$d"
  set -eux
  find "$d/" -type f -exec grep -q "packagecloud\.io" {} \; -delete
  set +eux; echo "exit=$?  remaining:"; ls "$d"'
  before:
  github-cli.list
  github_git-lfs.list
  ubuntu.list
  + find /tmp/.../sources.list.d/ -type f -exec grep -q 'packagecloud\.io' '{}' ';' -delete
  exit=0  remaining:
  github-cli.list
  ubuntu.list
  ```
  Only `github_git-lfs.list` (the packagecloud source) was removed; `ubuntu.list` and `github-cli.list` survived.

- [x] **No-match / empty dir does not abort the build under `set -e`** *(Level 3 — the build must continue when there is no packagecloud source)*
  ```bash
  $ bash --noprofile --norc -c '
  empty=$(mktemp -d)
  set -eu
  find "$empty/" -type f -exec grep -q "packagecloud\.io" {} \; -delete
  echo "exit=$?  REACHED-NEXT"'
  exit=0  REACHED-NEXT
  ```

- [x] **Fix is grep-implementation-agnostic (adversarial)** *(Level 3 — the original `grep -rlZ … | xargs -0 -r rm -f` depends on GNU grep emitting NUL-delimited filenames; I reproduced a silent no-op on a non-GNU grep, then confirmed the shipped `find -exec grep -q … -delete` works on the same grep)*
  ```bash
  # The local grep is ugrep 7.5.0; `grep -rlZ` emits a trailing newline, not NUL:
  $ grep -rlZ 'packagecloud\.io' "$dir" | od -c | tail -1
  ... i   s   t  \n      # <-- newline, so `xargs -0` got a name with a trailing \n; `rm -f` silently no-op'd
  # The shipped find-based step relies only on grep's EXIT CODE (-q), not its output format,
  # so it deleted the right file on this same ugrep (see check above). It also works on the
  # GNU grep in catthehacker/ubuntu, so both runtimes are covered.
  ```
  This is why the implementation was changed from the `grep -Z | xargs -0` form to `find -exec grep -q … -delete`.

**Note:** doc-drift check ran clean — `docs/operations/running-workflows-locally.md` references this Dockerfile but asserts none of the facts the change touches (no `packagecloud`/`git-lfs`/apt-source mentions), so no doc update is needed.

Issue #1431, opened 2026-06-04
chore(code-health): stop VS Code from deleting unused imports (F401) on save


ktinubu:
## Problem

The committed `.vscode/settings.json` runs `source.fixAll.ruff` + `source.organizeImports.ruff` on save. `fixAll` applies F401 (unused-import) autofix, so an import you just typed but haven't used yet is deleted out from under you mid-edit.

## Change

Add an **editor-server-only** ruff override:

```json
"ruff.configuration": { "lint": { "unfixable": ["F401"] } }
```

This makes F401 unfixable for the VS Code ruff LSP only. Behavior preserved everywhere else:

- All other on-save autofixes and import sorting keep working.
- CLI `ruff check --fix` and the pre-commit `ruff` hook read `pyproject.toml` (no `unfixable`) and still strip unused imports.
- F401 is still surfaced in-editor as a diagnostic squiggle; it's just no longer auto-deleted.

## Scope

6-line addition to `.vscode/settings.json`. No Python/CI changes.

PR #1432, opened 2026-06-04
build: keep VS Code from deleting unused imports (F401) on save

ktinubu:
## What

Adds an **editor-server-only** ruff override to `.vscode/settings.json` so the VS Code ruff LSP stops auto-deleting unused imports (F401) on save:

```json
"ruff.configuration": { "lint": { "unfixable": ["F401"] } }
```

## Why

The committed settings run `source.fixAll.ruff` on save, which applies the F401 autofix — so an import you've typed but not yet used is deleted out from under you mid-edit.

## Scope of the behavior change

Only the in-editor ruff LSP is affected:

- ✅ All other on-save autofixes and import sorting keep working.
- ✅ CLI `ruff check --fix` and the pre-commit `ruff` hook read `pyproject.toml` (no `unfixable`) and **still strip** unused imports.
- ✅ F401 stays visible as an editor diagnostic squiggle — it's just no longer auto-deleted.

Verified `check-json` pre-commit passes (strict JSON, no comments).

Refs #1431

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -7,6 +7,12 @@
   "python.defaultInterpreterPath": "${workspaceFolder}/.venv/bin/python",
   "ruff.importStrategy": "fromEnvironment",
 
+  "ruff.configuration": {
+    "lint": {
+      "unfixable": ["F401"]
+    }
+  },
+
   "[python]": {
     "editor.defaultFormatter": "charliermarsh.ruff",
     "editor.formatOnSave": true,


ktinubu:
## Verification Results

The VS Code ruff LSP runs the ruff engine with `ruff.configuration` merged over `pyproject.toml`. Launching the actual VS Code extension headless isn't feasible (no editor LSP in this environment), so each check below runs the **real ruff CLI** with the exact config the LSP would use — a faithful Level-2 proxy for the editor's save-time behavior. The core contract (editor override does NOT leak into CLI/pre-commit) is proven by comparing the two config paths directly.

- [x] **Committed `settings.json` passes the strict `check-json` gate** *(Level 2 — runs the same pre-commit hook CI runs; the strict JSON parser is the actual consumer)*
  ```bash
  $ pre-commit run check-json --files .vscode/settings.json
  check json...............................................................Passed
  ```

- [x] **CLI / pre-commit (project config, no override) STILL strip unused imports** *(Level 2 — runs `ruff check --fix` with the repo's pyproject config; promise is "F401 still removed outside the editor")*
  ```bash
  $ printf 'import os\n\nx = 1\n' > b.py
  $ ruff check --select F401 --fix --no-cache b.py   # repo pyproject discovery
  Found 1 error (1 fixed, 0 remaining).
  $ cat b.py
  
  x = 1
  ```
  `import os` is removed — pre-commit/CLI behavior is unchanged by the editor override.

- [x] **Editor override (`lint.unfixable=["F401"]`) is accepted by ruff AND makes F401 unfixable** *(Level 2 — runs ruff with the exact config the LSP merges; promise is "save no longer deletes the import")*
  ```bash
  $ printf 'import os\n\nx = 1\n' > c.py
  $ ruff check --select F401 --fix --no-cache --config 'lint.unfixable=["F401"]' c.py
  help: Remove unused import: `os`
  Found 1 error.
  $ cat c.py
  import os
  
  x = 1
  ```
  ruff accepts the config (no schema error), keeps `import os`, and still **reports** F401 as a diagnostic — exactly the desired editor behavior (squiggle, no auto-delete).

- [x] **Override does NOT disable other autofixes — import sorting still applies** *(Level 2 — adversarial: proves the override is additive, not a blanket fix-disable)*
  ```bash
  $ printf 'import sys\nimport os\n\nprint(os, sys)\n' > d.py
  $ ruff check --select I001 --fix --no-cache --config 'lint.unfixable=["F401"]' d.py
  Found 1 error (1 fixed, 0 remaining).
  $ cat d.py
  import os
  import sys
  
  print(os, sys)
  ```
  Imports are re-sorted (`os` before `sys`) under the override — only F401 deletion is suppressed, every other on-save fix keeps working.

- [ ] **End-to-end: VS Code save action leaves unused import in place** *(Level 1 — SKIP: requires an interactive VS Code instance with the Ruff extension; no editor LSP available headless)*
  To verify manually: open a `.py` file in VS Code on this branch, type `import os` without using it, and save — the import should remain (yellow squiggle) instead of disappearing. The Level-2 checks above exercise the identical ruff invocation the LSP performs on save.

Issue #1445, opened 2026-06-05
refactor(testing): replace over-mocked MagicMock stubs with hand-written fakes in VST + pipeline tests


ktinubu:
## Context

A deep audit of test-double usage across the suite (104 test files) found the suite to be **disciplined overall** — `monkeypatch` (42 files) and hand-written fakes (19 classes) dominate over `unittest.mock` (18 files), with strong patterns like `fake_r2_remote` (real rclone-local backend), `FakeVST3Plugin`, and `_SpyClient`. This issue tracks the small set of localized over-mocking / brittleness candidates the audit surfaced, to bring the laggards in line with the patterns their sibling files already use.

## Scope (recommended fixes)

- [x] **1. `tests/data/vst/test_writers.py` — `_stub_render_dependencies()` (~lines 202-241).** Replace the `MagicMock` stubs for `load_plugin` / `load_preset` / `generate_sample` with a lightweight hand-written fake that tracks calls via lists. The sibling `tests/data/vst/test_generate_vst_dataset.py` already does this well with `_silent_render` / `_loud_render`; mirror that style. Behaviour must be unchanged.
- [x] **2. `tests/data/vst/test_core.py` — plugin `process()` MagicMock (~line 277) + `logger.warning.call_count == 1` assertion (~line 203).** Swap the `process()` mock for a tiny fake returning a correctly-shaped audio array, and replace the brittle log-call-count assertion with an assertion on the raised `RenderWorkerLeaked` exception. The threading mocks in this file are justified and should stay.
- [x] **3. `sky.Resources` / `sky.Task` `spec=[...]` lists in `tests/pipeline/entrypoints/test_skypilot_launch.py`.** The hand-maintained `spec=[...]` lists can drift from the real SkyPilot SDK surface. Evaluate `create_autospec` against the real classes (guarded by import availability) so the spec tracks the SDK automatically.
- [x] **4. `tests/pipeline/entrypoints/test_finalize_dataset.py` — `load_spec_from_uri` call-order coupling.** The bad-spec error test assumes `load_spec_from_uri` is called before dispatch; it would silently pass if that call moved later. Add a positive assertion that the dispatch is actually short-circuited (e.g. that no download/upload side effect occurred) so the test fails loudly if the order changes.

## Non-goals

[...]

PR #1446, opened 2026-06-05
test(testing): assert finalize short-circuits on bad output_format

ktinubu:
## What

Strengthen `test_finalize_raises_on_unsupported_output_format` in
`tests/pipeline/entrypoints/test_finalize_dataset.py` so it no longer
passes by accident if the spec-load were to move *after* dispatch.

## Why

The test monkeypatched `load_spec_from_uri` to return a `parquet`
(unsupported) spec and asserted only that `finalize()` raises
`ValueError`. That assertion is coupled to call **order**: it implicitly
assumes the spec load happens before dispatch, and would still pass if
that load moved later — never positively proving dispatch was
short-circuited.

## How

Added fail-fast `download_to_path` / `upload` stubs (reusing the exact
pattern already in this file's
`test_finalize_hdf5_marker_idempotency_short_circuits_before_download`).
If any format branch ran, a download or upload would fire and call
`pytest.fail` before the `ValueError`, failing the test. The original
`pytest.raises(ValueError, match="unsupported output_format")` assertion
is preserved. No production code changed.

`uv run pytest tests/pipeline/entrypoints/test_finalize_dataset.py -q`:
22 passed. `make format`: all hooks pass.

Refs #1445

### DIFF
--- a/tests/pipeline/entrypoints/test_finalize_dataset.py
+++ b/tests/pipeline/entrypoints/test_finalize_dataset.py
@@ -323,9 +323,14 @@ def test_finalize_raises_on_unsupported_output_format(
     Pins the dispatcher's exhaustiveness contract — adding a third format
     without wiring its branch must trip this test rather than silently
     skip the artifact upload and write a misleading ``dataset.complete``.
+    The fail-fast ``download_to_path`` / ``upload`` stubs make this a
+    positive short-circuit check: the ValueError alone would still pass if
+    the spec load moved *after* dispatch, but a download or upload firing
+    before the raise proves dispatch ran and fails the test.
 
     :param tmp_path: Pytest tmp dir; hosts the Hydra-style output_dir.
-    :param monkeypatch: Pytest fixture used to install a stub loader.
+    :param monkeypatch: Pytest fixture used to install a stub loader plus
+        fail-fast download/upload stubs.
     :param stub_finalize_setup: Installs the auth + marker-probe stubs so the
         dispatcher (not the marker check) is the failure surface.
     """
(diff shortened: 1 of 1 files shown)


ktinubu:
## Verification Results

The PR's promise is narrow: `test_finalize_raises_on_unsupported_output_format` now *positively* proves `finalize()` short-circuits before any R2 transport on an unsupported `output_format`, rather than passing by accident if the spec-load moved after dispatch.

- [x] **Strengthened test passes on the PR branch** *(Level 2 — invoked pytest directly on the exact test; promise is test behavior)*
  ```bash
  $ uv run pytest "tests/pipeline/entrypoints/test_finalize_dataset.py::test_finalize_raises_on_unsupported_output_format" -q
  1 passed in 1.06s
  ```

- [x] **The fail-fast stubs are load-bearing (guard bites when dispatch actually runs)** *(Level 2 — negative control: ran a throwaway probe that installs the same `download_to_path` / `upload` fail-fast stubs but feeds a **valid** wds spec so a real format branch executes; the download stub fires and fails with the PR's exact message. This proves a non-short-circuited dispatch is caught, not silently passed. Probe deleted after the run; no production code was modified.)*
  ```text
  src/synth_setter/cli/finalize_dataset.py:180: in finalize_from_spec
      finalize_wds(spec, work_dir)
  src/synth_setter/cli/finalize_dataset.py:54: in _download_train_shards_one_at_a_time
      r2_io.download_to_path(spec.r2.shard_uri(shard), local)
  >   lambda *a, **kw: pytest.fail("download_to_path should not be reached"),
  E   Failed: download_to_path should not be reached
  FAILED tests/.../test_guard_bites_probe.py::test_guard_bites_when_dispatch_runs
  ```

- [x] **Full file suite stays green** *(Level 2 — the change touches one test in this module; ran the whole module to confirm no fixture/stub regressions)*
  ```bash
  $ uv run pytest tests/pipeline/entrypoints/test_finalize_dataset.py -q
  22 passed in 2.01s
  ```

No production code changed. `make format` (ruff, pydoclint, pyright, prettier) passed pre-commit before the commit; CI re-runs both regardless, so they are not listed as standalone checks here.

PR #1449, opened 2026-06-05
refactor(testing): replace MagicMock render stubs with handwritten fakes

ktinubu:
## What

Replace the `MagicMock` doubles in `_stub_render_dependencies()` (in `tests/data/vst/test_writers.py`) with lightweight hand-written fakes (`_FakePlugin` / `_FakeVSTDataSample`).

## Why

The helper only threads the loaded plugin through the writer loop (tests assert identity via `is`) and collects rendered samples into a flush batch (tests assert batch length). Neither needs the attribute-fabricating behaviour of `MagicMock`. Explicit fakes match the list-based capture idiom the sibling `test_generate_vst_dataset.py` already uses (`_silent_render` / `_loud_render`, `load_plugin_calls` / `load_preset_calls`), and make the test doubles' surface area explicit.

## Scope

- Only the doubling mechanism inside `_stub_render_dependencies` changed.
- Behaviour and assertions are identical; the `monkeypatch` seams targeting the real `load_plugin` / `load_preset` / `generate_sample` boundaries are untouched.
- No production code changed.

## Verification

- `uv run pytest tests/data/vst/test_writers.py -q` → 26 passed.
- `make format` → clean (docformatter, ruff, pyright, pydoclint, interrogate all pass).

Refs #1445

### DIFF
--- a/tests/data/vst/test_writers.py
+++ b/tests/data/vst/test_writers.py
@@ -199,6 +199,28 @@ def test_main_rejects_unknown_suffix(tmp_path: Path) -> None:
     mock_wds.assert_not_called()
 
 
+class _FakePlugin:
+    """Stand-in for a loaded VST plugin handle.
+
+    Carries no behaviour — the writer loop only threads it through and tests
+    assert identity (``is``), so a bare object with a debug ``repr`` suffices.
+    """
+
+    def __repr__(self) -> str:
+        return "_FakePlugin()"
+
+
+class _FakeVSTDataSample:
+    """Stand-in for the sample object returned by ``generate_sample``.
+
+    The writer loop only collects these into the flush batch; tests assert on batch length, so a
+    bare object suffices.
+    """
+
+    def __repr__(self) -> str:
+        return "_FakeVSTDataSample()"
+
+
 def _stub_render_dependencies(
     monkeypatch: pytest.MonkeyPatch,
     *,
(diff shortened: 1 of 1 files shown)


ktinubu:
## Verification Results

Verified on a fresh checkout of the PR branch (`git worktree add /tmp/verify-pr-1449 origin/refactor/testing-writers-fake`, HEAD `6d9197b4`).

- [x] **The refactored test doubles preserve identical test behaviour — all 26 tests in `test_writers.py` still pass** *(Level 1 — for a test-only refactor the system under test is the test suite itself; running pytest end-to-end on the affected file is the full-integration check. A failure here would mean the `MagicMock`→fake swap broke a writer-loop assertion.)*
  ```bash
  $ python -m pytest tests/data/vst/test_writers.py -p no:cacheprovider -q 2>&1 | tail -1
  ============================== 26 passed in 0.67s ==============================
  ```

- [x] **The new fakes carry no auto-fabricated attributes — the `MagicMock` failure mode is genuinely eliminated** *(Level 2 — invokes the fakes directly; the promise is that they are NOT mock-like. A `MagicMock` would silently fabricate `.synth_params` / arbitrary methods and let a typo'd assertion pass; `_FakePlugin` / `_FakeVSTDataSample` must raise `AttributeError` instead. This is the adversarial check that the refactor actually achieves its goal rather than just swapping one opaque double for another.)*
  ```bash
  $ python -c "
  import importlib.util
  spec = importlib.util.spec_from_file_location('twr', 'tests/data/vst/test_writers.py')
  m = importlib.util.module_from_spec(spec); spec.loader.exec_module(m)
  p = m._FakePlugin(); s = m._FakeVSTDataSample()
  try: p.some_undefined_method(); print('FAIL: _FakePlugin fabricated an attribute')
  except AttributeError: print('PASS: _FakePlugin raises AttributeError on undefined attr')
  try: s.synth_params; print('FAIL: _FakeVSTDataSample fabricated an attribute')
  except AttributeError: print('PASS: _FakeVSTDataSample raises AttributeError on undefined attr')
  print('PASS: distinct instances' if m._FakePlugin() is not m._FakePlugin() else 'FAIL: shared instance')
  "
  PASS: _FakePlugin raises AttributeError on undefined attr
  PASS: _FakeVSTDataSample raises AttributeError on undefined attr
  PASS: distinct instances
  ```

### Notes

- `make test` / `make format` are intentionally omitted — CI runs them. The two checks above target this PR's specific promise (behaviour-preserving double swap), not generic suite health.

PR #1451, opened 2026-06-05
refactor(eval): type decode + predict_vst_audio notes with NoteParams

### LINKED ISSUE #1428: refactor(eval): type ParamSpec.decode + predict_vst_audio note params with NoteParams
Follow-up to #1426 (PR #1427). That PR threaded `NoteParams` through the VST *generate* path but deliberately left the *decode* side untyped: `ParamSpec.decode` still returns `tuple[dict[str, float], dict[str, float]]`, where the note half should be `NoteParams` (its `note_start_and_end` is a tuple, not a float).

Correcting `decode` requires also untangling `predict_vst_audio.py`'s unrelated pre-existing type errors (`param_spec`/`Path` typed as `str`, unguarded `None` subscripts at :195/197), which is why it was scoped out of #1427.

Scope: type `ParamSpec.decode`'s note return as `NoteParams` (cast at the construction source, mirroring `sample()`), fix the consuming annotations in `predict_vst_audio.py`, and remove the now-stale `# pyright: ignore[reportArgumentType]` on the decode-fed append in `tests/data/vst/test_generate_vst_dataset.py`.

ktinubu:
## What

Follow-up to #1427, which threaded the `NoteParams` TypedDict through the VST *generate* path but deliberately left the *decode* side returning `tuple[dict[str, float], dict[str, float]]` — imprecise because the note half's `note_start_and_end` is a tuple, not a float. That side was scoped out of #1427 because correcting it required also untangling `predict_vst_audio.py`'s unrelated pre-existing type errors.

This PR closes that gap. **Type-only — no runtime behavior change** (`cast` is a no-op at runtime; the local renames are mechanical).

## Changes

- **`param_spec.py`**: `ParamSpec.decode` now returns `tuple[dict[str, float], NoteParams]`, asserted via a single `cast(NoteParams, …)` at the dynamic-construction source — mirroring the existing `sample()` (keys come from runtime `Parameter.name`, so the checker can't prove the mapping).
- **`predict_vst_audio.py`**: untangle the pre-existing errors that blocked this in #1427 —
  - `param_spec`/`pred_dir` are no longer reassigned across `str`/`ParamSpec`/`Path` (new `spec`/`pred_path` locals), removing a real shadowing hazard;
  - `params_to_csv`'s `target_*` params widen to optionals because the `--no-params` path genuinely passes `None`;
  - note params adopt `NoteParams`, and the `:195`/`:197` None-subscript errors fall out of the now-narrowed `decode` return type.
- **tests**: drop the now-stale `reportArgumentType` ignores (decode- and encode-fed) in `test_generate_vst_dataset.py` and the `params_to_csv(None, …)` `# type: ignore[arg-type]` in `test_predict_vst_audio.py`; `_sample_param_dicts` returns `NoteParams`. The stale explanatory comments referencing "a sibling PR" go with them.
- **`.pydoclint-baseline.txt`**: `params_to_csv`'s existing DOC103 row updated in place for the new annotations (no new rows — append-freeze respected, matching #1427's precedent).

## Scope

`make_spectrogram` / `write_spectrograms` return-type errors (`:36`/`:44`) are unrelated to note params and left untouched. Both `param_spec.py` and `predict_vst_audio.py` remain pyright-excluded; graduating either from the exclude list is a separate `/lint-cleanup` task. The type-checking value here lands in the non-excluded test files, which now drive `decode`/`params_to_csv` with the real `NoteParams` shape.

## Verification

- Full-repo pyright: **0 errors** (matches pre-change state).
- `predict_vst_audio` suite green (15 passed serially); non-slow VST data tests green (99 passed).
- `make format` / pre-commit clean (ruff, pyright, pydoclint, …).
- Pre-PR gate: `/repo-review-full-no-comments` → 0 BLOCK across 7 skills.

Fixes #1428

### DIFF
--- a/.pydoclint-baseline.txt
+++ b/.pydoclint-baseline.txt
@@ -85,7 +85,7 @@ src/synth_setter/evaluation/compute_audio_metrics.py
 --------------------
 src/synth_setter/evaluation/predict_vst_audio.py
     DOC101: Function `params_to_csv`: Docstring contains fewer arguments than in function signature.
-    DOC103: Function `params_to_csv`: Docstring arguments are different from function arguments. (Or could be other formatting issues: https://jsh9.github.io/pydoclint/violation_codes.html#notes-on-doc103 ). Arguments in the function signature but not in the docstring: [param_spec: ParamSpec, pred_note_params: dict[str, float], pred_synth_params: dict[str, float], save_path: str, target_note_params: dict[str, float], target_synth_params: dict[str, float]].
+    DOC103: Function `params_to_csv`: Docstring arguments are different from function arguments. (Or could be other formatting issues: https://jsh9.github.io/pydoclint/violation_codes.html#notes-on-doc103 ). Arguments in the function signature but not in the docstring: [param_spec: ParamSpec, pred_note_params: NoteParams, pred_synth_params: dict[str, float], save_path: str, target_note_params: NoteParams | None, target_synth_params: dict[str, float] | None].
 --------------------
 src/synth_setter/metrics.py
     DOC101: Method `LogSpectralDistance.__init__`: Docstring contains fewer arguments than in function signature.
--- a/src/synth_setter/data/vst/generate_vst_dataset.py
+++ b/src/synth_setter/data/vst/generate_vst_dataset.py
@@ -1,6 +1,6 @@
 from dataclasses import dataclass
 from pathlib import Path
-from typing import Any, cast
+from typing import Any
 
 import h5py
 import hdf5plugin
@@ -279,9 +279,7 @@ def fixed_params_from_dataset(
     for row in param_array:
         synth_params, note_params = param_spec.decode(row)
         synth_params_list.append(synth_params)
-        # decode is annotated dict[str, float]; pitch is int and note_start_and_end
-        # a tuple at runtime, matching NoteParams' closed shape.
-        note_params_list.append(cast(NoteParams, note_params))
+        note_params_list.append(note_params)
     return synth_params_list, note_params_list
 
 
--- a/src/synth_setter/data/vst/param_spec.py
+++ b/src/synth_setter/data/vst/param_spec.py
@@ -269,7 +269,7 @@ class ParamSpec:
 
         return np.concatenate((synth_params, note_params))
 
-    def decode(self, params: np.ndarray) -> tuple[dict[str, float], dict[str, float]]:
+    def decode(self, params: np.ndarray) -> tuple[dict[str, float], NoteParams]:
         synth_params_to_process = [(p, len(p)) for p in self.synth_params]
         note_params_to_process = [(p, len(p)) for p in self.note_params]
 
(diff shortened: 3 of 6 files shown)


ktinubu:
## Verification Results

Type-only refactor (`ParamSpec.decode` → `NoteParams`; consuming annotations + stale-ignore removal). Verified against branch HEAD `8bef7428`. The promise is "types are now accurate, the type-checker accepts the change with no stale ignores, and runtime behavior is unchanged."

- [x] **Type-checker accepts the change repo-wide** *(Level 2 — invokes pyright, the system under test for a typing change; promise is tool behavior, not end-to-end)*
  ```bash
  $ uv run pyright 2>&1 | tail -1
  ```
  Matches the pre-change baseline (main was also 0 errors), confirming the relaxed `params_to_csv` signature and `decode` return type introduce no regressions.

- [x] **decode → main → params_to_csv still works end-to-end** *(Level 1 — runs the `predict_vst_audio` CLI `main()` with the VST3 render patched out, exercising the real decode→render→CSV path on CPU; lower levels would not exercise the full contract)*
  ```bash
  $ uv run pytest tests/evaluation/test_predict_vst_audio.py tests/evaluation/test_predict_vst_audio_pin.py -p no:xdist -q 2>&1 | tail -1
  15 passed in 3.30s
  ```
  Includes `test_params_to_csv_none_target_leaves_target_column_nan`, which pins the `--no-params` (`None` target) path the widened optional annotation now expresses.

- [x] **No stale type-ignores remain in the touched test files** *(Level 2, adversarial — re-runs pyright with `reportUnnecessaryTypeIgnoreComment=error`; proves the removed ignores were genuinely obsolete AND none were left behind)*
  ```bash
  $ uv run pyright -p <tmp config: the two touched test files, reportUnnecessaryTypeIgnoreComment=error> 2>&1 | tail -1
  ```

- [x] **`decode`'s `cast(NoteParams, …)` does not lie about the runtime shape** *(Level 3, adversarial — feeds a real encoding through the real `surge_simple` `ParamSpec.decode` and asserts the structure the TypedDict promises)*
  ```bash
  $ uv run python -c "<see below>"
  PASS: keys= ['note_start_and_end', 'pitch']
    pitch = 65 ( int )
    note_start_and_end = (np.float32(3.7096958), np.float32(3.8717048)) (tuple of float32 )
  ```
  ```python
  import numbers
  import numpy as np
  from synth_setter.data.vst import param_specs
  spec = param_specs['surge_simple']
  encoded = np.random.default_rng(0).random(len(spec)).astype(np.float32)
  synth, note = spec.decode(encoded)
  assert set(note.keys()) == {'pitch', 'note_start_and_end'}
  assert isinstance(note['pitch'], int)
  nse = note['note_start_and_end']
  assert isinstance(nse, tuple) and len(nse) == 2
  assert all(isinstance(x, numbers.Real) for x in nse)
  assert isinstance(synth, dict)
  ```
  Note: `note_start_and_end` values are `np.float32` scalars, which `NoteParams` models as `float`. That is a pre-existing imprecision in the `NoteParams` TypedDict (introduced with `sample()` in #1427), not changed here.

- [x] **pydoclint accepts the in-place baseline-row edit** *(Level 2 — runs the pydoclint hook against the baseline; proves the updated `params_to_csv` DOC103 row matches the new signature so the append-frozen baseline stays consistent)*
  ```bash
  $ uv run pre-commit run pydoclint --files src/synth_setter/evaluation/predict_vst_audio.py src/synth_setter/data/vst/param_spec.py 2>&1 | tail -1
  ```

PR #1459, opened 2026-06-05
internal-feat(data): validate a dataset-copy source against its spec

### LINKED ISSUE #1429: feat(data): copy an existing dataset's params into a new generate_dataset run
## Summary

Add a `datasetsrc` config to dataset generation so a run can **copy an existing dataset**: instead of sampling fresh parameters, each output shard re-renders the parameters of the same-named shard under `datasetsrc.copy_dataset_root`.

## Behavior

- New nested config `DatasetSrcConfig` with a single field `copy_dataset_root` (a worker-local directory of source shards), surfaced as `DatasetSpec.datasetsrc` (default `None` = sample fresh).
- The launcher forwards `--copy_dataset_root` to the `generate_vst_dataset` subprocess only when set.
- `generate_vst_dataset` decodes the same-named source shard's `param_array` into `fixed_synth_params_list` / `fixed_note_params_list` (via `param_spec.decode`) and renders those, reproducing the source dataset's parameters under a fresh render.
- The source must share the target's `render.param_spec_name` (same encoding width); a width or row-count mismatch raises with a clear message.

## Use case

Re-render an existing dataset's exact parameter set under a different renderer version, sample rate, or audio settings — without re-sampling.

## Acceptance

- Tests exercising and validating the conversion, CLI wiring, resume indexing, and a VST round-trip that reproduces the source `param_array`.
- Green CI.

ktinubu:
## What

Adds a launch-time preflight that validates a dataset-copy run against the source dataset's persisted spec before any shard renders.

The copy path (merged in #1430) matches source→target shards purely by filename, silently assuming the target run mirrors the source's layout. A mismatched `param_spec_name`, `samples_per_shard`, split sizes, or shard count would otherwise surface one render — or one re-run — at a time, or produce a faithful-looking copy with a different train/val/test partition.

## How

- `DatasetSpec.validate_copy_source(source)` — a pure comparison asserting the source reproduces the target on every copy-relevant value (`param_spec_name`, `samples_per_shard`, `train_val_test_sizes`, and the derived shard-filename set). All mismatches are aggregated into one error.
- `_validate_copy_source(spec)` in `generate()` — loads the source's `input_spec.json` (via `spec_io.load_spec_from_uri`) from `<copy_dataset_root>/` and runs the comparison before the first render. A missing source spec is itself a launch error; a clean pass logs launch-log evidence.

Functional-core / imperative-shell split: the pure comparison is unit-tested directly; the I/O shell is tested via real spec round-trips and a `generate()`-level abort test (`_renderer_argv_lists == []`).

**Out of scope** (deliberately): re-sharding, param-spec/encoding migration, subset copy.

## Testing

- 13 new tests: 7 pure-core (each field mismatch with value-pair assertions, partial-overlap, aggregation count, irrelevant-field-varies, happy path) + 6 shell/integration (no-op, match, missing-spec, mismatch, and two `generate()`-level abort tests).
- Full affected modules green (`make format` / pre-commit clean: ruff, pyright, pydoclint, docformatter).

Fixes #1429


ktinubu:
## Verification Results

Verified on branch `feat/dataset-copy-source-preflight` @ `e98bbf87` (matches origin). The PR's promise: a misconfigured dataset-copy **fails at launch, before any render**, with an accurate aggregated error; a matching source passes.

- [x] **A misconfigured copy aborts at launch before any renderer subprocess runs** *(Level 1 — drives the real `generate()` launch entrypoint; the renderer subprocess is the audio boundary and is the only thing patched, so "no renderer invoked" is the end-to-end abort contract)*

  ```bash
  $ uv run --active pytest \
      "tests/pipeline/entrypoints/test_generate_dataset_unit.py::TestRun::test_aborts_before_render_when_copy_source_spec_is_missing" \
      "tests/pipeline/entrypoints/test_generate_dataset_unit.py::TestRun::test_aborts_before_render_when_copy_source_spec_mismatches" -v
  ...::test_aborts_before_render_when_copy_source_spec_mismatches PASSED [ 50%]
  ...::test_aborts_before_render_when_copy_source_spec_is_missing  PASSED [100%]
  2 passed in 1.13s
  ```

  Both tests assert `_renderer_argv_lists(patched_subprocess) == []` after `generate()` raises — no shard render fired.

- [x] **A multi-field mismatch produces one aggregated, value-bearing error** *(Level 2 — black-box the real `DatasetSpec.validate_copy_source` with two real specs; shows the actual message)*

  ```
  dataset-copy source spec does not match the target on copy-relevant values:
    - param_spec_name: source='surge_xt' != target='surge_simple'
    - shard filenames: source has 1, target has 1 (only in source: ['shard-000000.tar']; only in target: ['shard-000000.h5'])
  ```

  One error lists every mismatch with both source and target values — not one failure per launch.

- [x] **A missing source `input_spec.json` fails with an actionable error naming the file** *(Level 2/3 — black-box `_validate_copy_source` against a real empty copy root on disk)*

  ```
  dataset-copy source has no input_spec.json at /tmp/.../empty-src/input_spec.json; sync the source dataset's spec alongside its shards (it lives beside the shards at the R2 dataset prefix root) so the copy can be validated.
  ```

- [x] **A matching source passes — including identity fields that legitimately differ, and a real on-disk round-trip** *(Level 2/3 — black-box, source carries a different `task_name`; shell reads a real `input_spec.json` from disk)*

  ```
  [1] matching source (differing task_name): PASS (no raise)
  [4] on-disk matching input_spec.json: PASS (no raise)
  2026-06-05 ... | INFO | ..._validate_copy_source:248 - dataset-copy source OK: /tmp/.../good-src/input_spec.json matches the target spec
  ```

  Only the four copy-relevant fields gate the copy; the success path logs launch-log evidence the preflight ran.

- [ ] **True end-to-end copy: render real shards from a synced source dataset through the VST renderer** *(Level 1, full integration — SKIP: requires a VST3 plugin + a synced source dataset with shards, unavailable in this environment)*
  To verify manually:

  ```bash
  # 1. Sync a source dataset (shards + input_spec.json) locally, e.g.
  rclone copy r2:intermediate-data/data/<task>/<run_id>/ /data/source-dataset/ --checksum
  # 2. Launch a copy run whose render.param_spec_name / samples_per_shard /
  #    train_val_test_sizes / output_format match the source, with:
  #    datasetsrc='{copy_dataset_root: /data/source-dataset}'
  # 3. Confirm: a matching config renders; a deliberately mismatched config
  #    (e.g. different samples_per_shard) aborts at launch before any shard render.
  ```

Issue #1464, opened 2026-06-05
feat(data-pipeline): bring generate_dataset entrypoint to train/eval parity (extras + wandb provenance)


ktinubu:
## Context

The `generate_dataset` Hydra entrypoint (`src/synth_setter/cli/generate_dataset.py`) is missing two pieces of the standard entrypoint hygiene that `train.py` and `eval.py` both apply. These were surfaced while diagnosing the wandb empty-log bug (fixed separately via `console: redirect`). They are **orthogonal to that fix** and tracked here so the logs PR stays focused.

### 1. `extras(cfg)` — apply optional pre-task utilities

`train`/`eval` call `extras(cfg)` (`synth_setter.utils.extras`) in their `main()`, gated by an `extras: default` config group. `dataset.yaml` has neither the call nor the group. This is a **two-part change**:

- Add `- extras: default` to `dataset.yaml`'s `defaults:`.
- Call `extras(cfg)` at the top of the worker entrypoint `from_hydra` (and `main` as appropriate).

**Tags handling (decided):** the stock `extras/default.yaml` sets `enforce_tags: True`, but `dataset.yaml` has no `tags` field — it uses `task_name`. Rather than disable `enforce_tags`, **add `tags` support to the generate_dataset config** so the enforce-tags path works as designed. (Note: `enforce_tags` hard-raises under multirun when tags are absent, so the cadence sweep `sweeps/generate_dataset_cadence*.yaml` must supply tags or this needs care.)

### 2. `log_wandb_provenance()` — stamp git SHA / image tag / command

`train`/`eval` call `log_wandb_provenance()` (`synth_setter.utils.logging_utils`) after logger instantiation; it records `github_sha`, `image_tag` (`$IMAGE_TAG`), and the full command into `wandb.config`. `generate_dataset` has `_log_hyperparams` + `_log_spec_artifact` but **not** provenance.

This matters *more* for dataset generation than for training: the pipeline already pins `renderer_version` and cross-checks the plugin for reproducibility, and knowing which image/SHA rendered a dataset is exactly that lineage. Suggested placement: inside `generate()` right after `_log_hyperparams(loggers, spec)`, where the wandb run is live and shared by both the local `main()` and worker `from_hydra` paths.

## Explicitly out of scope

[...]

PR #1466, opened 2026-06-05
fix(monitoring): capture wandb console at fd level for non-empty logs

### LINKED ISSUE #1465: bug(monitoring): wandb runs from generate_dataset have an empty log stream
## Symptom

Dataset-generation runs appear in wandb, but the run's **Logs** stream is empty — no loguru output and none of the per-shard render-subprocess output is captured. Observed on error/crash runs and normal runs alike.

## Root cause

`configs/logger/wandb.yaml` left console capture at the wandb default `console=auto`, which resolves to `wrap` on non-TTY workers. `wrap` only patches the parent process's `sys.stdout`, so it captures:

- **Neither** generate_dataset's loguru output — loguru's default handler binds a reference to `sys.stderr` at import time (`from loguru import logger`), long before `wandb.init` runs inside `generate()`, so the wrapper never sees it;
- **Nor** the render-subprocess output — `wrap` is Python-level only and cannot capture a child process's stdout/stderr.

Only fd-level `console=redirect` captures both: it dups file descriptors 1/2, which survives loguru's early stderr binding and is inherited by subprocesses.

## Fix

Pin `console: redirect` in the wandb logger `settings` block. Shared with train/eval (wandb-recommended default; coexists with tqdm).

## Not addressed here

This is the empty-logs fix only. `@task_wrapper` was evaluated and rejected (its `wandb.finish()` is already done by `generate()`'s `_close_loggers`); entrypoint parity (`extras`/provenance) is tracked separately in #1464.

ktinubu:
## Problem

Dataset-generation runs appear in wandb but their **Logs** stream is empty — no loguru output and none of the per-shard render-subprocess output is captured.

## Root cause

`configs/logger/wandb.yaml` left console capture at the wandb default `console=auto`, which resolves to `wrap` on non-TTY workers. `wrap` only patches the parent's `sys.stdout`, so it captures:

- **Neither** generate_dataset's loguru output — loguru's default handler binds a reference to `sys.stderr` at import time (`from loguru import logger`), before `wandb.init` runs inside `generate()`, so the wrapper never sees it;
- **Nor** the render-subprocess output — `wrap` is Python-level only and cannot capture a child process's fds.

Only fd-level `console=redirect` captures both (it dups fds 1/2, surviving loguru's early binding and inherited by subprocesses).

## Change

- Pin `console: redirect` in the wandb logger `settings` block.
- Add a config-composition test (`TestWandbConsoleCapture`) asserting the composed `wandb.settings.console == "redirect"` through real Hydra composition.

Shared with `train`/`eval` (wandb-recommended default; coexists with tqdm progress bars).

## Verification

- `tests/schemas/test_logger_config.py` — green (config-pin via real Hydra composition).
- Confirmed `wandb.Settings(console="redirect")` is accepted by the installed wandb 0.26.1.
- [ ] **Manual e2e (needs a wandb-authenticated run):** run a `generate_dataset` smoke shard that fails mid-render and confirm the wandb run's log stream contains the loguru lines **and** the `generate_vst_dataset.py` subprocess output. The committed test pins the config value but does not exercise runtime capture.

## Scope notes

- `@task_wrapper` was evaluated and **rejected**: its only effect is `wandb.finish()` on exception, which `generate()`'s existing `_close_loggers` already does (and more carefully — it marks `finalize("failed")` and guards against finishing a foreign `wandb.run`). It also does not affect console capture, so it would not fix this bug.
- Entrypoint parity (`extras(cfg)` + `log_wandb_provenance()`, plus adding a `tags` field to the dataset config) is tracked separately in #1464.

Fixes #1465

### DIFF
--- a/docs/reference/wandb-integration.md
+++ b/docs/reference/wandb-integration.md
@@ -21,15 +21,16 @@ ______________________________________________________________________
 
 ## 1. Initialization
 
-| Concern           | How it works                                                                                             | File                                                |
-| ----------------- | -------------------------------------------------------------------------------------------------------- | --------------------------------------------------- |
-| W&B run creation  | `WandbLogger` instantiated by Hydra — included in the default `many_loggers` compose (W&B + CSV + TB)    | `src/synth_setter/configs/logger/wandb.yaml`        |
-| Entity / project  | Env-var driven: `entity: ${oc.env:WANDB_ENTITY,null}`, `project: "${oc.env:WANDB_PROJECT,synth-setter}"` | `src/synth_setter/configs/logger/wandb.yaml:10,13`  |
-| Default compose   | `many_loggers` composes `csv + tensorboard + wandb` (W&B enabled by default)                             | `src/synth_setter/configs/logger/many_loggers.yaml` |
-| Run ID            | `null` (W&B auto-generates)                                                                              | `src/synth_setter/configs/logger/wandb.yaml:8`      |
-| Checkpoint upload | `log_model: "all"`                                                                                       | `src/synth_setter/configs/logger/wandb.yaml:11`     |
-| Code saving       | `wandb.Settings(code_dir=".")`                                                                           | `src/synth_setter/configs/logger/wandb.yaml:17-19`  |
-| Run teardown      | `wandb.finish()` in `task_wrapper` finally block                                                         | `src/synth_setter/utils/utils.py:101-106`           |
+| Concern           | How it works                                                                                                                      | File                                                            |
+| ----------------- | --------------------------------------------------------------------------------------------------------------------------------- | --------------------------------------------------------------- |
+| W&B run creation  | `WandbLogger` instantiated by Hydra — included in the default `many_loggers` compose (W&B + CSV + TB)                             | `src/synth_setter/configs/logger/wandb.yaml`                    |
+| Entity / project  | Env-var driven: `entity: ${oc.env:WANDB_ENTITY,null}`, `project: "${oc.env:WANDB_PROJECT,synth-setter}"`                          | `src/synth_setter/configs/logger/wandb.yaml:10,13`              |
+| Default compose   | `many_loggers` composes `csv + tensorboard + wandb` (W&B enabled by default)                                                      | `src/synth_setter/configs/logger/many_loggers.yaml`             |
+| Run ID            | `null` (W&B auto-generates)                                                                                                       | `src/synth_setter/configs/logger/wandb.yaml:8`                  |
+| Checkpoint upload | `log_model: "all"`                                                                                                                | `src/synth_setter/configs/logger/wandb.yaml:11`                 |
+| Code saving       | `wandb.Settings(code_dir=".")`                                                                                                    | `src/synth_setter/configs/logger/wandb.yaml` § `wandb.settings` |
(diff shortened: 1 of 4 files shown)


ktinubu:
## Verification Results

Branch `fix/wandb-console-redirect` @ `e8ef34e`. The end-to-end check below was run **offline** (`WANDB_MODE=offline`), which still writes wandb's console-capture `output.log` — so no cloud auth was needed to exercise the real capture mechanism.

- [x] **The end-to-end promise: `console=redirect` captures loguru (stderr) + subprocess output; `wrap` drops both** *(Level 1 — exercises wandb's real console capture with a real loguru import and a real child process; contrastive before/after)*
  ```bash
  $ uv run --no-sync python /tmp/verify_console_capture.py
  2026-06-05 22:51:19 | INFO | LOGURU_MARKER_via_stderr
  SUBPROC_MARKER_child_stdout
  === console=wrap ===
    loguru (stderr) captured:    False
    subprocess stdout captured:  False
  2026-06-05 22:51:20 | INFO | LOGURU_MARKER_via_stderr
  SUBPROC_MARKER_child_stdout
  === console=redirect ===
    loguru (stderr) captured:    True
    subprocess stdout captured:  True
  ```
  `wrap` (what `auto` resolves to on non-TTY workers) misses both → empty log stream, reproducing the bug. `redirect` captures both. The script imports `loguru` before `wandb.init` (mirroring the real import-time stderr binding) and reads each offline run's `files/output.log`.

- [x] **The composed config pins `console=redirect` through real Hydra composition** *(Level 2 — runs the committed test, which composes via `hydra.compose`, not raw-YAML grep)*
  ```bash
  $ uv run --no-sync pytest tests/schemas/test_logger_config.py::TestWandbConsoleCapture::test_wandb_settings_console_is_redirect --confcutdir=tests/schemas -q
  1 passed in 0.45s
  ```

- [x] **`redirect` is a value `wandb.Settings` actually accepts — instantiated through the production `_target_` path** *(Level 2 — closes the reviewers' validity-gap WARN: this would fail at instantiate-time on a typo, unlike a string-equality assert)*
  ```bash
  $ uv run --no-sync python -c "
  import hydra
      cfg = compose(config_name='train.yaml', overrides=['datamodule=ksin','model=ffn','trainer=cpu','logger=wandb'])
  settings = hydra.utils.instantiate(cfg.logger.wandb.settings)
  print(type(settings).__module__ + '.' + type(settings).__name__, settings.console)
  "
  wandb.sdk.wandb_settings.Settings redirect
  ```

- [x] **Adversarial: the shared config does not break `eval.yaml` composition** *(Level 2 — `console` lives in the wandb logger group shared by train/eval; composes + instantiates cleanly for eval too)*
  ```bash
  $ uv run --no-sync python -c "... compose eval.yaml + logger=wandb; instantiate settings ..."
  eval.yaml + logger=wandb composes; wandb.Settings.console = redirect
  PASS
  ```

- [ ] **Full `generate_dataset` smoke shard on the cluster confirms a non-empty wandb log stream** *(Level 1 — SKIP: needs WANDB cloud auth + the Surge render plugin/worker, unavailable here. The mechanism above is proven offline; this would confirm it on the real worker path.)*
  To run manually:
  ```bash
  # On a wandb-authenticated host with the render plugin:
  uv run synth-setter-generate-dataset experiment=generate_dataset/smoke-shard logger=wandb
  # Then open the run in wandb and confirm the Logs tab shows the loguru lines
  # AND the per-shard generate_vst_dataset.py subprocess output.
  ```

**Summary:** the load-bearing promise (fd-level capture of loguru + subprocess output) is verified end-to-end offline; config-pin, wandb-value-validity, and eval-composition are all green. Only the full cluster-path smoke remains, which requires infra not available here.

Issue #1475, opened 2026-06-05
chore(code-health): clean up smell-checker findings (print/structlog, error-cause, leaked handle, hardcoded paths, test smells)


ktinubu:
Tracking bug for code-smell cleanups surfaced by a `/smell-checker` audit of `origin/main` (`9c16c71`). Each item is anchored to `path:line` and verified. Grouped into batches sized for one PR each. `Part of #879`.

### Batch A — Hygiene: `print()` → structured logging — ✅ #1476 (merged)
- [x] `src/synth_setter/pipeline/data/stats.py:165` — 9 `print()` calls → module `logging` logger (file already standardized on stdlib `logging`). [LOW]
- [x] `src/synth_setter/utils/callbacks.py:316` — `print("plotting")` debug line deleted. [LOW]

### Batch B — Error handling: re-raise loses cause — ✅ #1477 (merged)
- [x] `src/synth_setter/cli/generate_dataset.py:348` — root cause was `r2_io.object_size`'s context-free `int()` parse on the failure path; now re-raises `RuntimeError(...) from exc` naming the URI. (The `:348` bare `raise` already preserved context.) [MED]

### Batch C — Resources & typing — 🔄 #1478 (open, green, mergeable)
- [ ] `src/synth_setter/data/surge_datamodule.py:49` — h5py handle stored on `self`, closed only in `teardown()`; added `weakref.finalize` safety-net close. [MED]
- [ ] `src/synth_setter/pipeline/schemas/spec.py:715` — annotated `_default_r2_location -> R2Location`; removed `# type: ignore[arg-type]`. [MED]
- [ ] `src/synth_setter/pipeline/ci/validate_shard.py:228` — replaced unchecked `cast` with an `isinstance` guard that appends a structured error naming the member. [LOW]

### Batch E — Test smells — ✅ #1479 (merged)
- [x] `tests/data/vst/test_preset_coverage.py:59` — sleep-for-editor → `threading.Event` via `warmup_plugin`. [HIGH]
- [x] `tests/data/vst/test_core.py:224` — `time.sleep(0.05)` → two-event handshake. [HIGH]
- [x] `tests/test_callbacks.py:43` — mock-only `log_image` → real `_log_figure` dispatch against recording loggers. [HIGH]
- [x] `tests/data/vst/test_core.py:98` — `assert_called_once` → observable warmup effect via recording fake plugin. [HIGH]
[...]

PR #1477, opened 2026-06-06
chore(r2): preserve exception cause on generate_dataset failure path

ktinubu:
## What

On the `generate` dataset failure path, `r2_io.object_size` parsed `rclone lsf --format=s` stdout with a bare `int(out)`. Non-integer stdout (an unexpected listing, garbage, a protocol change) raised a context-free `ValueError: invalid literal for int()` that named neither the probed URI nor the rclone payload. As that error propagated up through `generate()`'s `except BaseException` / `finally`, the cause was effectively lost — the operator saw only the literal-parse failure with no provenance.

The anchored line (`generate_dataset.py:348`) is a bare `raise` that already preserves context; the genuine cause-loss reachable on that failure path is this implicit `int()` parse in the R2 probe each shard calls via `_render_one_owned_shard`.

## Change

- Guard the `int(out)` parse in `object_size`; on `ValueError`, re-raise a contextual `RuntimeError` (naming the probed URI and the offending payload) chained via `from exc`, preserving the original `ValueError` as `__cause__`.
- Keep the empty-stdout → `None` and zero-size → `0` semantics unchanged.
- Document the new `:raises RuntimeError:` contract.

## Test

`test_non_integer_stdout_raises_with_original_cause` (added first, Red→Green) asserts the contextual message includes the probed URI and that `excinfo.value.__cause__` is the original `ValueError`.

Refs #1475

### DIFF
--- a/src/synth_setter/pipeline/r2_io.py
+++ b/src/synth_setter/pipeline/r2_io.py
@@ -333,6 +333,9 @@ def object_size(r2_uri: str) -> int | None:
 
     A zero-size object exists and returns ``0``. Callers that want to treat zero-size as absent
     (e.g. defending against half-uploaded objects) test ``size and size > 0`` themselves.
+
+    :raises RuntimeError: stdout is non-empty but not an integer; chained to the
+        underlying ``ValueError`` so the probed URI survives the failure path.
     """
     args = [  # noqa: S607 — rclone resolved by image's PATH
         "rclone",
@@ -344,7 +347,16 @@ def object_size(r2_uri: str) -> int | None:
         args, check=True, capture_output=True, text=True
     )
     out = result.stdout.strip()
-    return int(out) if out else None
+    if not out:
+        return None
+    try:
+        return int(out)
+    except ValueError as exc:
+        # Chain the parse failure so the probed URI and rclone payload survive
+        # the generate failure path instead of a bare ``invalid literal``.
+        raise RuntimeError(
+            f"rclone lsf --format=s returned unparsable size {out!r} for {r2_uri}"
+        ) from exc
 
 
 def purge_prefix(bucket: str, prefix: str) -> None:
--- a/tests/pipeline/test_r2_io.py
+++ b/tests/pipeline/test_r2_io.py
@@ -565,6 +565,19 @@ class TestObjectSize:
             with pytest.raises(subprocess.CalledProcessError):
                 r2_io.object_size("r2://bucket/key.h5")
 
+    def test_non_integer_stdout_raises_with_original_cause(self) -> None:
+        """Unparsable rclone stdout raises a contextual error chained to the ValueError.
+
+        Guards the generate failure path: ``int(out)`` on garbage stdout would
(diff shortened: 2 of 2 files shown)


ktinubu:
## Verification Results

Behavioral verification of the cause-preservation fix. The renderer/rclone/R2 backends are unreachable in this sandbox, so subprocess output is stubbed; every check invokes the real `object_size` / `generate` code, not a parsed config.

- [x] **Non-integer rclone stdout raises a contextual `RuntimeError` chained to the original `ValueError`** *(Level 2 — promise is function behavior at the R2 probe; invoked `object_size` directly with stubbed `subprocess.run`)*
  ```bash
  $ uv run python3 -c "<stub subprocess.run -> 'garbage-not-int'; call object_size('r2://my-bucket/data/shard-000007.h5')>"
  PASS promise-1: RuntimeError names URI + payload, __cause__ is the original ValueError
    message : rclone lsf --format=s returned unparsable size 'garbage-not-int' for r2://my-bucket/data/shard-000007.h5
    __cause__: ValueError("invalid literal for int() with base 10: 'garbage-not-int'")
  ```
  The message names the probed URI and the offending payload; `__cause__` is the original `ValueError` (cause chain preserved via `from exc`).

- [x] **Existing `object_size` semantics are unchanged** *(Level 2 — guards against a behavior regression in the happy paths the fix wraps)*
  ```bash
  $ uv run python3 -c "<stub stdout '12345' -> 12345; '' -> None; '0' -> 0>"
  PASS promise-2: integer->int, empty->None, zero->0 unchanged
  ```

- [x] **Environmental failures still propagate as `CalledProcessError`** *(Level 2 — adversarial: the new `except ValueError` must not widen to swallow rclone's non-zero exit)*
  ```bash
  $ uv run python3 -c "<stub subprocess.run side_effect=CalledProcessError; call object_size>"
  PASS promise-3: rclone non-zero exit still propagates as CalledProcessError
  ```

- [x] **Cause survives end-to-end through `generate()`'s `except BaseException` / `finally`** *(Level 1 — the promise is end-to-end: the probed-size error must reach the caller with its cause intact after traversing the dispatch + finalize path)*
  ```bash
  $ uv run python3 -c "<build a 1-shard DatasetSpec; stub extract_renderer_version + r2_io.subprocess.run -> 'not-a-size'; call generate(spec, tmpdir, [])>"
  PASS Level-1: error propagated through generate() failure path with cause intact
    top    : rclone lsf --format=s returned unparsable size 'not-a-size' for r2://b/data/cause-e2e/cause-e2e-20260606T000000000Z/shard-000000.h5
    __cause__: ValueError("invalid literal for int() with base 10: 'not-a-size'")
  ```
  Confirms the original `ValueError` is retained as `__cause__` after propagating through `_render_one_owned_shard` -> dispatch -> `generate`'s `except`/`finally` to the caller — the exact chain this PR set out to preserve.

All four checks passed. The added unit test `tests/pipeline/test_r2_io.py::TestObjectSize::test_non_integer_stdout_raises_with_original_cause` pins this contract in CI.

PR #1498, opened 2026-06-06
feat(storage): log dataset W&B artifact with R2 refs on finalize

ktinubu:
## Summary

Logs the canonical `data-{dataset_config_id}` W&B artifact (type `dataset`) when the data pipeline finalizes, making finalize the **producer node of the W&B lineage DAG** (storage-provenance-spec §4–6). Closes the Tier-2 head of the provenance work tracked in #1467.

## What changed

- **`build_dataset_artifact(spec)`** (`cli/finalize_dataset.py`) — pure constructor: name `data-{config_id}`, type `dataset`, `add_reference(s3://…, checksum=False)` for each finalized R2 object (per-split `.h5` + `stats.npz` for hdf5; the shard prefix + `stats.npz` for wds), and `metadata = {shard_count, n_samples, git_sha}`. `r2://`→`s3://` per spec §4; `checksum=False` because R2's custom S3 endpoint isn't reachable by W&B's reference handler (lineage-only reference).
- **`_log_dataset_artifact(loggers, spec)`** — best-effort logging to any `WandbLogger`, mirroring `_log_spec_artifact` (a wandb failure warns, never aborts a completed finalize).
- **`finalize(cfg)`** now composes a wandb logger pinned to `spec.run_id` with `job_type=data-generation` + `resume=allow`, so the artifact attaches to the data-generation run. `finalize_dataset.yaml` gains `- logger: wandb`, and `finalize-dataset.yaml` forwards the inherited `WANDB_API_KEY` so the live `generate→finalize` chain actually logs it. WandbLogger inits lazily + logging is best-effort, so a finalize without `WANDB_API_KEY` degrades to a no-op rather than aborting.
- **Refactor:** extracted the logger-finalize + guarded `wandb.finish()` dance from `generate_dataset._close_loggers` into a public `utils.instantiators.close_loggers`, reused by both data-pipeline entrypoints (behavior-preserving).
- **Chore:** synced `uv.lock` to 8.22.0 (the `[skip ci]` release bumped `pyproject.toml` but not the lock), unblocking CI's lock-check.

## Testing

- `tests/.../test_finalize_dataset_artifact.py` — asserts on a **real** `wandb.Artifact` (name/type/refs/metadata, empty-split omission, hdf5-vs-wds reference shapes) + the `_r2_to_s3_uri` non-`r2://` guard.
[...]

PR #1494, opened 2026-06-06
feat(storage): R2 prefix write-time assertion helper

ktinubu:
## Summary

- Adds `assert_r2_prefix_matches()` to `pipeline/schemas/prefix.py` — a write-time guard that verifies a materialized `R2Location.prefix` matches what `make_r2_prefix()` would derive from the same `config_id`, `run_id`, and `prefix_root`.
- Calls it in `finalize_from_spec()` after the idempotency probe, before any directory creation or R2 writes, so a drifted prefix aborts cleanly.
- 6 new tests in `TestAssertR2PrefixMatches` covering match, custom root, all three mismatch axes, and error-message content.

## Test plan

- [ ] `uv run pytest tests/pipeline/schemas/test_prefix.py tests/pipeline/entrypoints/test_finalize_dataset.py -x -q` — 41 passed
- [ ] `make format` — all hooks passed

Closes #281
Refs #1467

Issue #1471, opened 2026-06-05
feat(data-pipeline): log canonical dataset W&B artifact with R2 references on finalize


ktinubu:
## Summary

Log a canonical `dataset` W&B artifact (name `data-{dataset_config_id}`, type `dataset`) when the data pipeline finalizes, with `add_reference("s3://...")` pointing at the finalized R2 objects (`train/val/test.h5`, `stats.npz`). Per `storage-provenance-spec.md` §4–5.

## Current state

`cli/generate_dataset.py:355` logs only a `*-input-spec` metadata artifact. The finalized splits are pushed to R2 with no W&B artifact wrapper, so there is nothing for training to consume as lineage.

## Deliverables

- [ ] On finalize, create `wandb.Artifact(name=f"data-{dataset_config_id}", type="dataset")`
- [ ] `artifact.add_reference(<R2 uri of splits/stats>)` (R2 prefix from existing `R2Location`/`make_r2_prefix()`)
- [ ] `artifact.metadata` carries shard_count, n_samples, git_sha
- [ ] Tests assert artifact name/type/references (offline/mocked W&B)

## Dependency

Producer end of the lineage DAG — must land before training's `use_artifact()`. Refs #120, #1470, #1467.

Issue #281, opened 2026-03-24
Task 5.1: R2 prefix assertion helper


ktinubu:
Write a small helper or assertion that constructs the R2 path from config IDs and checks it matches what's being written.

Expected prefixes per spec §3:
- `data/{dataset_config_id}/{dataset_wandb_run_id}/`
- `train/{dataset_config_id}/{dataset_wandb_run_id}/{train_config_id}/{train_wandb_run_id}/`
- `eval/{dataset_config_id}/{dataset_wandb_run_id}/{train_config_id}/{train_wandb_run_id}/{eval_config_id}/{eval_wandb_run_id}/`

PR #1510, opened 2026-06-06
fix(storage): make finalize R2 prefix check advisory, not fatal

ktinubu:
## Summary

Makes the finalize-time R2 prefix check **advisory** (warn, never abort). The merged #281 guard (`assert_r2_prefix_matches` in `finalize_from_spec`) aborted finalize whenever `spec.r2.prefix` diverged from the canonical `data/{task_name}/{run_id}/` shape — but a custom prefix is legitimate, so the guard could not tell intentional override from drift.

## Why this is needed

`generate-dataset-shards.yaml`'s unconditional finalize chain runs the **"Generate dataset shards → real R2 (Docker + VST)"** job, whose `test_oracle_eval_inline/shuffled_audio_metrics` cases isolate their R2 objects under `test-runs/<test>/<uuid>/`. The #281 guard rejected that prefix and aborted finalize, turning the job red. (#1494's merge commit never ran this job, so the regression landed undetected; it surfaced on #1498's CI.)

finalize reads the **same** prefix generate wrote to, so the spec is self-consistent — a non-canonical prefix is not a data-integrity risk. Marker-last ordering, idempotency, and the format dispatch are unchanged.

## Changes

- `finalize_from_spec`: wrap the `assert_r2_prefix_matches` call in `try/except ValueError` → log a warning and proceed. The helper and its `test_prefix.py` unit tests are untouched (still strict for callers that want it).
- Replace the obsolete `test_finalize_from_spec_drifted_prefix_raises_before_any_upload` (which pinned the now-removed abort, landed via #1498) with `test_finalize_from_spec_non_canonical_prefix_warns_and_proceeds` — seeds shards at a custom prefix on the real `fake_r2_remote` rclone backend and asserts finalize both **warns** and **lands** stats + marker there.
- `doc-map.yaml`: describe the advisory behavior.
- `uv.lock`: sync to 8.23.1 (the release bumped `pyproject.toml` but not the lock).

## Follow-up (not in scope)

A custom-prefix finalize still logs the W&B artifact under the canonical `data-{task_name}` name; namespacing/skipping artifact logging for non-canonical prefixes is a candidate follow-up (latent — the e2e runs without `WANDB_API_KEY`).

Refs #281, #1471

### DIFF
--- a/docs/doc-map.yaml
+++ b/docs/doc-map.yaml
@@ -81,7 +81,7 @@ docs:
       - pattern: "src/synth_setter/configs/finalize_dataset.yaml"
         covers: "Finalize @hydra.main entry config; required `dataset_spec_uri` loads the persisted DatasetSpec from R2 (file://, r2://, or bare path); overrides `hydra.run.dir` because `task_name`/`run_name` are not in this cfg; composes the `logger: wandb` group so finalize logs the dataset artifact (best-effort no-op without `WANDB_API_KEY`)"
       - pattern: "src/synth_setter/cli/finalize_dataset.py"
-        covers: "Finalize entrypoint: loads the frozen DatasetSpec, dispatches on `spec.output_format` (wds → `finalize_wds`, hdf5 → `finalize_hdf5`), asserts the R2 prefix before any write, writes the `dataset.complete` marker last for idempotency, then logs the canonical `data-{config_id}` dataset W&B artifact with s3:// R2 references"
+        covers: "Finalize entrypoint: loads the frozen DatasetSpec, dispatches on `spec.output_format` (wds → `finalize_wds`, hdf5 → `finalize_hdf5`), warns (never aborts) on a non-canonical R2 prefix, writes the `dataset.complete` marker last for idempotency, then logs the canonical `data-{config_id}` dataset W&B artifact with s3:// R2 references"
       - pattern: ".github/workflows/generate-dataset-shards.yaml"
         covers: "Reusable launcher fan-out across providers (skypilot-local kind, runpod, oci); per-provider artifact upload; emits a `spec_uri` workflow output (the materialized spec's canonical under-prefix `input_spec.json` URI, computed by the `synth-setter-spec-uri` console script — see `src/synth_setter/pipeline/ci/spec_uri.py`) consumed by validate-dataset-shards"
       - pattern: ".github/workflows/validate-dataset-shards.yaml"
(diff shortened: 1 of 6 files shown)


ktinubu:
## Verification Results

**PR #1510 — fix(storage): make finalize R2 prefix check advisory, not fatal**
Behavior verified at the fix commit (`729ad0a6`); the later `6c68669d` is docs-only.

- [x] **The actual previously-broken oracle-eval e2e now finalizes a custom-prefix run** *(Level 1 — the real generate→finalize chain with a real render + real R2 and the `test-runs/<test>/<uuid>/` prefix that the #281 guard rejected; this is the exact "Generate dataset shards → real R2 (Docker + VST)" failure)*
  ```bash
  $ make link-plugins   # worktree needs the plugins/ symlink the Docker image has
  $ timeout 540 uv run pytest "tests/test_generate_dataset.py::test_oracle_eval_inline_writes_bounded_audio_metrics" -p no:randomly -m "" -v 2>&1 | tail -4
  tests/test_generate_dataset.py::test_oracle_eval_inline_writes_bounded_audio_metrics PASSED [100%]
  235.46s call     tests/test_generate_dataset.py::test_oracle_eval_inline_writes_bounded_audio_metrics
  1 passed in 236.89s (0:03:56)
  ```
  Before the fix this run aborts with `ValueError: R2 prefix mismatch: got 'test-runs/…' expected 'data/smoke-shard/…'`. After it, generate renders + uploads and finalize completes against real R2 (VST + R2 both available in this devcontainer; probed `ls "$SYNTH_SETTER_PLUGIN_PATH"` and `rclone lsd r2:`).

- [x] **finalize warns and proceeds on a non-canonical prefix (both halves of the contract)** *(Level 2 — drives the real `finalize_from_spec` SUT against the `fake_r2_remote` rclone backend; asserts the warning fires AND stats + marker land at the custom prefix)*
  ```bash
  $ uv run pytest "tests/pipeline/entrypoints/test_finalize_dataset.py::test_finalize_from_spec_non_canonical_prefix_warns_and_proceeds" -p no:randomly -q 2>&1 | tail -2
  1 passed in 1.40s
  ```

- [x] **A canonical prefix is accepted with no warning (no false-positive regression)** *(Level 2 — adversarial; invokes the guard helper on a canonical spec and confirms no `ValueError` → the advisory warning only fires on genuine divergence)*
  ```bash
  $ uv run python3 -c "
  from pipeline.entrypoints.test_finalize_dataset import _build_wds_smoke_spec
  from synth_setter.pipeline.schemas.prefix import make_r2_prefix, assert_r2_prefix_matches
  spec = _build_wds_smoke_spec(task_name='canonical-nowarn')
  assert spec.r2.prefix == make_r2_prefix(spec.task_name, spec.run_id)
  assert_r2_prefix_matches(spec.r2.prefix, spec.task_name, spec.run_id, spec.r2.prefix_root)
  print('canonical prefix accepted, no ValueError, no warning: PASS')
  "
  canonical prefix accepted, no ValueError, no warning: PASS
  ```

- [x] **The `assert_r2_prefix_matches` helper is unchanged — still raises on a mismatch** *(Level 2 — the fix only changes the finalize call site to advisory; the helper's strict contract is intact for any caller that wants it)*
  ```bash
  $ uv run pytest "tests/pipeline/schemas/test_prefix.py::TestAssertR2PrefixMatches" -p no:randomly -q 2>&1 | tail -2
  6 passed in 0.34s
  ```

All checks pass. The hotfix unblocks the Docker+VST job that #281's guard broke, without weakening the helper's strict contract.

PR #1519, opened 2026-06-06
docs(agents): add import-ordering rule to avoid F401 autofix churn

ktinubu:
## Problem

Agents repeatedly lose a cycle to ruff's `F401` autofix: an import is added in
one edit, `make format` (pre-commit `ruff --fix`) runs before the using code
lands, the momentarily-unused import is deleted, and the agent has to re-add it
once the usage exists. This recurs across sessions and drags on velocity.

## Change

Document the ordering rule so every agent avoids the trap:

> Add an import in the same edit as its first use, or add imports last — ruff's
> `F401` autofix deletes an import that is momentarily unused if `make format`
> runs before the using code lands, costing a re-add cycle.

- **AGENTS.md** (canonical) — added to the "Writing code" section.
- **CLAUDE.md** — mirrored into the "writing or modifying Python code" block so
  the two stay in sync (no generator exists between them).

Docs-only; no Python/CI changes.

## Relationship to #1431

`Refs #1431`. That issue stops VS Code's on-save ruff from deleting unused
imports but explicitly leaves the CLI / pre-commit (`make format`) path
unchanged. This PR documents the guidance for exactly that path — the
complementary half of the same F401 friction.

Refs #1431

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -41,6 +41,9 @@ Architecture: [docs/architecture.md](docs/architecture.md).
   from R2, worker reports). Dataclasses for internal typed containers.
 - `structlog` in pipeline code; stdlib `logging` elsewhere.
 - All `rclone` operations use `--checksum`.
+- Add an import in the same edit as its first use, or add imports last —
+  ruff's `F401` autofix deletes an import that is momentarily unused if
+  `make format` runs before the using code lands, costing a re-add cycle.
 - Run `make format` before committing. Pre-commit (ruff, ruff-format,
   pydoclint, prettier, mdformat, gitlint) is authoritative; suppressing
   rules to make CI green is forbidden — see
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -29,6 +29,7 @@ Never run `make docker-*` or RunPod commands without asking — they spend money
 - Pydantic `BaseModel(strict=True)` at trust boundaries (config parsing, JSON from R2, worker reports); dataclasses for internal typed containers.
 - `structlog` in pipeline code; stdlib `logging` elsewhere.
 - All `rclone` operations use `--checksum`.
+- Add an import in the same edit as its first use, or add imports last — ruff's `F401` autofix deletes an import that is momentarily unused if `make format` runs before the using code lands, costing a re-add cycle.
 
 </important>
 


ktinubu:
## Verification Results

This is a docs-only PR, so the one behavioral promise to verify is the *technical claim* the new rule makes: **ruff's `F401` autofix deletes a momentarily-unused import (the exact behavior `make format` applies via the pre-commit `ruff --fix` hook), and an import that is actually used survives.** Both directions reproduced below.

- [x] **An unused import is deleted by ruff's F401 autofix** *(Level 2 — invokes ruff directly with the same F401 rule `make format` applies; the promise is tool behavior, not end-to-end)*
  ```bash
  $ printf 'from omegaconf import OmegaConf\n\n\ndef noop() -> None:\n    pass\n' > _f401_demo.py
  $ uv run ruff check --select F401 --fix _f401_demo.py
  Found 1 error (1 fixed, 0 remaining).
  $ cat _f401_demo.py
  

  def noop() -> None:
      pass
  ```
  The import line is gone — exactly the churn the rule warns about.

- [x] **An import that is used survives the same autofix** *(Level 2 — converse case; proves the rule's remedy works: add the usage and the import sticks)*
  ```bash
  $ printf 'from omegaconf import OmegaConf\n\n\ndef load() -> object:\n    return OmegaConf.create({})\n' > _f401_demo.py
  $ uv run ruff check --select F401 --fix _f401_demo.py
  All checks passed!
  $ cat _f401_demo.py


  def load() -> object:
      return OmegaConf.create({})
  ```
  With the usage present, ruff leaves the import — confirming "add the import with its first use" resolves the problem.

- [x] **Config wiring matches the claim** *(Level 5 — supporting evidence, not the primary check)*: ruff `F` is in `select` (`pyproject.toml`), pre-commit runs `ruff` with `args: [--fix]` (`.pre-commit-config.yaml`), and `make format` runs pre-commit — so the reproduction above is the path real edits hit.

**Not applicable / out of scope:**
- No Level 1 end-to-end check exists — the change is two markdown lines of agent guidance; there is no pipeline or runtime contract to exercise. The behavioral reproduction above is the strongest meaningful check.
- `make format` / mdformat / test-suite passing are CI's job and intentionally omitted per the skill's noise-reduction rule. (mdformat/prettier/codespell did pass locally on the diff.)

Temp file `_f401_demo.py` was removed after the runs; working tree is clean.

PR #1522, opened 2026-06-06
ci(vst-slow): raise VST Slow Tests job timeout to 120 min

### LINKED ISSUE #1521: fix(ci): VST Slow Tests chronically cancelled at 60-min job timeout
## Problem

The `VST Slow Tests` workflow (`.github/workflows/test-vst-slow.yml`, job `run_vst_slow_tests`) is configured with `timeout-minutes: 60`. The slow audio-similarity suite routinely runs longer than that and is **cancelled at the timeout** rather than failing on an assertion.

This is chronic on `main`: **3 of the last 5 `main` runs** of the workflow were cancelled at the 60-min mark (2026-06-06 00:28, 2026-06-05 23:46, 2026-06-05 23:10). It also surfaces as a red (non-required) check on unrelated PRs.

## Fix

Raise `timeout-minutes` from 60 to 120 so the suite has room to finish. The job is non-required and runs on `ubuntu-latest-4core`; the longer budget only applies when the suite actually needs it.

## Acceptance

- `run_vst_slow_tests` completes within the budget instead of being cancelled at 60 min.

ktinubu:
## What

Raise `timeout-minutes` from 60 to 120 on the `run_vst_slow_tests` job in `.github/workflows/test-vst-slow.yml`.

## Why

The `VST Slow Tests` workflow chronically exceeds the 60-min job timeout and is **cancelled** rather than failing on an assertion — **3 of the last 5 `main` runs** were cancelled at the 60-min mark. The cancellation also surfaces as a red (non-required) check on unrelated PRs. The job runs on `ubuntu-latest-4core`; the larger budget only applies when the slow audio-similarity suite actually needs it.

## Scope

Single-line value change. No renderer, test, or benchmark code is touched.

Fixes #1521

### DIFF
--- a/.github/workflows/test-vst-slow.yml
+++ b/.github/workflows/test-vst-slow.yml
@@ -78,7 +78,7 @@ jobs:
   run_vst_slow_tests:
     name: VST slow tests (Docker)
     runs-on: ubuntu-latest-4core
-    timeout-minutes: 60
+    timeout-minutes: 120
 
     # Job-level grant so only this job can push to ``gh-pages`` for the
     # benchmark history. Workflow-level default is ``contents: read``.


ktinubu:
## Verification Results

PR promise: the `run_vst_slow_tests` job gets a **120-min** budget so the slow VST suite finishes instead of being **cancelled at 60 min**.

- [ ] **Live VST run completes within budget, not cancelled at 60 min** *(Level 1 — promise is end-to-end; IN PROGRESS)*
  The push triggered the workflow on this PR (it path-matches `.github/workflows/test-vst-slow.yml`). It runs under the new 120-min budget:
  ```bash
  $ gh pr checks 1522 | grep -i vst
  VST slow tests (Docker)	pending	0	https://github.com/tinaudio/synth-setter/actions/runs/27067263236/job/79890187178
  ```
  Cannot tick yet — the run takes 60–120 min. A PASS here means it reaches a real terminal state (success/failure on assertions) rather than a 60-min cancellation. I cannot re-trigger it myself (token lacks Actions `workflow_dispatch`/re-run); this PR run is the live check.

- [x] **GitHub accepts the modified workflow (no schema/syntax break)** *(Level 2 — ran the validator the repo uses, not a YAML parse)*
  ```bash
  $ uv run pre-commit run check-github-workflows --files .github/workflows/test-vst-slow.yml
  ```

- [x] **The 120 value lands on the correct job (`run_vst_slow_tests`), not a step or another job** *(Level 5 — confirms the value is on the right key; the behavioral budget is the Level 1 check above)*
  ```bash
  $ python3 -c "import yaml; j=yaml.safe_load(open(\".github/workflows/test-vst-slow.yml\"))[\"jobs\"][\"run_vst_slow_tests\"]; print(j[\"name\"], j[\"timeout-minutes\"]); assert j[\"timeout-minutes\"]==120"
  VST slow tests (Docker) 120
  ```

- [x] **No other timeout overrides it** — `timeout-minutes` appears exactly once in the file (the job-level key); there are no step-level timeouts that would still cap the run at a lower value.
  ```bash
  $ grep -c "timeout-minutes" .github/workflows/test-vst-slow.yml
  1
  ```
- [x] **Benchmark-publish gating is untouched** — the diff is a single value line; the `push`-to-main / `workflow_dispatch` `publish_metrics` gates on the gh-pages publish step are unchanged, so PR runs still never publish.

**Net:** structural correctness is verified; the end-to-end "not cancelled at 60 min" promise is pending the live PR run above — re-check it when run 27067263236 finishes.

PR #1524, opened 2026-06-06
test(testing): clamp pytest-xdist -n auto by available memory

### LINKED ISSUE #1520: bug(testing): -n auto OOM-kills test-fast on shared host (exit 143)
## Symptom

`make test-fast` (`pytest -n auto`) is terminated with exit code 143 (SIGTERM / OOM-killer) on the shared dev host running 60+ worktrees.

## Root cause

The `pytest_xdist_auto_num_workers` hook in `tests/conftest.py` (the cgroup-aware CPU clamp) is **CPU-quota-only**. On the shared host neither `cpu.max` nor `memory.max` is set, so `sched_getaffinity` returns all 32 host cores and `-n auto` spawns 32 torch-importing workers. Aggregate worker RSS — compounded by other concurrent worktree runs — exhausts host RAM and the kernel OOM-killer reaps the process group.

The earlier CPU-aware clamp only addresses a *single container over-subscribing its own CPU quota*; it cannot see host memory pressure or cross-process contention.

## Fix

- Add a memory-aware clamp: `min(host MemAvailable, cgroup memory limit)` divided by a per-worker budget (`PYTEST_XDIST_WORKER_MEM_MB`, default 1 GiB). Hook returns `min(cpu_workers, mem_workers)`.
- Add a low `PYTEST_XDIST_AUTO_NUM_WORKERS` default to the devcontainer configs so every worktree on the shared host defaults low (per-run override preserved).

Sibling/follow-on to the CPU-aware clamp work.

ktinubu:
## Problem

`make test-fast` (`pytest -n auto`) is OOM-killed (SIGTERM, exit 143) on the shared dev host running 60+ worktrees.

The existing `pytest_xdist_auto_num_workers` hook is **CPU-quota-only**. On the shared host neither `cpu.max` nor `memory.max` is set, so `sched_getaffinity` returns all 32 host cores and `-n auto` spawns 32 torch-importing workers. Aggregate worker RSS — compounded by other concurrent worktree runs — exhausts host RAM and the kernel OOM-killer reaps the run. The CPU clamp can only catch a single container over-subscribing its own CPU quota; it can't see host memory pressure.

## Change

- **Memory-aware clamp** in `tests/conftest.py`: `min(host MemAvailable, cgroup memory limit)` ÷ a per-worker budget (`PYTEST_XDIST_WORKER_MEM_MB`, default 1 GiB). The hook now returns `min(cpu_workers, mem_workers)`, falling back to CPU-only when no memory signal is readable. Mirrors the v2-then-v1 structure of the CPU helper; the v1 unlimited sentinel and `memory.max=max` both degrade to "no cap".
- **Low default** `PYTEST_XDIST_AUTO_NUM_WORKERS: "4"` added to all three devcontainer `containerEnv` blocks (cpu/gpu/root_gpu) so every worktree on the shared host defaults low; the per-run override is preserved.

### Interaction note

The hook checks `PYTEST_XDIST_AUTO_NUM_WORKERS` **first**, so the devcontainer pin short-circuits the memory clamp *inside the devcontainer*. That's intentional: the fixed low pin is the shared-host default, and the adaptive memory clamp is the safety net for every environment that doesn't set the var (CI runners, a future memory-capped container). CI does not consume `devcontainer.json` `containerEnv`, so CI parallelism is unaffected and benefits from the adaptive clamp.

### Notes

- `uv.lock` carries a one-line `8.23.1 → 8.23.2` version sync: the `uv-lock` pre-commit hook auto-corrected pre-existing drift between main's release commit and its lockfile. Unavoidable (reverting just makes the hook redo it) and correct.

## Tests

`tests/test__conftest_mem.py` — 20 new tests covering host/cgroup `min` selection (both directions, v1/v2), the v1 unlimited sentinel and `0` boundary, v2-malformed→v1 fallthrough, fractional floor-to-1, env-budget override and invalid-budget fallback, and the hook-level `min(cpu, memory)` composition with the env hard-override. The 10 existing CPU-hook tests still pass (30 total).

## Verification

Live on this host the clamp drops `-n auto` from 32 → 24 workers (24.6 GiB avail ÷ 1 GiB); with the devcontainer pin it is 4.

Fixes #1520

### DIFF
--- a/.devcontainer/cpu/devcontainer.json
+++ b/.devcontainer/cpu/devcontainer.json
@@ -16,7 +16,8 @@
     ".env"
   ],
   "containerEnv": {
-    "MODE": "idle"
+    "MODE": "idle",
+    "PYTEST_XDIST_AUTO_NUM_WORKERS": "4"
   },
   "remoteUser": "${localEnv:DEVCONTAINER_USER:dev}",
   "mounts": [
--- a/.devcontainer/gpu/devcontainer.json
+++ b/.devcontainer/gpu/devcontainer.json
@@ -19,7 +19,8 @@
     ".env"
   ],
   "containerEnv": {
-    "MODE": "idle"
+    "MODE": "idle",
+    "PYTEST_XDIST_AUTO_NUM_WORKERS": "4"
   },
   "remoteUser": "${localEnv:DEVCONTAINER_USER:dev}",
   "mounts": [
--- a/.devcontainer/root_gpu/devcontainer.json
+++ b/.devcontainer/root_gpu/devcontainer.json
@@ -21,7 +21,8 @@
     ".env"
   ],
   "containerEnv": {
-    "MODE": "idle"
+    "MODE": "idle",
+    "PYTEST_XDIST_AUTO_NUM_WORKERS": "4"
   },
   "remoteUser": "${localEnv:DEVCONTAINER_USER:root}",
   "updateRemoteUserUID": true,
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -919,20 +919,107 @@ def _cgroup_aware_cpu_count() -> int:
     return max(1, int(cpus))
 
 
+# Per-worker resident-memory budget for the -n auto memory clamp, overridable
+# via PYTEST_XDIST_WORKER_MEM_MB. 1 GiB suits a torch+h5py+Hydra worker.
+_DEFAULT_WORKER_MEM_MB = 1024
+
+# A v1 memory.limit_in_bytes at or above this is the kernel's unlimited sentinel.
+_MEM_UNLIMITED_SENTINEL = 1 << 62
+
+
+def _meminfo_available_bytes() -> int | None:
+    """Read ``MemAvailable`` from ``/proc/meminfo`` as bytes.
+
+    :returns: Host available memory in bytes, or None if the field is unreadable.
+    """
+    try:
+        with open("/proc/meminfo") as fh:
+            for line in fh:
+                if line.startswith("MemAvailable:"):
+                    return int(line.split()[1]) * 1024  # field is in kibibytes
+    except (OSError, ValueError, IndexError):
+        pass
+    return None
+
+
+def _cgroup_memory_limit_bytes() -> int | None:
+    """Read the cgroup memory limit in bytes, honouring v2 then v1.
+
+    :returns: The cgroup memory cap in bytes, or None when unset or unlimited.
+    """
+    try:  # cgroup v2: unified hierarchy, kernel >= 4.5
+        with open("/sys/fs/cgroup/memory.max") as fh:
+            limit_field = fh.read().split()[0]
+            if limit_field != "max":
+                return int(limit_field)
+            return None
+    except (OSError, ValueError, IndexError):
+        pass
+    try:  # cgroup v1: legacy per-subsystem hierarchy
+        with open("/sys/fs/cgroup/memory/memory.limit_in_bytes") as fh:
+            limit = int(fh.read())
+        if 0 < limit < _MEM_UNLIMITED_SENTINEL:
+            return limit
+    except (OSError, ValueError):
+        pass
+    return None
+
+
+def _available_memory_bytes() -> int | None:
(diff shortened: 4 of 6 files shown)


ktinubu:
## Verification Results

Verified on this host (32 cores, ~20–23 GiB available). The PR's promise is that `pytest -n auto` workers are clamped by `min(cpu, memory)`, with `PYTEST_XDIST_AUTO_NUM_WORKERS` as a hard override floored to 1. Checks lead with the end-to-end xdist integration, not just the unit helpers.

- [x] **`pytest -n auto` actually spawns the hook's clamped count** *(Level 1 — exercises the real pytest-xdist → `pytest_xdist_auto_num_workers` path end-to-end; lower levels only prove the function returns a number, not that xdist consults it)*
  ```bash
  $ PYTEST_XDIST_AUTO_NUM_WORKERS=5 uv run pytest tests/test__conftest_cpu.py -n auto -q | grep -iE "workers \["
  5 workers [10 items]
  ```
  xdist honors the hook's return value verbatim.

- [x] **The memory-budget knob drives the real worker count** *(Level 1 — proves the memory path is live in an actual run, not just unit-tested)*
  ```bash
  $ PYTEST_XDIST_WORKER_MEM_MB=8000 uv run pytest tests/test__conftest_cpu.py -n auto -q | grep -iE "workers \["
  2 workers [10 items]
  $ PYTEST_XDIST_WORKER_MEM_MB=2000 uv run pytest tests/test__conftest_cpu.py -n auto -q | grep -iE "workers \["
  11 workers [10 items]
  ```
  Cross-check — the hook computes the same values it spawns: `budget=8000MB -> hook=2`, `budget=2000MB -> hook=11`. The clamp tracks available memory ÷ budget.

- [x] **Invalid budget degrades to the 1 GiB default instead of crashing** *(Level 1 — adversarial: malformed env must not abort collection)*
  ```bash
  $ PYTEST_XDIST_WORKER_MEM_MB=0 uv run pytest tests/test__conftest_cpu.py -n auto -q | grep -iE "workers \["
  20 workers [10 items]
  ```
  `0` is rejected and the default divisor applies (≈20 GiB ÷ 1 GiB), no error.

- [x] **`min(cpu, memory)` clamp is active on this shared host** *(Level 2 — direct hook invocation; the host has no cpu.max/memory.max, which is exactly the case the CPU-only clamp missed)*
  ```bash
  $ uv run python -c "from tests.conftest import pytest_xdist_auto_num_workers as h, _cgroup_aware_cpu_count as c, _memory_aware_worker_count as m; print('cpu', c(), 'mem', m(), 'hook', h(None))"
  cpu 32 mem 23 hook 23
  ```
  Without this PR, `-n auto` would spawn 32 workers here; now it is memory-bounded to 23 (and lower as the host fills).

- [x] **Devcontainer pin and floor behave as specified** *(Level 1 + Level 2 — the `containerEnv` value is 4; the override floors at 1)*
  ```bash
  $ PYTEST_XDIST_AUTO_NUM_WORKERS=4 uv run pytest tests/test__conftest_cpu.py -n auto -q | grep -iE "workers \["
  4 workers [10 items]
  $ uv run python -c "import os; os.environ['PYTEST_XDIST_AUTO_NUM_WORKERS']='0'; from tests.conftest import pytest_xdist_auto_num_workers as h; print(h(None), 'worker')"
  1 worker
  ```
  The devcontainer default of `4` is what every shared-host worktree gets; a degenerate `0`/negative override can never produce `-n 0`.

- [ ] **No-memory-signal fallback to CPU count** *(Level 1 — SKIP: simulating an unreadable `/proc/meminfo` + cgroup inside a live xdist run requires patching `builtins.open` in the xdist controller process, which the unit suite already covers via `test_no_memory_signal_uses_cpu_count`.)*
  To exercise behaviorally on a host without `/proc/meminfo` (e.g. some non-Linux CI): run `pytest -n auto` and confirm the worker count equals the CPU-only clamp.

**Unit contract:** the 30 tests in `tests/test__conftest_mem.py` (20) + `tests/test__conftest_cpu.py` (10) pin the helper-level branches (host/cgroup `min`, v1/v2 sentinels, malformed input, env fallback). CI runs these; not duplicated as a standalone checkbox here.

PR #1530, opened 2026-06-06
test(testing): guard non-integer PYTEST_XDIST_AUTO_NUM_WORKERS override

### LINKED ISSUE #1528: bug(testing): non-integer PYTEST_XDIST_AUTO_NUM_WORKERS aborts pytest collection
## Symptom

A misconfigured non-integer (or empty) `PYTEST_XDIST_AUTO_NUM_WORKERS` raises `ValueError` in the `pytest_xdist_auto_num_workers` hook (`tests/conftest.py`), aborting pytest collection before any test runs — the same "suite dies before it starts" failure class the `-n auto` memory clamp (#1520) was meant to prevent.

## Root cause

The env-var escape hatch did `return max(1, int(env_override))` with no guard. A typo like `PYTEST_XDIST_AUTO_NUM_WORKERS=auto` crashes collection. Pre-existing since the CPU-aware clamp (#1492); surfaced while hardening #1520.

## Fix

Treat a non-integer/empty pin as "no override" and fall through to the adaptive CPU/memory clamps — mirroring the malformed-input handling already used for `PYTEST_XDIST_WORKER_MEM_MB`. Adds two regression tests (non-integer, empty-string).

ktinubu:
## Problem

A non-integer or empty `PYTEST_XDIST_AUTO_NUM_WORKERS` (a typo like `=auto`) raised `ValueError` in the `pytest_xdist_auto_num_workers` hook (`tests/conftest.py:1019` on main), aborting pytest collection before any test ran — the same "suite dies before it starts" failure class the `-n auto` memory clamp (#1524) was added to prevent. Pre-existing since the CPU-aware clamp (#1492); surfaced while reviewing #1524.

## Change

Guard the env-var escape hatch: treat a non-integer/empty pin as "no override" and fall through to the adaptive CPU/memory clamps, mirroring the malformed-input handling already used for `PYTEST_XDIST_WORKER_MEM_MB`.

```python
env_override = os.environ.get("PYTEST_XDIST_AUTO_NUM_WORKERS")
if env_override:
    try:
        return max(1, int(env_override))
    except ValueError:
        pass  # non-integer pin -> ignore and fall through to the adaptive clamps
```

The `if env_override:` (truthy) change also treats an empty string as "unset", which is the more correct reading of a blank env var.

## Tests

Two regression tests added to `tests/test__conftest_mem.py` (`"auto"` non-integer, `""` empty) — both fail against the pre-fix code (ValueError) and pass with the guard, asserting the hook falls through to the affinity-derived clamp count. 32 tests pass (`tests/test__conftest_mem.py` + `tests/test__conftest_cpu.py`).

Fixes #1528

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1017,15 +1017,19 @@ def pytest_xdist_auto_num_workers(config: pytest.Config) -> int:  # noqa: ARG001
 
     Checks ``PYTEST_XDIST_AUTO_NUM_WORKERS`` first so the env-var escape hatch
     that xdist's built-in implementation honours is preserved even when this
-    hook wins the ``firstresult`` race. Otherwise returns ``min(cpu, memory)``
-    so neither resource is over-subscribed.
+    hook wins the ``firstresult`` race. A non-integer or empty pin is ignored,
+    not fatal. Otherwise returns ``min(cpu, memory)`` so neither resource is
+    over-subscribed.
 
     :param config: The pytest config object (unused; required by the hook signature).
     :returns: Worker count clamped to the host's real CPU and memory allocation.
     """
     env_override = os.environ.get("PYTEST_XDIST_AUTO_NUM_WORKERS")
-    if env_override is not None:
-        return max(1, int(env_override))
+    if env_override:
+        try:
+            return max(1, int(env_override))
+        except ValueError:
+            pass  # non-integer pin -> ignore and fall through to the adaptive clamps
     cpu_workers = _cgroup_aware_cpu_count()
     mem_workers = _memory_aware_worker_count()
     return cpu_workers if mem_workers is None else min(cpu_workers, mem_workers)
--- a/tests/test__conftest_mem.py
+++ b/tests/test__conftest_mem.py
@@ -216,3 +216,29 @@ class TestHookCombinesCpuAndMemory:
         """
         monkeypatch.setenv("PYTEST_XDIST_AUTO_NUM_WORKERS", "0")
         assert pytest_xdist_auto_num_workers(cast("pytest.Config", None)) == 1
+
+    def test_env_override_non_integer_falls_back_to_clamps(
+        self, monkeypatch: pytest.MonkeyPatch
+    ) -> None:
+        """A non-integer env override is ignored, not fatal — clamps still apply.
+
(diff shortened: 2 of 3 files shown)


ktinubu:
## Verification Results

The promise: a non-integer/empty `PYTEST_XDIST_AUTO_NUM_WORKERS` no longer aborts pytest collection — the hook ignores it and falls through to the adaptive clamps.

- [x] **A non-integer pin no longer crashes `pytest -n auto`** *(Level 1 — runs the real pytest-xdist collection end-to-end; before/after on the same command, swapping only `tests/conftest.py` so the env is identical)*
  ```bash
  # BEFORE — origin/main's unguarded hook
  $ git checkout origin/main -- tests/conftest.py
  $ PYTEST_XDIST_AUTO_NUM_WORKERS=auto uv run pytest tests/test__conftest_cpu.py -n auto -q
  ValueError: invalid literal for int() with base 10: 'auto'

  # AFTER — this branch's guarded hook
  $ git checkout HEAD -- tests/conftest.py
  $ PYTEST_XDIST_AUTO_NUM_WORKERS=auto uv run pytest tests/test__conftest_cpu.py -n auto -q
  10 passed in 16.87s
  ```
  Pre-fix: collection aborts before any test runs. Post-fix: the bad pin is ignored and the suite runs on the adaptive clamp. This is the exact "suite dies before it starts" failure the guard removes.

- [x] **Regression tests pin both fall-through triggers** *(Level 2 — runs the new tests directly; they assert the resulting worker count, not just absence-of-crash)*
  ```bash
  $ uv run pytest "tests/test__conftest_mem.py::TestHookCombinesCpuAndMemory::test_env_override_non_integer_falls_back_to_clamps" "tests/test__conftest_mem.py::TestHookCombinesCpuAndMemory::test_env_override_empty_string_falls_back_to_clamps" -q
  2 passed
  ```
  `"auto"` (non-integer) and `""` (empty) both fall through to the affinity-derived clamp (asserted `== 3` and `== 4`). Each fails against the pre-fix code and passes with the guard.

**Unit contract:** full conftest suite is 32 tests (`tests/test__conftest_mem.py` + `tests/test__conftest_cpu.py`); CI runs them, not duplicated here.

PR #1535, opened 2026-06-06
automation(review): scan gh stdout for self-review 422 fallback

### LINKED ISSUE #1533: fix(review): self-review 422 fallback misses gh's stdout error body, aborting the review post
`post_review.py`'s `submit_review` retries a self-review 422 ("Can not request changes on your own pull request") as `event=COMMENT` with a banner — but it only scans `result.stderr` for the self-review signature. `gh api` writes the JSON error body (which carries that message) to **stdout**, and only `gh: Unprocessable Entity (HTTP 422)` to stderr. So on a real self-review the regex never matches, the fallback never fires, and the helper exits 1 — the whole `/repo-review-full` post fails instead of degrading to COMMENT.

Observed live on PR #1488 (bot is the PR author): the review had to be posted by manually replicating the COMMENT fallback.

Fix: scan both stdout and stderr for the self-review signature. Refs #1405 (the feature that introduced the REQUEST_CHANGES → COMMENT self-review fallback).

ktinubu:
## Problem

`post_review.py`'s `submit_review` retries a self-review 422 ("Can not request changes on your own pull request") as `event=COMMENT` with a banner — but it scanned only `result.stderr` for the self-review signature. `gh api` writes the JSON error body (which carries that message) to **stdout**, and only `gh: Unprocessable Entity (HTTP 422)` to stderr. So on a real self-review the regex never matched, the fallback never fired, and the helper exited 1 — the whole `/repo-review-full` post aborted instead of degrading to COMMENT.

Hit live on PR #1488 (bot is the PR author): the review had to be posted by manually replicating the COMMENT fallback.

## Fix

Scan both streams: `_SELF_REVIEW_422_RE.search(f"{result.stdout}\\n{result.stderr}")`. One-line change plus a regression test that puts the 422 body on stdout (gh's real placement) and asserts the COMMENT downgrade still fires.

Fixes #1533

### DIFF
--- a/agent/skills/_shared/post_review.py
+++ b/agent/skills/_shared/post_review.py
@@ -373,7 +373,8 @@ def submit_review(
     result = _post_review(repo, pr_number, payload)
     if result.returncode == 0:
         return json.loads(result.stdout)
-    if _SELF_REVIEW_422_RE.search(result.stderr):
+    # gh writes the JSON error body to stdout and only "HTTP 422" to stderr, so scan both.
+    if _SELF_REVIEW_422_RE.search(f"{result.stdout}\n{result.stderr}"):
         sys.stderr.write(
             "Self-review 422: falling back to event=COMMENT with the fallback banner.\n"
         )
--- a/tests/claude_hooks/test_post_review_event.py
+++ b/tests/claude_hooks/test_post_review_event.py
@@ -142,6 +142,44 @@ def test_submit_review_self_review_422_falls_back_to_comment(
     assert "Original review." in retried["body"]
 
 
+def test_submit_review_self_review_422_on_stdout_falls_back_to_comment(
+    helper: ModuleType, monkeypatch: pytest.MonkeyPatch
+) -> None:
+    """A self-review 422 whose body lands on stdout still retries as COMMENT.
+
+    :param helper: The loaded ``post_review`` module.
+    :param monkeypatch: Pytest fixture for patching ``subprocess.run``.
+    """
+    err_422 = SimpleNamespace(
+        returncode=1,
+        stdout=json.dumps(
+            {
+                "message": "Unprocessable Entity",
+                "errors": ["Review Can not request changes on your own pull request"],
+            }
+        ),
+        stderr="gh: Unprocessable Entity (HTTP 422)",
+    )
+    ok = SimpleNamespace(
+        returncode=0,
+        stdout=json.dumps({"html_url": "https://example/r/4"}),
+        stderr="",
+    )
+    fake_run, calls = _fake_run_factory([err_422, ok])
+    monkeypatch.setattr(helper.subprocess, "run", fake_run)
+    monkeypatch.setattr(helper, "gh_executable", lambda: "/usr/bin/gh")
+
+    payload = {"body": "Original review.", "event": "REQUEST_CHANGES", "comments": []}
+    response = helper.submit_review("o/r", 7, payload, fallback_banner="⛔ BANNER")
+
+    assert response["html_url"] == "https://example/r/4"
+    assert len(calls) == 2
+    retried = json.loads(calls[1])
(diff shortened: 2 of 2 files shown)


ktinubu (commented):
⛔ 0 BLOCKING finding(s) — changes required (self-review: posted as COMMENT)

Verification probe for #1533: exercising the self-review 422 → COMMENT fallback end-to-end. This REQUEST_CHANGES is expected to 422 (bot is the PR author) and downgrade to COMMENT via the stdout-aware fallback.

ktinubu:
## Verification Results

**Promise under test:** When a self-review 422 ("Can not request changes on your own pull request") arrives with the error body on **stdout** (gh's real placement), `submit_review` retries as `event=COMMENT` and posts successfully — instead of scanning stderr only, missing the signature, and aborting with exit 1.

- [x] **Live self-review fallback fires end-to-end against a real PR** *(Level 1 — promise is the live gh-API self-review path; the only faithful exercise is running the real helper against a self-authored PR. Lower levels can't reproduce gh's actual stdout/stderr split.)*
  ```bash
  $ python3 agent/skills/_shared/post_review.py < /tmp/payload.json   # event=REQUEST_CHANGES, pr 1535 (bot is author)
  Resolved 0 inline anchor(s) (0 fell back to nearest-in-hunk); 0 finding(s) on files outside the diff rolled into review body.
  Self-review 422: falling back to event=COMMENT with the fallback banner.
  https://github.com/tinaudio/synth-setter/pull/1535#pullrequestreview-4443514551
  $ echo $?
  0
  ```
  The REQUEST_CHANGES 422'd (self-review), the stdout-aware scan matched the signature, the fallback fired, and the COMMENT review posted (exit 0). On pre-fix code this path exited 1 — see the red/green check below. Payload used:
  ```json
  {"pr_number":1535,"repo":"tinaudio/synth-setter","event":"REQUEST_CHANGES","review_body":"Verification probe for #1533 ...","findings":[]}
  ```

- [x] **Regression test fails on pre-fix code, passes after (true red/green)** *(Level 2 — drives the real `submit_review` regex path; `subprocess.run` is faked at the process boundary, the correct seam since a real GitHub self-review 422 can't be triggered deterministically.)*
  ```bash
  # RED: revert the production line to `search(result.stderr)` only
  $ uv run pytest tests/claude_hooks/test_post_review_event.py::test_submit_review_self_review_422_on_stdout_falls_back_to_comment -q
  FAILED tests/claude_hooks/test_post_review_event.py::test_submit_review_self_review_422_on_stdout_falls_back_to_comment - SystemExit: 1
  1 failed in 0.49s

  # GREEN: fix restored (scan both streams)
  $ uv run pytest tests/claude_hooks/test_post_review_event.py -q
  10 passed in 0.33s
  ```
  Without the fix the helper takes the `sys.exit(result.returncode)` branch (the PR #1488 abort). With it, the COMMENT fallback fires.

- [x] **Broadened scan does not regress the original stderr-only path** *(Level 2 — the pre-existing `test_submit_review_self_review_422_falls_back_to_comment` and `_self_approve_422_` still pass, so concatenating stdout+stderr keeps the old stderr-carried case working.)*
  ```bash
  $ uv run pytest tests/claude_hooks/test_post_review_event.py -q
  10 passed in 0.33s
  ```

- [ ] **Non-self-review 422 still aborts (no false fallback)** *(Level 2 — SKIP: not run this session. The contract is that only a self-review signature triggers the COMMENT downgrade; an unrelated 422 must still exit non-zero. Covered by the existing `test_submit_review_non_self_422_exits` in the suite above, but I did not isolate-run it as a dedicated negative check.)*
  To verify manually:
  ```bash
  uv run pytest tests/claude_hooks/test_post_review_event.py::test_submit_review_non_self_422_exits -q
  ```

**Reviewed at:** f9688680be366fdf721a74b567bf9c92d3a86d15

PR #1545, opened 2026-06-06
feat(eval): add surge_xt render-order shuffle-probe sweep for #489

ktinubu:
## Summary

Adds `sweeps/generate_dataset_shuffle_probe_surge_xt.yaml` — the **render-order vs phase-init attribution** experiment for the #489 non-deterministic-rendering investigation. This is Experiment 1 of a batch designed around functionality that landed since the issue's last study (the shuffle render-order probe, `param_sample_cadence`, per-split inline oracle eval).

## What it does

The sweep drives `generate_dataset/smoke-shard-with-oracle-eval` on the **full `surge_xt` (300-param) set** with `render.param_sample_cadence=shard` pinned, gridding `plugin_reload_cadence {once, render}` × `gui_toggle_cadence {never, once, render, always_on}` (the `render`×`always_on` cell is schema-invalid and skipped).

`param_sample_cadence=shard` makes every sample in a shard render **one identical patch**, so the inline oracle eval sees uniform params and **auto-fires the shuffle render-order probe** (`shuffled_audio/*` beside `audio/*`). Because the patch is identical, the permutation isolates *render-order position* as the only variable.

## Question it answers

The prior #489 cadence studies concluded "phase-init dominates, not stale state" — but only as a **mechanistic inference**, never a measurement, because aggregate divergence can't separate render-order dependence from oscillator phase-init noise. This probe measures it directly:

- **`audio/*` ≈ `shuffled_audio/*`** → divergence is position-independent → **phase-init dominates**.
- **gap opens under `plugin_reload_cadence=once` only** → **render-order / stale plugin state** (the #489 root-cause hypothesis).

## Validation

Ran one grid cell end-to-end locally (surge_xt, `param_sample_cadence=shard`, `plugin_reload_cadence=once`, Xvfb, offline) at the smoke size. The probe fired on all three splits and produced `aggregated_metrics_shuffled.csv`. Preview result: test split `audio/mss_mean=2.32` vs `shuffled_audio/mss_mean=2.33` — at reuse depth 4, render-order is *not* the driver (consistent with the phase-init hypothesis).

A parallel offline structural test (`tests/test_sweep_yaml_structure.py::test_generate_dataset_shuffle_probe_yaml_shape`) pins the sweep's command shape and the probe-trigger overrides (`param_sample_cadence=shard`, `param_spec_name=surge_xt`, `preset_path`), and guards that `param_sample_cadence` stays a fixed override rather than a swept grid dimension.

## Testing

- Local end-to-end smoke run of one grid cell — green, shuffle probe fired.

Refs #489

### DIFF
--- a/sweeps/generate_dataset_shuffle_probe_surge_xt.yaml
+++ b/sweeps/generate_dataset_shuffle_probe_surge_xt.yaml
@@ -1,0 +1,35 @@
+program: src/synth_setter/cli/generate_dataset.py
+# Pinned in YAML — wandb sweep CLI does not support OmegaConf resolvers and ignores
+# WANDB_ENTITY / WANDB_PROJECT at sweep-creation time (those only steer wandb.init in each trial).
+entity: tinaudio
+project: synth-setter
+# Render-order vs phase-init probe for #489: param_sample_cadence=shard gives every sample one
+# identical patch, so the inline oracle eval auto-fires the shuffle probe (shuffled_audio/* vs audio/*).
+command:
+  - ${interpreter}
+  - ${program}
+  - experiment=generate_dataset/smoke-shard-with-oracle-eval
+  - render.param_spec_name=surge_xt
+  - render.preset_path=presets/surge-base.vstpreset
+  - render.param_sample_cadence=shard
+  - ${args_no_hyphens}
+name: generate_dataset_shuffle_probe_surge_xt
+method: grid
+# grid ignores `metric` at scheduling time; the field stays for dashboard
+# legibility on runs that do log it. mss_mean is the oracle-eval audio loss.
+metric:
+  goal: minimize
+  name: audio/mss_mean
+# 2x4 grid: `once` reuses one plugin instance across the shard (the #489 stale-state hypothesis),
+# `render` reloads per call (phase-init floor). The render x always_on cell is skipped as schema-invalid.
+parameters:
+  render.plugin_reload_cadence:
+    values:
+      - once
+      - render
+  render.gui_toggle_cadence:
+    values:
+      - never
+      - once
+      - render
+      - "always_on"
--- a/tests/test_sweep_yaml_structure.py
+++ b/tests/test_sweep_yaml_structure.py
@@ -1,8 +1,8 @@
-"""Offline structural assertion for ``sweeps/generate_dataset_cadence.yaml``.
+"""Offline structural assertions for the operator-facing ``sweeps/*.yaml``.
 
-Pins the operator-facing sweep YAML's ``command:`` shape and parameter
-grid so the contract that ``wandb agent`` executes (subprocess launch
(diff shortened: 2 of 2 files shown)


ktinubu:
## Verification Results

The sweep's promise: running its `command:` on the full `surge_xt` set with `param_sample_cadence=shard` makes the inline oracle eval see uniform params and **auto-fire the shuffle render-order probe**, emitting `shuffled_audio/*` beside `audio/*` for every split.

- [x] **The sweep command fires the shuffle render-order probe end-to-end and writes `shuffled_audio` metrics for all three splits** *(Level 1 — full generate→finalize→inline-oracle-eval integration on the real Surge XT VST; the promise is end-to-end)*
  ```bash
      experiment=generate_dataset/smoke-shard-with-oracle-eval \
      render.param_spec_name=surge_xt render.preset_path=presets/surge-base.vstpreset \
      render.param_sample_cadence=shard \
      render.plugin_reload_cadence=once render.gui_toggle_cadence=never
  # exit 0

  $ for s in train val test; do
      find logs -path "*oracle_eval/$s/*aggregated_metrics_shuffled.csv";
    done
  logs/.../oracle_eval/train/.../metrics/aggregated_metrics_shuffled.csv
  logs/.../oracle_eval/val/.../metrics/aggregated_metrics_shuffled.csv
  logs/.../oracle_eval/test/.../metrics/aggregated_metrics_shuffled.csv
  ```
  Test split, `audio/*` vs `shuffled_audio/*` (the probe's discriminator):
  | metric | audio (each target vs its own pred) | shuffled (target vs another sample's pred) |
  |---|---|---|
  | mss | 0.294 | 0.255 |
  | wmfcc | 1.680 | 1.270 |
  | sot | 0.0460 | 0.0458 |
  | rms | 0.961 | 0.977 |

  At reuse depth 4 the shuffled and unshuffled metrics are within noise (shuffled is, if anything, marginally lower), i.e. **render-order is not the divergence driver here** — consistent with the phase-init hypothesis the prior #489 studies could only infer. The depth sweep that pushes this toward the catastrophe regime is the sibling reuse-depth PR.

- [x] **The sweep's launch contract is pinned against drift** *(Level 2 — runs pytest on the specific structural test; the promise is that the `wandb agent` command shape + probe-trigger overrides can't silently change)*
  ```bash
  $ uv run pytest tests/test_sweep_yaml_structure.py::test_generate_dataset_shuffle_probe_yaml_shape -q
  1 passed
  ```
  Pins `program`, the `${interpreter} ${program} … ${args_no_hyphens}` template, the load-bearing overrides (`param_sample_cadence=shard`, `param_spec_name=surge_xt`, `preset_path`), and that `param_sample_cadence` stays a fixed override rather than a swept grid dimension (gridding it would un-gate the probe).

_Note: a live multi-trial `wandb agent` sweep can't run in CI; the Level-1 check exercises one grid cell of the sweep's exact command, which is the unit that could regress._

ktinubu:

PR #1546, opened 2026-06-06
feat(eval): add surge_xt reuse-depth junk-probe sweep for #489

ktinubu:
## Summary

Adds `sweeps/generate_dataset_reuse_depth_surge_xt.yaml` — the **reuse-depth junk-reproduction** experiment for the #489 non-deterministic-rendering investigation (Experiment 3 of the batch). It aims to reproduce the every-other-render junk (`mss≈21`, PR #706) inside the instrumented oracle harness, scaled across reuse depth — something the prior #489 studies never managed (their Tier 3 only reached `samples_per_shard=40`).

## What it does

Grids `render.samples_per_shard ∈ {4, 20, 40, 80}` on the full `surge_xt` set with `render.param_sample_cadence=shard` + `render.plugin_reload_cadence=once` pinned. That combination reuses **one cached plugin instance** across **one identical patch** for the whole shard, so `samples_per_shard` *is* the reuse depth. Only sample 0 is loudness-gated, so the reused samples expose any stale-state junk the gate would otherwise resample away. `train_val_test_sizes` is pinned to `[80,80,80]` so every split is a whole number of shards at all four depths.

## Question it answers

The #489 status update explicitly asked for a **strict regression guard** that fails if `plugin_reload_cadence="once"` reintroduces the every-other-render variance. This sweep is the controlled instrument for that:

- If `mss` blows up only at deep reuse → the junk is reuse-depth-dependent stale state → the per-render-reload workaround is load-bearing and surgepy migration is justified.
- If even depth 80 stays clean → strong evidence the catastrophe was specific to the raw-VST path the reload workaround already eliminated → #489 can be de-escalated.

Because shard cadence makes params uniform, the **shuffle render-order probe auto-fires** at every depth (`shuffled_audio/*` vs `audio/*`), attributing any junk to render-order vs phase-init — complementing the sibling shuffle-probe PR #1545.

## Validation

Ran one grid cell (depth 20, one shard/split of 20 reuses) end-to-end locally on the real Surge XT VST (Xvfb, offline) at a reduced split size. Generation + inline oracle eval completed and the shuffle probe fired. A parallel offline structural test (`tests/test_sweep_yaml_structure.py::test_generate_dataset_reuse_depth_yaml_shape`) pins the command shape, the load-bearing reuse overrides (`param_sample_cadence=shard`, `plugin_reload_cadence=once`, `gui_toggle_cadence=never`, `param_spec_name=surge_xt`), and that every pinned split size is a multiple of every swept depth (mirroring the real `DatasetSpec` validator).

## Testing

- Local end-to-end smoke run of one depth-20 grid cell — green, shuffle probe fired.

Refs #489

### DIFF
--- a/docs/doc-map.yaml
+++ b/docs/doc-map.yaml
@@ -62,7 +62,7 @@ docs:
       - pattern: "src/synth_setter/configs/dataset.yaml"
         covers: "Default `logger: wandb` for dataset generation; entity/project sourced from logger/wandb.yaml"
       - pattern: "sweeps/**"
-        covers: "Operator-authored W&B sweep configs driving `synth-setter-generate-dataset`; entity/project pinned in YAML, cadence-grid parameters threaded as Hydra overrides"
+        covers: "Operator-authored W&B sweep configs driving `synth-setter-generate-dataset`; entity/project pinned in YAML, grid parameters (render cadences, reuse depth) threaded as Hydra overrides"
       - pattern: "src/synth_setter/utils/logging_utils.py"
         covers: "Metric logging helpers, hyperparameter logging, log_hyperparameters"
       - pattern: "src/synth_setter/utils/callbacks.py"
--- a/sweeps/generate_dataset_reuse_depth_surge_xt.yaml
+++ b/sweeps/generate_dataset_reuse_depth_surge_xt.yaml
@@ -1,0 +1,35 @@
+program: src/synth_setter/cli/generate_dataset.py
+# Pinned in YAML — wandb sweep CLI does not support OmegaConf resolvers and ignores
+# WANDB_ENTITY / WANDB_PROJECT at sweep-creation time (those only steer wandb.init in each trial).
+entity: tinaudio
+project: synth-setter
+# Reuse-depth junk probe for #489: param_sample_cadence=shard + plugin_reload_cadence=once reuse one
+# cached plugin instance across one identical patch, exposing the every-other-render junk (mss~21, #706).
+command:
+  - ${interpreter}
+  - ${program}
+  - experiment=generate_dataset/smoke-shard-with-oracle-eval
+  - render.param_spec_name=surge_xt
+  - render.preset_path=presets/surge-base.vstpreset
+  - render.param_sample_cadence=shard
+  - render.plugin_reload_cadence=once
+  - render.gui_toggle_cadence=never
+  # 80 is a multiple of every swept reuse depth, so each split is a whole number of shards.
+  - train_val_test_sizes=[80,80,80]
+  - ${args_no_hyphens}
+name: generate_dataset_reuse_depth_surge_xt
+method: grid
+# grid ignores `metric` at scheduling time; the field stays for dashboard legibility on runs that do
+# log it. The science is the junk tail (mss_max / %mss>3); mss_mean is just the W&B handle.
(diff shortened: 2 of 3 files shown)


ktinubu:
## Verification Results

The sweep's promise: each grid cell runs `generate_dataset` with `param_sample_cadence=shard` + `plugin_reload_cadence=once` so `samples_per_shard` renders reuse one cached plugin instance across one identical patch, and the inline oracle eval measures each reuse against a clean re-render (with the shuffle probe auto-firing on the uniform params).

- [x] **A real reuse-depth grid cell runs end-to-end and the oracle eval + shuffle probe measure every reuse** *(Level 1 — full generate→finalize→inline-oracle-eval integration on the real Surge XT VST; the promise is end-to-end reuse-depth measurement)*
  ```bash
      experiment=generate_dataset/smoke-shard-with-oracle-eval \
      render.param_spec_name=surge_xt render.preset_path=presets/surge-base.vstpreset \
      render.param_sample_cadence=shard render.plugin_reload_cadence=once \
      render.gui_toggle_cadence=never \
      train_val_test_sizes=[20,20,20] render.samples_per_shard=20
  # exit 0  — one shard per split = 20 reuses of one patch

  $ python3 -c "import csv,glob; m=sorted(float(r['mss']) for f in glob.glob('logs/*/runs/*/oracle_eval/*/*/metrics/metrics.csv') for r in csv.DictReader(open(f))); print(f'n={len(m)} median={m[len(m)//2]:.3f} max={m[-1]:.3f} junk(mss>3)={sum(x>3 for x in m)}')"
  n=60 median=1.069 max=1.692 junk(mss>3)=0
  ```
  Depth-20 result, `audio/*` vs `shuffled_audio/*` (test split):
  | metric | audio | shuffled |
  |---|---|---|
  | mss | 1.568 | 1.562 |
  | wmfcc | 1.297 | 1.308 |

  At 20 reuses of a single patch the per-sample mss tail tops out at 1.69 with **zero** `mss>3` samples, and `shuffled ≈ audio` on every split — i.e. the every-other-render catastrophe does **not** reproduce at depth 20, and what divergence exists is position-independent (phase-init), not render-order. The `{40, 80}` cells of the committed grid are what push toward #706's regime; this validates the harness produces the interpretable signal the experiment needs.

- [x] **The sweep's reuse-path contract is pinned against drift** *(Level 2 — runs pytest on the specific structural test; the promise is that the reuse overrides + split divisibility can't silently change)*
  ```bash
  $ uv run pytest tests/test_sweep_yaml_structure.py::test_generate_dataset_reuse_depth_yaml_shape -q
  1 passed
  ```
  Pins the `wandb agent` command shape, the four load-bearing overrides (`param_sample_cadence=shard`, `plugin_reload_cadence=once`, `gui_toggle_cadence=never`, `param_spec_name=surge_xt`), that `samples_per_shard` is the sole swept axis, and — parsed from the pinned `train_val_test_sizes` rather than a literal — that every split size is a multiple of every swept depth (mirroring the real `DatasetSpec` validator, so an invalid grid cell can't slip in).

_Note: a live multi-trial `wandb agent` sweep can't run in CI; the Level-1 check exercises one grid cell (the deepest validated depth, 20) of the sweep's exact command — the unit that could regress._

ktinubu:

Issue #884, opened 2026-05-10
Deterministic per-sample seeding for reproducible datagen with sharded execution + retries


ktinubu:
We want reproducible dataset generation: rendering with a random parameter sampler, retrying when audio is silent, and producing a dataset where sample N always has the same content regardless of worker/sharding/order.

The current schema has `train_val_test_seeds` and `ShardSpec.seed` as fields, but neither is plumbed into the writer. Workers don't seed any RNG. Spec advertises reproducibility it can't deliver.

### Proposed approach (design transcript)

> The core idea: deterministic seed derivation. Instead of one global RNG that you advance, give every sample a deterministic seed derived from a master seed plus its index. Something like:
>
> ```python
> import hashlib, numpy as np
>
> def seed_for_sample(master_seed: int, sample_idx: int, attempt: int = 0) -> int:
>     h = hashlib.sha256(f"{master_seed}:{sample_idx}:{attempt}".encode()).digest()
>     return int.from_bytes(h[:8], "big")
>
> def rng_for_sample(master_seed, sample_idx, attempt=0):
>     return np.random.default_rng(seed_for_sample(master_seed, sample_idx, attempt))
> ```
>
> Now sample 12345 always gets the same seed regardless of which worker generates it, what order workers process samples in, whether earlier samples crashed/succeeded/were retried, or how the work is sharded. A single global RNG can't give this because its state depends on consumption order.
>
> **Handling silent samples** — the retry attempt is part of the seed input, not a global counter:
>
> ```python
> def generate_sample(master_seed, sample_idx, max_attempts=20):
>     for attempt in range(max_attempts):
>         rng = rng_for_sample(master_seed, sample_idx, attempt)
>         params = sample_params(rng)
>         audio = render(params)
>         if not is_silent(audio):
>             return {"idx": sample_idx, "attempt": attempt, "params": params, "audio": audio}
>     raise RuntimeError(f"sample {sample_idx} silent after {max_attempts} attempts")
> ```
>
> Sample 12345 always tries `(master, 12345, 0)` first; if silent, always tries `(master, 12345, 1)` next. The accepted attempt number is deterministic given the silence-detection function. Save `attempt` in metadata so reproducibility is verifiable.
>
> **Worker crashes & sharded execution** — because seeds are pure functions of `(master_seed, sample_idx)`, workers don't need shared RNG state. Sharding becomes:
>
> ```python
> for sample_idx in range(worker_id, N, num_workers):
[...]

PR #1553, opened 2026-06-07
feat(eval): add paired-copy reload-cadence sweep for #489

ktinubu:
## Summary

Adds `sweeps/generate_dataset_copy_paired_reload_surge_xt.yaml` — the **paired plugin_reload_cadence** experiment (Exp 2) for the #489 investigation, now possible via the merged `copy_dataset_root_uri` feature (#1548).

## What it does

Grids `render.plugin_reload_cadence ∈ {once, render}` on the copy base (`experiment=generate_dataset/copy-paired-surge-xt`), replaying the **fixed 120-patch surge_xt reference** (`copy_dataset_root_uri=r2://intermediate-data/data/ref-surge-xt-489/paired-ref-v1`) in **both arms**. Because both arms render the *same* patches, the once-vs-render contrast is **paired per patch**.

## Question it answers

The prior #489 cadence studies found no `once`-vs-`render` effect — but **unpaired** (the unseeded param RNG, #884, drew different patches each run), so cross-run patch variance swamped the signal. This sweep removes that confound: a **Wilcoxon signed-rank on the per-patch metric deltas** is ~an order of magnitude more sensitive. Still null → strong confound-free confirmation that cadence doesn't matter; a real effect emerges → the prior null was underpowered.

## Dependency

Needs **PR #1551** (the `copy-paired-surge-xt` base config + the reference dataset). **Merge #1551 first.** CI here only structurally validates the sweep YAML, so it's green independently.

## Validation

The copy mechanism was validated end-to-end against the live reference (from the #1551 worktree): `_validate_copy_source` logged *"dataset-copy source OK: r2://…/ref-surge-xt-489/paired-ref-v1 matches the target spec"*, 120 patches re-rendered under `plugin_reload_cadence=once`, finalized to `dataset.complete`, and the inline oracle eval produced `train/audio/mss_mean=2.00`. The structural test pins the copy base, the exact reference URI, `method=grid`, and the `{once,render}` grid.

## Testing

- Copy mechanism validated end-to-end (preflight + re-render + finalize + oracle eval).

Refs #489

### DIFF
--- a/sweeps/generate_dataset_copy_paired_reload_surge_xt.yaml
+++ b/sweeps/generate_dataset_copy_paired_reload_surge_xt.yaml
@@ -1,0 +1,27 @@
+program: src/synth_setter/cli/generate_dataset.py
+# Pinned in YAML — wandb sweep CLI does not support OmegaConf resolvers and ignores
+# WANDB_ENTITY / WANDB_PROJECT at sweep-creation time (those only steer wandb.init in each trial).
+entity: tinaudio
+project: synth-setter
+# Paired plugin_reload_cadence probe for #489: copy_dataset_root_uri replays the fixed surge_xt
+# reference patches, so once-vs-render is paired per patch (needs the copy base + reference, PR #1551).
+command:
+  - ${interpreter}
+  - ${program}
+  - experiment=generate_dataset/copy-paired-surge-xt
+  - copy_dataset_root_uri=r2://intermediate-data/data/ref-surge-xt-489/paired-ref-v1
+  - ${args_no_hyphens}
+name: generate_dataset_copy_paired_reload_surge_xt
+method: grid
+# grid ignores `metric` at scheduling time; the field stays for dashboard
+# legibility on runs that do log it. mss_mean is the oracle-eval audio loss.
+metric:
+  goal: minimize
+  name: audio/mss_mean
+# Both arms replay the same patches, so a paired test (Wilcoxon on per-patch deltas) answers whether
+# the prior unpaired cadence null was real or just underpowered by the patch confound.
+parameters:
+  render.plugin_reload_cadence:
+    values:
+      - once
+      - render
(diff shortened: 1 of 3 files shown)


ktinubu:
## Verification Results

The sweep's promise: each arm runs the copy base against the fixed reference, replaying the **same** patches under a different `plugin_reload_cadence`, so the oracle measures a per-patch-paired round-trip. The `once` arm was run end-to-end against the live reference.

- [x] **The copy base + reference URI preflight-pass, re-render the reference patches, and produce paired oracle metrics** *(Level 1 — full copy-preflight → re-render → finalize → inline-oracle-eval on the real Surge XT VST; the promise is an end-to-end paired round-trip)*
  ```bash
      experiment=generate_dataset/copy-paired-surge-xt \
      copy_dataset_root_uri=r2://intermediate-data/data/ref-surge-xt-489/paired-ref-v1 \
      render.plugin_reload_cadence=once \
      task_name=copy-validate run_id=validate-once-v1
  # exit 0
  # log: _validate_copy_source: "dataset-copy source OK:
  #   r2://intermediate-data/data/ref-surge-xt-489/paired-ref-v1 matches the target spec"

  $ rclone lsf r2:intermediate-data/data/copy-validate/validate-once-v1/
  dataset.complete
  input_spec.json
  shard-000000.h5 … shard-000005.h5
  stats.npz
  test.h5
  ```
  The copy preflight (`validate_copy_source`) accepted the reference, all 120 patches were re-rendered from the reference's saved `param_array` under `plugin_reload_cadence=once`, the dataset finalized, and the inline oracle eval produced `train/audio/mss_mean=2.00` (wmfcc 2.73, rms 0.977) — a per-patch-paired round-trip metric. This is the exact `once` arm of the sweep; the `render` arm differs only in the gridded knob.

- [x] **The sweep's paired contract is pinned against drift** *(Level 2 — runs pytest on the structural test; the promise is the copy base + reference URI + grid can't silently change)*
  ```bash
  $ uv run pytest tests/test_sweep_yaml_structure.py::test_generate_dataset_copy_paired_reload_yaml_shape -q
  1 passed
  ```
  Pins the `wandb agent` command shape, `experiment=copy-paired-surge-xt`, the exact `copy_dataset_root_uri` reference string (the patch set every arm replays), `method=grid`, and the `{once, render}` grid values.

_The reference dataset and copy base land in #1551 (merge first). A live multi-trial `wandb agent` sweep can't run in CI; the Level-1 check is the sweep's exact `once`-arm command against the live reference._

ktinubu:

PR #1584, opened 2026-06-10
fix(testing): give citest checkpoint artifacts run-scoped names

ktinubu:
## Why

`test_train_resumes_from_wandb_resolved_checkpoint[mps-surge/ffn_full]` (and the eval equivalent) fail on the MPS CI job with:

```
ValueError: ArtifactSaver.upsertManifest: createManifest returned error 400:
cannot create manifest for artifact in state PENDING_DELETION
```

The failure is **unrelated to the PR's own diff** — it's a wedged W&B server-state. `tests/helpers/wandb_artifacts.py` published every run's smoke checkpoint to a **fixed** collection name (`model-citest-ffn_full-resume` / `-eval`) and read it back via `:latest`. Once that single collection enters W&B's `PENDING_DELETION` state (versions deleted, async GC not finalized), `log_artifact` → `createManifest` is rejected with HTTP 400 and stays wedged until GC completes. A fixed name also races on `:latest` across concurrent runs and risks stale-cache reuse on warm self-hosted runners.

Confirmed live via the W&B API: the collection exists but `…-resume:latest` resolves to *no membership* (zero live versions) while publish still 400s — the half-deleted-collection signature.

## What changed

- `publish_checkpoint_artifact` → **`published_checkpoint_artifact`**, now a `@contextlib.contextmanager`.
- **Run-scoped names**: each run publishes under `<base>-<run.id>` (`_run_scoped_name`), so a half-deleted collection can never block a future run, and the `:latest` race vanishes. The resolver's cache key already hashes the full ref, so unique names cache cleanly.
- **Post-resolve cleanup**: on context exit the published version is deleted best-effort (`delete_aliases=True`, swallow+log on error) so checkpoint bytes don't accumulate in the scratch project.
- Both `test_train` and `test_eval` call sites wrapped in the `with` block.

## Test plan

- `tests/helpers/test_wandb_artifacts.py` (new): 4 fast, network-free tests — run-scoped naming, cleanup-on-exception (the core invariant), best-effort swallow, and `delete_aliases=True`. All green.
- The live publish→resolve→cleanup round-trip stays covered by the `WANDB_API_KEY`-gated slow tests in `test_train` / `test_eval`.
- `make format` (pre-commit) clean.
- `/repo-review-full-no-comments`: 5 parallel passes (code-health, synth-setter-project-standards, python-style, tdd-implementation, comment-hygiene), **0 BLOCK**; the two actionable WARNs were applied.

Refs #155

### DIFF
--- a/tests/helpers/test_wandb_artifacts.py
+++ b/tests/helpers/test_wandb_artifacts.py
@@ -1,0 +1,67 @@
+"""Unit tests for the citest checkpoint-artifact helper's offline-testable seams.
+
+The live round-trip (publish → resolve → cleanup) is exercised by the slow,
+``WANDB_API_KEY``-gated tests in ``test_train`` / ``test_eval``. These cover the
+two pieces that need no network: run-scoped naming and best-effort deletion.
+"""
+
+from __future__ import annotations
+
+import pytest
+
+from tests.helpers import wandb_artifacts
+
+
+def test_run_scoped_name_distinct_runs_distinct_names() -> None:
+    """Distinct run ids yield distinct, base-prefixed collection names."""
+    base = "model-citest-ffn_full-resume"
+    name_a = wandb_artifacts._run_scoped_name(base, "abc123")
+    name_b = wandb_artifacts._run_scoped_name(base, "def456")
+
+    assert name_a != name_b, "same name across runs would collide on the shared collection"
+    assert name_a.startswith(f"{base}-") and name_b.startswith(f"{base}-")
+    assert "abc123" in name_a and "def456" in name_b
+
+
+def test_delete_artifact_best_effort_api_error_does_not_raise(
+    monkeypatch: pytest.MonkeyPatch,
+) -> None:
+    """A failing delete is swallowed — scratch cleanup must never fail the test.
+
+    :param monkeypatch: Stubs ``wandb.Api`` to raise on construction.
+    """
+    import wandb
+
+    def _raise() -> None:
+        raise RuntimeError("network down")
+
+    monkeypatch.setattr(wandb, "Api", _raise)
+    # Must return normally despite the API blowing up.
+    wandb_artifacts._delete_artifact_best_effort("entity/proj/name:latest")
+
+
+def test_delete_artifact_best_effort_passes_delete_aliases(
+    monkeypatch: pytest.MonkeyPatch,
+) -> None:
+    """The version carries a ``:latest`` alias, so deletion must pass ``delete_aliases=True``.
+
+    :param monkeypatch: Stubs ``wandb.Api`` with a fake recording the delete call.
+    """
+    import wandb
+
+    calls: dict[str, object] = {}
+
+    class _FakeArtifact:
+        def delete(self, delete_aliases: bool = False) -> None:
+            calls["delete_aliases"] = delete_aliases
+
+    class _FakeApi:
+        def artifact(self, ref: str) -> _FakeArtifact:
+            calls["ref"] = ref
+            return _FakeArtifact()
+
+    monkeypatch.setattr(wandb, "Api", _FakeApi)
+    wandb_artifacts._delete_artifact_best_effort("entity/proj/name:latest")
+
+    assert calls["ref"] == "entity/proj/name:latest"
+    assert calls["delete_aliases"] is True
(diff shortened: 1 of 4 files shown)


ktinubu:
## Verification Results

Branch `test/citest-artifact-unique-cleanup` @ `20f81b8c`. The change's promises: (a) checkpoints publish under a **run-scoped** name, never a fixed one; (b) the published version is **deleted on context exit even when the `with`-body raises**; (c) a failing delete is **swallowed**; (d) the context-manager migration + rename doesn't break the two callers or the resolver.

- [x] **Cleanup deletes the run-scoped version even when the body raises** *(Level 2 — invokes the real `published_checkpoint_artifact` context manager with the W&B publish/delete surface faked; the promise is the CM's lifecycle contract, not an end-to-end W&B round-trip)*
  ```bash
  $ uv run pytest tests/helpers/test_wandb_artifacts.py::test_published_checkpoint_artifact_deletes_version_when_body_raises -q
  1 passed
  ```
  The test asserts the yielded ref is `ent/synth-setter-citest/model-citest-x-run42:latest` (run id embedded → run-scoped, promise a) and that `Api().artifact(ref).delete(delete_aliases=True)` is called once after the body raises (promise b).

- [x] **A failing delete is swallowed, not raised** *(Level 2 — drives the real `_delete_artifact_best_effort` with `wandb.Api` stubbed to raise)*
  ```bash
  $ uv run pytest tests/helpers/test_wandb_artifacts.py -v 2>&1 | grep -E "PASSED|WARNING|passed"
  tests/helpers/test_wandb_artifacts.py::test_delete_artifact_best_effort_passes_delete_aliases PASSED [ 25%]
  WARNING  tests.helpers.wandb_artifacts:wandb_artifacts.py:101 failed to delete citest artifact entity/proj/name:latest
  PASSED                                                                   [ 50%]
  tests/helpers/test_wandb_artifacts.py::test_published_checkpoint_artifact_deletes_version_when_body_raises PASSED [ 75%]
  tests/helpers/test_wandb_artifacts.py::test_run_scoped_name_distinct_runs_distinct_names PASSED [100%]
  4 passed in 1.09s
  ```
  The `failed to delete citest artifact …` WARNING is the swallow path firing (logged, not raised) — the test returns normally, proving promise c.

- [x] **Context-manager migration + import rename keeps both callers loadable** *(Level 2 — pytest collection imports & parametrizes the two `with published_checkpoint_artifact(...)` call sites; a broken rename/import or malformed `with` block would error at collection)*
  ```bash
  $ uv run pytest tests/test_train.py tests/test_eval.py --collect-only -q 2>&1 | tail -1
  41 tests collected in 1.56s
  ```

- [x] **Resolver still handles the ref shape unchanged** *(Level 2 — adversarial: run-scoping changes only the `name` component; the resolver hashes the full ref for its cache key, so the `entity/project/name:latest` contract must still resolve. Runs the resolver suite.)*
  ```bash
  $ uv run pytest tests/test_wandb_resolver.py -q 2>&1 | tail -1
  31 passed in 1.85s
  ```

- [ ] **Live publish→resolve→resume round-trip on the real W&B API** *(Level 1 — SKIP: the original failing test is `test_train_resumes_from_wandb_resolved_checkpoint[mps-surge/ffn_full]`, which needs an Apple-Silicon MPS runner and a `WANDB_API_KEY` that writes to the shared `synth-setter-citest` project. `torch.backends.mps.is_available()` is `False` on this Linux box, and writing to the shared CI W&B project is disallowed from this session.)*
  This is exactly what the **MPS Tests** CI job runs on the PR. To verify manually on a macOS Apple-Silicon machine with the CI key:
  ```bash
  # Plugin + key set up per docs; then:
  WANDB_API_KEY=<key> uv run pytest -m mps -k "resumes_from_wandb_resolved_checkpoint" -vv
  # Expect: publishes model-citest-ffn_full-resume-<run_id>, resolves it, resumes,
  #         asserts step_after_resume > step_after_first, then deletes the version on exit.
  ```
  The CPU variant `…[cpu-surge/ffn_full]` (and the eval equivalent) exercise the same publish→resolve path off-device with the same key.

**Net:** the behavioral invariants this PR adds (run-scoped naming, cleanup-on-exception, best-effort swallow) are verified locally at Level 2. The end-to-end W&B round-trip that originally wedged is left to the MPS CI job — the only environment that can run it.

Issue #1621, opened 2026-06-10
test(testing): parametrize duplicate rank/world/shard partition tests


ktinubu:
Sub-issue of #1610 (Phase 4), Epic #1603.

`tests/pipeline/entrypoints/test_generate_dataset_unit.py:703-791` has three near-identical tests (`test_rank_0_of_2_...`, `test_rank_1_of_2_...`, `test_excess_worker_...`) that differ only in rank/world/shard counts and expected indices. Collapse to one `@pytest.mark.parametrize`. Audit the rest of the file for sibling clusters and parametrize those too. Behavior-preserving: same assertions, same coverage.

ktinubu:
Closing as complete. Verified by multi-agent triage+audit+close-verify pass (tracking #1734).

## What was asked
Collapse three near-identical rank/world/shard partition tests in `test_generate_dataset_unit.py:703-791` into one `@pytest.mark.parametrize`, and audit the rest of the file for similar clusters.

## What landed
PR #1624 (commit `dff6aeff`) — parametrized duplicate rank/world/shard partition tests.

- `@pytest.mark.parametrize` found at test file lines 705, 789, confirming the three-case parametrize was applied
- The file went from three near-identical tests to one parametrized test covering rank_0, rank_1, and excess-worker cases
- No additional clusters requiring collapse were identified

## Verification
- [x] Commit on `main`: `dff6aeff` reachable (`git merge-base --is-ancestor` exit 0)
- [x] Cited deliverable present: `tests/pipeline/entrypoints/test_generate_dataset_unit.py:705` — `@pytest.mark.parametrize`
- [x] No subsequent revert: `git log --all -- tests/pipeline/entrypoints/test_generate_dataset_unit.py` clean

Refs #1734

PR #1624, opened 2026-06-10
test(testing): parametrize duplicate rank/world/shard partition tests

ktinubu:
## What

Collapses near-identical tests in `tests/pipeline/entrypoints/test_generate_dataset_unit.py` into `@pytest.mark.parametrize`d tests, preserving every assertion and the full case matrix.

- `test_rank_0_of_2_renders_only_first_half_of_shards`, `test_rank_1_of_2_renders_only_remaining_shard`, and `test_excess_worker_renders_no_shards` → one `test_worker_renders_only_its_partition_of_shards` parametrized over `(rank, world, expected_indices)` with descriptive ids. The empty-slice (excess worker) case keeps its stronger `assert_not_called()` + prefix-not-created guarantees via a guard clause.
- `test_run_renders_when_object_absent` and `test_run_renders_when_object_zero_size` → one `test_run_renders_when_object_treated_as_absent` parametrized over the `object_size` probe return (`None` / `0`). The `None` case reproduces the autouse fixture default exactly.

## Why

Both clusters differed only in trivial rank/world/shard or probe-value variation with identical assertions — the textbook parametrize case. Other near-duplicate-looking clusters in the file (oracle-eval missing-artifact pair, finalize-inline pair, copy-dataset-root pair) were left alone because they differ in setup, assertions, or intent.

## Verification

- `uv run pytest tests/pipeline/entrypoints/test_generate_dataset_unit.py -q` → 86 passed (same count as before).
- `make test-fast` → green.
- `make format` → all hooks pass.

Refs #1621

### DIFF
--- a/tests/pipeline/entrypoints/test_generate_dataset_unit.py
+++ b/tests/pipeline/entrypoints/test_generate_dataset_unit.py
@@ -700,24 +700,42 @@ class TestRun:
         patched_subprocess.assert_not_called()
         assert not (fake_r2_remote / spec.r2.bucket / spec.r2.prefix).exists()
 
-    def test_rank_0_of_2_renders_only_first_half_of_shards(
+    @pytest.mark.parametrize(
+        ("rank", "world", "expected_indices"),
+        [
+            pytest.param(0, 2, [0, 1], id="rank0-of-2-renders-first-half"),
+            pytest.param(1, 2, [2], id="rank1-of-2-renders-remainder"),
+            pytest.param(3, 4, [], id="excess-worker-renders-none"),
+        ],
+    )
+    def test_worker_renders_only_its_partition_of_shards(
         self,
         patched_subprocess: MagicMock,
         fake_r2_remote: Path,
         tmp_path: Path,
         monkeypatch: pytest.MonkeyPatch,
+        rank: int,
+        world: int,
+        expected_indices: list[int],
     ) -> None:
-        """Worker 0 of a 2-node partition with 3 shards renders shards 0 and 1 only.
+        """A worker renders (and uploads) exactly its contiguous slice of the 3 shards.
+
+        Covers in-range ranks that own a non-empty slice and an excess rank
+        (``world`` > num_shards) whose empty range renders nothing and makes no
+        rclone calls.
 
         :param patched_subprocess: Subprocess dispatcher used to introspect
-            renderer argv.
+            renderer argv; never invoked when ``expected_indices`` is empty.
         :param fake_r2_remote: Local-typed R2 remote — asserted to contain only
-            shards 0 and 1.
+            the worker's shards.
         :param tmp_path: Pytest tmp dir used by ``_multi_shard_spec``.
         :param monkeypatch: Used to set the rank/world env vars.
+        :param rank: This worker's ``SYNTH_SETTER_WORKER_RANK``.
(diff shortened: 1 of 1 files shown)


ktinubu:
## Verification Results

Behavior-preserving test refactor: the promise is that the two parametrized tests cover exactly the same scenarios as the five originals, with identical assertions, and that no test items are lost.

- [x] **Partition test covers all three original rank/world scenarios** *(Level 2 — invoked pytest on the affected file; the system under test is the test suite itself, so an end-to-end pipeline run would add noise without exercising the contract)*
  ```bash
  $ uv run pytest tests/pipeline/entrypoints/test_generate_dataset_unit.py -v 2>&1 | grep test_worker_renders_only_its_partition_of_shards
  ...::test_worker_renders_only_its_partition_of_shards[rank1-of-2-renders-remainder] PASSED
  ...::test_worker_renders_only_its_partition_of_shards[excess-worker-renders-none] PASSED
  ...::test_worker_renders_only_its_partition_of_shards[rank0-of-2-renders-first-half] PASSED
  ```
  The three ids map 1:1 to the deleted `test_rank_0_of_2_...`, `test_rank_1_of_2_...`, and `test_excess_worker_...`.

- [x] **Object-probe test covers both original absent/zero-size cases** *(Level 2)*
  ```bash
  $ uv run pytest tests/pipeline/entrypoints/test_generate_dataset_unit.py -v 2>&1 | grep test_run_renders_when_object_treated_as_absent
  ...::test_run_renders_when_object_treated_as_absent[object-absent] PASSED
  ...::test_run_renders_when_object_treated_as_absent[object-zero-size] PASSED
  ```
  `object-absent` (None) reproduces the autouse-fixture default of the deleted `test_run_renders_when_object_absent`; `object-zero-size` (0) reproduces `test_run_renders_when_object_zero_size`.

- [x] **No test items lost — coverage preserved 1:1** *(Level 2 — adversarial: collapsing functions must not silently drop cases)*
  ```bash
  $ git show main:tests/pipeline/entrypoints/test_generate_dataset_unit.py | grep -cE "^\s+def test_"
  84
  $ grep -cE "^\s+def test_" tests/pipeline/entrypoints/test_generate_dataset_unit.py
  81
  $ uv run pytest tests/pipeline/entrypoints/test_generate_dataset_unit.py --co -q 2>/dev/null | grep -c "::"
  86
  ```
  5 standalone functions collapsed into 2 parametrized functions (84 → 81 functions). Collected test *items* stay at 86: the 2 parametrized functions expand to 3 + 2 = 5 items, matching the 5 original standalone items exactly.

- [ ] **Full generate-dataset pipeline end-to-end** *(Level 1 — SKIP: this PR changes only test code, not the `generate()` production path; the existing same-file tests already exercise `generate()` against the `fake_r2_remote` local-typed remote, and the full distributed render pipeline requires RunPod/OCI compute. To exercise the real partitioning path locally: `SYNTH_SETTER_WORKER_RANK=0 SYNTH_SETTER_NUM_WORKERS=2 uv run synth-setter-generate-dataset experiment=generate_dataset/smoke-shard render.plugin_path="$SYNTH_SETTER_PLUGIN_PATH"`)*

CI (run_tests_ubuntu/macos/conda, code-quality) is green on this commit and already runs the full file under the package context — see the checks tab.

PR #1633, opened 2026-06-11
fix(monitoring): revert #1617 subprocess tee to W&B

### LINKED ISSUE #1634: bug(data-pipeline): _check_call_streamed SIGKILL orphans the headless-VST Xvfb tree, breaking later renders
## Symptom

`python -m synth_setter.tools.cadence_sweep_489 --size 5` (and other generate_dataset runs) intermittently die mid-render with:

```
synth_setter.data.vst.core:load_plugin:80 - Loading plugin plugins/Surge XT.vst3
X connection to :0.0 broken (explicit kill or server shutdown).
```

The renderer subprocess exits non-zero, so `generate_dataset` exits 1 and the cadence-sweep runner raises `CalledProcessError`.

## Root cause (regression from #1617)

`_check_call_streamed` (added in #1617) puts each child in its own process group (`start_new_session=True`) and, on timeout or abrupt exit, kills it with a bare `os.killpg(pid, signal.SIGKILL)`.

SIGKILL is uncatchable, so the headless-VST wrapper (`run-linux-vst-headless.sh`) never runs its `trap cleanup EXIT`. Its `Xvfb`/`openbox`/`xsettingsd`/`dbus` tree is orphaned and the `/tmp/.X11-unix/X*` socket leaks. Across repeated cadence-sweep runs (each probe's inline oracle-eval has a 600s timeout that can fire `killpg`), orphaned X servers and stale sockets accumulate and corrupt the shared X display, so a later render's connection to `:0` breaks.

Reproduced directly: timing out the real wrapper through `_check_call_streamed` with SIGKILL leaves a leaked socket + orphaned Xvfb/openbox/xsettingsd; SIGTERM instead lets the cleanup trap run (socket removed, tree reaped).

## Fix

SIGTERM the process group first (bash runs the wrapper's EXIT trap on SIGTERM → it reaps its X tree and removes the socket), then escalate to SIGKILL only if the group is still alive after a bounded grace period, keeping the timeout real for a wedged child.

Related: #583 (wrapper leaks on normal exit), #1605 (sweeps stay RUNNING).

ktinubu:
## Summary

Rolls back #1617 ("tee generate_dataset subprocess output to W&B"). The change **broke generation runs**, so we are reverting to restore working `generate_dataset` behavior while a corrected fix is developed.

Fixes https://github.com/tinaudio/synth-setter/issues/1634


ktinubu:
## Verification Results

This PR is a pure `git revert` of #1617 (squash commit `2962ad75`). The promise is **coherent restoration of the pre-#1617 state** — the entrypoint runs, no references to the removed symbols survive, and the affected test suites pass against the restored `subprocess` seam. Verified in a worktree on the branch HEAD (`93a23a1`).

- [x] **Reverted source + config are byte-identical to the pre-#1617 state** *(Level 5 — the promise is "exactly undo #1617"; a tree-identity check is the precise contract, and the pre-#1617 tree was green on `main` before the merge)*

  ```bash
  $ for f in src/synth_setter/cli/generate_dataset.py src/synth_setter/configs/logger/wandb.yaml; do
      git diff --quiet 2962ad75^ HEAD -- "$f" && echo "MATCH  $f" || echo "DIFFERS $f"; done
  MATCH  src/synth_setter/cli/generate_dataset.py
  MATCH  src/synth_setter/configs/logger/wandb.yaml
  ```

- [x] **No dangling references to the removed symbols (`_check_call_streamed`, `console_multipart`)** *(Level 4 — an incomplete revert would leave callers/config readers referencing deleted names; greps the whole tree, not just the diff)*

  ```bash
  $ grep -rn "_check_call_streamed\|console_multipart" --include="*.py" --include="*.yaml" --include="*.yml" --include="*.md" .
  $ echo "exit=$?"
  exit=1   # no matches anywhere
  ```

- [x] **Entrypoint + affected suites pass against the restored seam** *(Level 2 — invokes the real `generate_dataset` entrypoint, spec composition, rclone arg building, shard partitioning, inline-oracle-eval, and the logger-config compose pin; 106 tests across the three files #1617 touched)*

  ```bash
  $ uv run pytest tests/pipeline/entrypoints/test_generate_dataset_unit.py \
      tests/schemas/test_logger_config.py tests/test_generate_dataset.py -q -m "not slow"
  106 passed, 5 deselected in 13.96s
  ```

- [ ] **End-to-end real-VST → real-R2 generation (`test_generate_dataset_renders_shards_to_r2`)** *(Level 1 — SKIP: environment-blocked, NOT a revert defect)*
  Two clean runs on this branch drove the restored `subprocess.check_call` renderer seam end-to-end for **114s** (matching #1617's own ~113s), loading the real Surge XT plugin and rendering multiple real samples — visible renderer-subprocess output, including the loudness-skip path — before the **shared `:0` X display was torn down externally**:

  ```bash
  $ uv run pytest "tests/test_generate_dataset.py::test_generate_dataset_renders_shards_to_r2" -q
  ...
  | DEBUG | ...generate_sample:129 - loudness: -22.307665678041843
  | DEBUG | ...generate_sample:141 - loudness too low, skipping
  | DEBUG | ...generate_sample:129 - loudness: -45.6298164459444
  X connection to :0 broken (explicit kill or server shutdown).
  FAILED ... - subprocess.CalledProcessError
  1 failed in 114.74s (0:01:54)
  ```

  The same test on `origin/main` (the #1617 code) in the same environment also fails — there for a missing `plugins/` symlink in the throwaway worktree — confirming the e2e is environment-sensitive on **both** branches and the failure is unrelated to the revert. To run it cleanly, use a headless environment with a dedicated, stable Xvfb (the headless-VST wrapper targets `:0`, which a shared devcontainer can tear down mid-render):

  ```bash
  # In a headless env with the plugins/ symlink present and a private Xvfb on :0
  uv run pytest "tests/test_generate_dataset.py::test_generate_dataset_renders_shards_to_r2" -q
  ```

**Summary:** The revert is a verified, coherent inverse of #1617 — files restored byte-for-byte to the previously-green pre-#1617 state, no symbol leakage, and all 106 entrypoint/logger tests green. The only unchecked item is the full e2e, blocked by shared-`:0` X-display teardown in this worktree (reproduces on `main` too); CI / a clean headless env should run it before merge.

PR #1647, opened 2026-06-11
test(testing): raise xdist per-worker mem budget to measured 2 GiB peak

### LINKED ISSUE #1646: bug(testing): -n auto still OOMs — per-worker budget (1 GiB) is half the measured peak
## Symptom

`make test-fast` (`pytest -n auto`) still OOM-kills on the shared dev host even after the memory-aware clamp landed in #1524. On a 32-core / 31 GiB box with ~18 GiB `MemAvailable`, the clamp picks ~18 workers and the kernel OOM-killer reaps the run.

## Root cause

The clamp in `tests/conftest.py` divides available memory by `_DEFAULT_WORKER_MEM_MB` (default **1 GiB**) to bound the worker count. That budget is roughly **half** the real per-worker footprint:

Measured on the fast suite (2114 tests) under a per-process RSS monitor:

- Whole-tree peak RSS = **12,467 MB across 6 worker slots → ~2,080 MB per worker** (worker process + the torch-importing subprocesses some tests spawn).
- A single xdist worker's own RSS peaked at ~1,353 MB.

So with the 1 GiB budget the clamp over-provisions ~2x and OOMs.

## Fix

Raise `_DEFAULT_WORKER_MEM_MB` from 1024 → **2048**, grounded in the measured ~2 GiB per-worker-slot peak. On the shared host this takes `-n auto` from ~18 to ~7-9 workers; the env override (`PYTEST_XDIST_WORKER_MEM_MB`) is preserved.

## Follow-up (not in this fix)

`_available_memory_bytes()` trusts `MemAvailable`, which counts reclaimable cache as free; on a swap-thrashing host that is optimistic. Hardening that estimate is tracked separately.

Refs #1520.

ktinubu:
## Problem

`make test-fast` (`pytest -n auto`) still OOM-kills on the shared dev host even after the memory-aware clamp landed in #1524. The clamp divides available memory by a **1 GiB** per-worker budget, but that is roughly half a worker slot's real peak RSS, so it over-provisions ~2x and the kernel OOM-killer reaps the run.

## Measurement

Ran the fast suite (2114 tests) under a per-process RSS monitor at `-n 6`:

- Whole-tree peak RSS = **12,467 MB across 6 worker slots → ~2,080 MB per worker** (worker process + the torch-importing subprocesses some tests spawn).
- A single xdist worker's own RSS peaked at ~1,353 MB.

## Fix

Raise `_DEFAULT_WORKER_MEM_MB` from 1024 → **2048**, grounded in the measured ~2 GiB per-worker-slot peak. On a 31 GiB / 32-core shared host, `-n auto` now resolves to ~7-9 workers instead of ~18. The `PYTEST_XDIST_WORKER_MEM_MB` override is unchanged.

## Verification

- Full fast suite under `-n auto` with the new budget: **2111 passed, 3 skipped, no OOM, no worker crashes** (79s, resolved to 7 workers on this host).
- `tests/test__conftest_mem.py`: 22/22 green. The env-override test now uses 4096 (distinct from the 2 GiB default) so it fails if the override is ignored.

## Follow-up (not in this PR)

`_available_memory_bytes()` trusts `MemAvailable`, which counts reclaimable cache as free — optimistic on a swap-thrashing host. Hardening that estimate is separate; tracked under #1646's follow-up note.

Fixes #1646

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -949,9 +949,9 @@ def _cgroup_aware_cpu_count() -> int:
     return max(1, int(cpus))
 
 
-# Per-worker resident-memory budget for the -n auto memory clamp, overridable
-# via PYTEST_XDIST_WORKER_MEM_MB. 1 GiB suits a torch+h5py+Hydra worker.
-_DEFAULT_WORKER_MEM_MB = 1024
+# Per-worker resident-memory budget for the -n auto clamp (override via
+# PYTEST_XDIST_WORKER_MEM_MB); 2 GiB = measured peak RSS per worker slot (#1646).
+_DEFAULT_WORKER_MEM_MB = 2048
 
 # A v1 memory.limit_in_bytes at or above this is the kernel's unlimited sentinel.
 _MEM_UNLIMITED_SENTINEL = 1 << 62
@@ -1014,7 +1014,7 @@ def _available_memory_bytes() -> int | None:
 def _memory_aware_worker_count() -> int | None:
     """Cap ``-n auto`` workers by available memory and a per-worker budget.
 
-    Divides available memory by ``PYTEST_XDIST_WORKER_MEM_MB`` (default 1 GiB) so
+    Divides available memory by ``PYTEST_XDIST_WORKER_MEM_MB`` (default 2 GiB) so
     a busy shared host doesn't OOM-kill the run — the failure #1490's CPU clamp
     can't catch, since neither cpu.max nor memory.max is set on a shared host.
 
--- a/tests/test__conftest_mem.py
+++ b/tests/test__conftest_mem.py
@@ -114,17 +114,17 @@ class TestMemoryAwareWorkerCount:
     """Unit tests for _memory_aware_worker_count covering budget division."""
 
     def test_divides_avail_by_default_budget(self, monkeypatch: pytest.MonkeyPatch) -> None:
-        """20 GiB avail / 1 GiB default budget → 20 workers.
+        """20 GiB avail / 2 GiB default budget → 10 workers.
 
         :param monkeypatch: Pytest monkeypatch fixture.
         """
         monkeypatch.delenv("PYTEST_XDIST_WORKER_MEM_MB", raising=False)
         files = {_MEMINFO: _meminfo(20 * 1024 * 1024), _V2_MEM: "max\n"}
         with patch("builtins.open", side_effect=_make_open(files)):
-            assert _memory_aware_worker_count() == 20
+            assert _memory_aware_worker_count() == 10
 
     def test_fractional_floors_to_at_least_one(self, monkeypatch: pytest.MonkeyPatch) -> None:
-        """1.5 GiB avail / 1 GiB budget → floored to 1, never 0.
+        """1.5 GiB avail / 2 GiB budget → 0.75 floors to 0, then lifted to 1.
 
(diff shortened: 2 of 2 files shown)


ktinubu:
## Verification Results

Promise under test: raising the per-worker budget to the measured 2 GiB peak bounds `-n auto` to ~half the workers for a given memory snapshot, and the fast suite completes without the OOM-kill that motivated #1646.

- [x] **`-n auto` resolves to ~half the workers at the new budget (resource claim, before/after)** *(Level 2 — invokes the real clamp logic; the promise is the resolved worker count, not an end-to-end run)*
  ```bash
  $ uv run python -c "
  from tests.conftest import _available_memory_bytes, _cgroup_aware_cpu_count
  GiB = 1024**3
  avail = _available_memory_bytes()
  cpu = _cgroup_aware_cpu_count()
  old = max(1, avail // (1024 * 1024 * 1024))   # previous 1 GiB budget
  new = max(1, avail // (2048 * 1024 * 1024))    # this PR's 2 GiB budget
  print(f'MemAvailable snapshot: {avail/GiB:.1f} GiB   cpu_workers={cpu}')
  print(f'old budget 1 GiB -> mem_workers={old} -> -n auto = min({cpu},{old}) = {min(cpu,old)}')
  print(f'new budget 2 GiB -> mem_workers={new} -> -n auto = min({cpu},{new}) = {min(cpu,new)}')
  print(f'delta: {min(cpu,old)} -> {min(cpu,new)} workers (same MemAvailable snapshot)')
  "
  MemAvailable snapshot: 11.3 GiB   cpu_workers=32
  old budget 1 GiB -> mem_workers=11 -> -n auto = min(32,11) = 11
  new budget 2 GiB -> mem_workers=5 -> -n auto = min(32,5) = 5
  delta: 11 -> 5 workers (same MemAvailable snapshot)
  ```
  Same MemAvailable snapshot, divisor doubled → worker count halves (11 → 5), and it stays memory-bounded (`min(32, …)` picks the memory term). The absolute count tracks live memory; the **delta** is the contract.

- [x] **Fast suite completes under `-n auto` with no OOM-kill** *(Level 1 — end-to-end: drives the real suite the bug is about, on the PR branch)*
  ```bash
  $ uv run pytest -m "not slow and not gpu and not mps and not requires_vst" -n auto -q
  2111 passed, 3 skipped, 5 warnings in 56.87s
  $ grep -ic 'killed\|signal 9\|exit code 143\|out of memory' <run.log>
  0
  ```
  No exit-143 / SIGKILL / OOM markers; the run that previously got reaped now finishes green. (An earlier idle-host run resolved to 7 workers, 2111 passed in 79s — same outcome at a different memory snapshot.)

- [x] **Env override still discriminates from the default** *(Level 2 — unit test drives the real `_memory_aware_worker_count` with a faked `/proc/meminfo`)*
  ```bash
  $ uv run pytest tests/test__conftest_mem.py -q
  22 passed in 0.43s
  ```
  `test_env_budget_override_changes_divisor` now sets `PYTEST_XDIST_WORKER_MEM_MB=4096` (distinct from the 2 GiB default) → 16 GiB / 4 GiB = 4 workers, so it fails if the override is ignored.

### Not covered here
- The follow-up `MemAvailable`-overcount hardening (cache/swap optimism) is **out of scope** for this PR — noted in the description and #1646. This PR only retunes the divisor.

PR #1658, opened 2026-06-12
test(testing): debit swap-in-use from xdist available-memory estimate

### LINKED ISSUE #1657: bug(testing): xdist mem clamp trusts MemAvailable — over-provisions on a swapping host
## Symptom

The `-n auto` memory clamp in `tests/conftest.py` sizes worker count from `_available_memory_bytes()`, which uses the kernel's `MemAvailable`. `MemAvailable` counts reclaimable page cache as free, so on a shared host that is **already swapping**, the figure is optimistic: it reports headroom that does not really exist, and the clamp can still pick too many workers.

Concrete: on the dev host the per-worker budget bump in #1646/#1647 took `-n auto` from ~18 to ~7 workers and stopped the immediate OOM, but `_available_memory_bytes()` still read ~18 GiB `MemAvailable` while swap was 5.4 GiB used of 8 — so the estimate leans on memory that is under active pressure.

## Root cause

`_available_memory_bytes()` returns `min(host MemAvailable, cgroup limit)` with no adjustment for live memory pressure. A host that is swapping has anonymous pages parked on disk that will contend for RAM; `MemAvailable` does not net that out.

## Fix

Debit swap-in-use (`SwapTotal − SwapFree`, from `/proc/meminfo`) from the host `MemAvailable` term before the `min` with the cgroup cap. Properties:

- **No-op when the host is not swapping** — fully backward compatible.
- Only ever **lowers** the worker count, never raises it past today's clamp.
- Applies to the host figure only; the cgroup hard-cap is untouched.
- Floored at 0, so severe thrash degrades to a single worker rather than going negative.

No tuning knob needed; the `PYTEST_XDIST_AUTO_NUM_WORKERS` / `PYTEST_XDIST_WORKER_MEM_MB` overrides still apply.

Follow-on to #1646 (closed by #1647), which fixed the per-worker budget half of the same OOM.

ktinubu:
## Problem

Follow-up to #1646 (closed by #1647). That PR fixed the *per-worker budget* half of the `-n auto` OOM. This fixes the other half: `_available_memory_bytes()` sized workers from the kernel's `MemAvailable`, which counts reclaimable page cache as free. On a host **already swapping**, that figure is optimistic — it reports headroom that does not really exist, so the clamp can still over-provision.

Concrete: on the dev host `MemAvailable` read ~16 GiB while **7.6 GiB of swap was in use**.

## Fix

Debit swap-in-use (`SwapTotal − SwapFree`, from `/proc/meminfo`) from the host `MemAvailable` term before the `min` with the cgroup cap. Properties:

- **No-op when the host is not swapping** — fully backward compatible (every pre-existing test, which emits no swap lines, still passes unchanged).
- Only ever **lowers** the worker count, never raises it past today's clamp.
- Applies to the **host term only**; the cgroup hard-cap is untouched.
- Floored at 0, so severe thrash degrades to a single worker rather than going negative.
- No tuning knob; `PYTEST_XDIST_AUTO_NUM_WORKERS` / `PYTEST_XDIST_WORKER_MEM_MB` overrides still apply.

Note: swap is a host-global figure, so on a shared host this conservatively also counts siblings' swap — intentional over-conservatism for a test-runner clamp whose failure mode is OOM.

## Verification

- **Live on the dev host** (genuinely swapping): raw `MemAvailable` 15.7 GiB → swap-in-use 7.6 GiB → swap-adjusted **8.1 GiB**, resolved workers **7 → 4**.
- **Fast suite under `-n auto`** at the swap-adjusted count: **2178 passed, 3 skipped, 0 OOM** (~82s).
- `tests/test__conftest_mem.py`: **37 passed**, incl. a new `TestSwapInUseBytes` (total−free, fully-free→0, free>total clamp, both one-field-missing guards, malformed/valueless/unreadable error arms) and discount tests covering the host-only target, zero-floor, no-op-when-unswapped, and host-unreadable-with-swap-present branches.

## Out of scope

Cross-process memory contention on a shared host (siblings grabbing RAM mid-run) is not modelled — a collection-time snapshot can't see it. The swap debit is the tractable proxy for live pressure.

Fixes #1657
Refs #1646

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1011,6 +1011,29 @@ def _meminfo_available_bytes() -> int | None:
     return None
 
 
+def _swap_in_use_bytes() -> int | None:
+    """Read swap currently in use from ``/proc/meminfo`` (``SwapTotal - SwapFree``).
+
+    :returns: Bytes of swap in use, or None if either field is unreadable.
+    """
+    total: int | None = None
+    free: int | None = None
+    try:
+        with open("/proc/meminfo") as fh:
+            for line in fh:
+                if line.startswith("SwapTotal:"):
+                    total = int(line.split()[1]) * 1024  # field is in kibibytes
+                elif line.startswith("SwapFree:"):
+                    free = int(line.split()[1]) * 1024
+                if total is not None and free is not None:
+                    break
+    except (OSError, ValueError, IndexError):
+        return None
+    if total is None or free is None:
+        return None
+    return max(0, total - free)
+
+
 def _cgroup_memory_limit_bytes() -> int | None:
     """Read the cgroup memory limit in bytes, honouring v2 then v1.
 
@@ -1035,18 +1058,19 @@ def _cgroup_memory_limit_bytes() -> int | None:
 
 
 def _available_memory_bytes() -> int | None:
-    """Return usable memory as ``min(host MemAvailable, cgroup limit)``.
+    """Return usable memory as ``min(swap-adjusted host MemAvailable, cgroup limit)``.
 
-    On a shared host the cgroup limit is usually unset, so host MemAvailable is the real ceiling;
-    in a memory-capped container the cgroup limit wins.
+    MemAvailable counts reclaimable cache as free, so it overstates headroom on a swapping host;
+    used swap is debited from the host term before the min, a no-op when nothing is swapped.
 
     :returns: The tighter of the two figures in bytes, or None if neither is known.
     """
-    candidates = [
-        value
-        for value in (_meminfo_available_bytes(), _cgroup_memory_limit_bytes())
-        if value is not None
-    ]
+    host = _meminfo_available_bytes()
+    if host is not None:
+        swap_used = _swap_in_use_bytes()
+        if swap_used is not None:
+            host = max(0, host - swap_used)
+    candidates = [value for value in (host, _cgroup_memory_limit_bytes()) if value is not None]
     return min(candidates) if candidates else None
 
 
--- a/tests/test__conftest_mem.py
+++ b/tests/test__conftest_mem.py
@@ -11,6 +11,7 @@ import pytest
 from tests.conftest import (
     _available_memory_bytes,
     _memory_aware_worker_count,
+    _swap_in_use_bytes,
     pytest_xdist_auto_num_workers,
 )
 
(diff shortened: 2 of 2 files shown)


ktinubu:
## Verification Results

Promise under test: debiting swap-in-use from `MemAvailable` lowers the `-n auto` worker count when the host is under memory pressure (swapping), is a no-op when it isn't, and keeps the fast suite green with no OOM.

- [x] **Swap-in-use is debited → fewer workers on a swapping host (resource claim, before/after)** *(Level 2 — invokes the real clamp on the live host; the promise is the resolved worker count, not an end-to-end run)*
  ```bash
  $ uv run python -c "
  from tests.conftest import _meminfo_available_bytes, _swap_in_use_bytes, _available_memory_bytes, _memory_aware_worker_count
  GiB = 1024**3
  raw, swap, adj = _meminfo_available_bytes(), _swap_in_use_bytes(), _available_memory_bytes()
  print(f'raw MemAvailable      : {raw/GiB:.1f} GiB')
  print(f'swap-in-use           : {swap/GiB:.1f} GiB')
  print(f'swap-adjusted available: {adj/GiB:.1f} GiB  (debit {(raw-adj)/GiB:.1f} GiB)')
  print(f'workers from raw       : {max(1, raw // (2*GiB))}')
  print(f'workers swap-adjusted  : {_memory_aware_worker_count()}')
  "
  raw MemAvailable      : 10.4 GiB
  swap-in-use           : 8.0 GiB
  swap-adjusted available: 2.4 GiB  (debit 8.0 GiB)
  workers from raw       : 5
  workers swap-adjusted  : 1
  ```
  Captured while the box was under test load (swap fully used). Raw `MemAvailable` would yield 5 workers; the swap debit nets available to 2.4 GiB → 1 worker. (Idle, the same host showed 15.7 GiB raw / 7.6 GiB swap → 8.1 GiB → **7 → 4 workers**.) The debit equals swap-in-use exactly.

- [x] **No-op when the host is not swapping (backward compatibility)** *(Level 2 — drives the real `_available_memory_bytes` with a controlled `/proc/meminfo`)*
  ```bash
  $ uv run python -c "
  import sys, io; sys.path.insert(0, 'tests')
  from unittest.mock import patch
  from tests.conftest import _available_memory_bytes
  GiB = 1024**3
  def feed(body):
      def _open(path, *a, **k):
          if path == '/proc/meminfo': return io.StringIO(body)
          if path == '/sys/fs/cgroup/memory.max': return io.StringIO('max\n')
          raise OSError(path)
      return _open
  no_swap   = 'MemAvailable:   20971520 kB\nSwapTotal:      8388608 kB\nSwapFree:       8388608 kB\n'
  with_swap = 'MemAvailable:   20971520 kB\nSwapTotal:      8388608 kB\nSwapFree:       3145728 kB\n'
  with patch('builtins.open', side_effect=feed(no_swap)):
      a = _available_memory_bytes()
  with patch('builtins.open', side_effect=feed(with_swap)):
      b = _available_memory_bytes()
  print(f'swap fully free (0 used)  -> {a/GiB:.0f} GiB  (== raw 20 GiB, no-op: {a==20*GiB})')
  print(f'5 GiB swap in use         -> {b/GiB:.0f} GiB  (debited by 5 GiB: {b==15*GiB})')
  "
  swap fully free (0 used)  -> 20 GiB  (== raw 20 GiB, no-op: True)
  5 GiB swap in use         -> 15 GiB  (debited by 5 GiB: True)
  ```
  Fully-free swap leaves the figure identical to raw `MemAvailable` — the every-pre-existing-test path is unchanged.

- [x] **Fast suite stays green with no OOM under the swap-adjusted clamp** *(Level 1 — end-to-end: drives the real suite, on the PR branch, while the host was actually swapping)*
  ```bash
  $ uv run pytest -m "not slow and not gpu and not mps and not requires_vst" -n auto -q
  2184 passed, 3 skipped, 5 warnings in 77.12s
  $ grep -ic 'killed\|signal 9\|exit code 143\|out of memory' <run.log>
  0
  ```
  No exit-143 / SIGKILL / OOM markers.

- [x] **New error/boundary branches are pinned by tests** *(Level 2 — the targeted unit suite)*
  ```bash
  $ uv run pytest tests/test__conftest_mem.py -q
  37 passed in 0.43s
  ```
  `TestSwapInUseBytes` covers total−free, fully-free→0, free>total clamp, both one-field-missing guards, and the malformed/valueless/unreadable error arms; the discount tests cover host-only targeting, the zero floor, the no-op, and the host-unreadable-with-swap-present branch.

Issue #1599, opened 2026-06-10
feat(training): Lance dataloader backend for train and eval (datamodule=surge_lance)


ktinubu:
Support loading Surge XT datasets from [Lance](https://github.com/lance-format/lance) datasets in addition to HDF5 shards, for both the train and eval entrypoints.

## Scope

- `LanceDataModule` / `LanceXTDataset` reading `train/val/test.lance` splits with the same batch-per-index contract, stats normalization, OT matching, and predict-split semantics as the HDF5 path.
- Hydra config `datamodule=surge_lance` usable by both `synth_setter.cli.train` and `synth_setter.cli.eval`.
- `pylance` dependency in the `data` group.

Part of #253

PR #1670, opened 2026-06-12
internal-fix(data): tighten Lance dataloader types and read assertions

ktinubu:
## Summary

Follow-up to the multi-skill review of #1601 (the now-merged Lance dataloader backend). Applies the advisory findings I agreed with — **no behavior changes**, just types, a small readability dedup, and stricter test assertions.

- **`surge_datamodule.py`** — annotate `VSTDataset.__getitem__` return as `dict[str, torch.Tensor | None]`. The `idx` param type was widened in #1601 but the return stayed bare; sibling readers (`LanceColumn.__getitem__`, `_index_dataset`) already carry explicit returns.
- **`lance_datamodule.py`** — build `Path(path)` once in `LanceShardFile.__init__` and reuse it for the directory / missing-file boundary checks instead of reconstructing it twice.
- **`tests/data/test_lance_datamodule.py`** — use `np.testing.assert_array_equal` for the bit-exact read-fidelity assertions (float16 `audio` / float32 columns round-trip losslessly through Lance). Expresses the no-tolerance intent that `assert_allclose` (default `rtol=1e-7`) obscured, matching the writer-compatibility test.
- **`tests/conftest.py`** — drop the redundant `:rtype: RenderConfig` from `_smoke_fake_render_cfg` (pydoclint `check-return-types` is off, so the `->` annotation is the single source).

## Considered but deferred

- **m2l column not emitted by the production Lance writer** (`conditioning="m2l"` would `KeyError` on a real shard) — a real gap, but it belongs in the pipeline writer (#1642), not this dataloader cleanup. Worth a tracked follow-up on #1599.
- **Pre-existing missing annotations** on `_get_fake_item` / `_load_dataset_statistics` / `__len__` — out of scope (not introduced by #1601).
- **`Mapping` vs `dict` return type** — kept `dict`; the value is a freshly-constructed mapping callers consume directly, consistent with `_get_fake_item`'s `dict(...)`.

## Review gate note

Opened with `REVIEW_COMMENT_GATE=warn`. The pre-PR review's only comment-hygiene finding is the pre-existing `:param tmp_path:` boilerplate shared by every test in the file — my diff does not touch those docstring lines, and pydoclint (`DOC101`) requires them, so the suggested removal is not applicable here.

## Testing

- `tests/data/test_lance_datamodule.py` + `tests/data/test_surge_datamodule.py`: 113 passed.
- `make format`: all hooks pass (ruff, pyright, pydoclint, interrogate).

Refs #1599

### DIFF
--- a/src/synth_setter/data/lance_datamodule.py
+++ b/src/synth_setter/data/lance_datamodule.py
@@ -90,13 +90,14 @@ class LanceShardFile:
             writes) — a stable contract independent of which exception
             ``LanceFileReader`` raises.
         """
+        path = Path(path)
         self._path = str(path)
-        if Path(path).is_dir():
+        if path.is_dir():
             raise ValueError(
                 f"expected a single-file Lance shard, got a directory "
                 f"(legacy Lance dataset layout?): {self._path}"
             )
-        if not Path(path).is_file():
+        if not path.is_file():
             raise ValueError(f"Lance shard file was not found: {self._path}")
         metadata = LanceFileReader(self._path).metadata()
         # The shard is immutable, so row count and per-column tensor shapes are
--- a/src/synth_setter/data/surge_datamodule.py
+++ b/src/synth_setter/data/surge_datamodule.py
@@ -170,7 +170,7 @@ class VSTDataset(torch.utils.data.Dataset):
 
         return ds[idx]
 
-    def __getitem__(self, idx: int | Sequence[int] | np.ndarray):
+    def __getitem__(self, idx: int | Sequence[int] | np.ndarray) -> dict[str, torch.Tensor | None]:
         if self.fake:
             return self._get_fake_item()
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -696,7 +696,6 @@ def _smoke_fake_render_cfg(param_spec_name: str) -> RenderConfig:
     :param param_spec_name: Key into :data:`synth_setter.data.vst.param_specs` and
         :data:`synth_setter.data.vst.preset_paths` selecting spec and preset.
     :returns: A CPU ``RenderConfig`` with the GUI toggle disabled.
-    :rtype: RenderConfig
     """
     return RenderConfig(
         plugin_path=PLUGIN_PATH,
--- a/tests/data/test_lance_datamodule.py
+++ b/tests/data/test_lance_datamodule.py
@@ -192,7 +192,7 @@ class TestLanceShardFile:
         out = shard["mel_spec"][2:5]
         assert out.shape == (3, _MEL_CHANNELS, _MEL_N_MELS, _MEL_N_FRAMES)
         assert out.dtype == columns["mel_spec"].dtype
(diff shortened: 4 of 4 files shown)


ktinubu:
## Verification Results

This PR is behavior-preserving cleanup (return annotation, `Path` dedup, `assert_allclose`→`assert_array_equal`, `:rtype:` removal), so each promise is narrow. Verified at `e0933516`.

- [x] **The new `__getitem__` return annotation is type-correct (pyright accepts it)** *(Level 2 — invoked the type checker on the changed file; promise is tool acceptance, not end-to-end)*
  ```bash
  $ uv run pyright src/synth_setter/data/surge_datamodule.py 2>&1 | tail -1
  ```

- [x] **Reads are genuinely bit-exact, so `assert_array_equal` holds without tolerance** *(Level 3 — exercises real Lance writes + reads; a lossy round-trip would now FAIL where `assert_allclose`'s `rtol=1e-7` masked it)*
  ```bash
  $ uv run python -m pytest tests/data/test_lance_datamodule.py::TestLanceShardFile \
      tests/data/test_lance_datamodule.py::TestLanceVSTDataset -q 2>&1 | tail -1
  28 passed in 1.73s
  ```

- [x] **`Path(path)` dedup preserves the directory / missing-file error contract** *(Level 3 — drives the real `LanceShardFile.__init__` boundary checks with a directory and a missing path)*
  ```bash
  $ uv run python -m pytest \
      tests/data/test_lance_datamodule.py::TestLanceShardFile::test_directory_path_raises_value_error \
      tests/data/test_lance_datamodule.py::TestLanceShardFile::test_missing_shard_file_raises_value_error -v 2>&1 | grep -E "PASSED|passed"
  tests/data/test_lance_datamodule.py::TestLanceShardFile::test_missing_shard_file_raises_value_error PASSED
  tests/data/test_lance_datamodule.py::TestLanceShardFile::test_directory_path_raises_value_error PASSED
  2 passed in 1.47s
  ```

- [x] **The Lance dataloader still works end-to-end through real `train(cfg)` and `evaluate(cfg)`** *(Level 1 — promise is that the cleanup leaves the dataloader's end-to-end behavior intact; these compose `train.yaml`/`eval` with `datamodule=surge_lance` and run real CPU steps)*
  ```bash
  $ uv run python -m pytest tests/test_train.py::test_train_fast_dev_run_lance_datamodule -p no:randomly -q 2>&1 | tail -1
  1 passed in 3.48s
  $ uv run python -m pytest tests/test_eval.py::test_evaluate_validate_mode_lance_datamodule_runs_oracle -p no:randomly -q 2>&1 | tail -1
  1 passed in 11.77s
  ```

### Note on the multi-worker test
`TestLanceVSTDataModule::test_val_dataloader_multi_worker_matches_single_worker` flakes intermittently on this memory-constrained host with a forked-worker Lance panic (`range end out of bounds`). It is pre-existing (reproduces on `main` without this PR's changes), passes in isolation 5/5, and this PR touches none of the `read_range` slice path it exercises. Tracked separately.

PR #1701, opened 2026-06-15
test: deflake clean-exit sweep test via heartbeat liveness

### LINKED ISSUE #1700: fix(testing): flaky test_clean_exit_sweep_reaps_ingroup_grandchild — bare-PID liveness probe hits PID reuse
## Symptom

`tests/pipeline/test_subprocess_stream.py::TestPostExitSweep::test_clean_exit_sweep_reaps_ingroup_grandchild` flakes intermittently on CI with `post-exit sweep missed the in-group grandchild`, while staying green locally.

## Root cause

The test asserts the swept grandchild is dead via a **bare-PID liveness probe** (`_pid_alive(grandchild_pid)` on the pid read from a pidfile). The SUT (`check_call_streamed_async`) is correct — its `finally` unconditionally `os.killpg(pgid, SIGTERM)`, and the grandchild is always an in-group member, so it always dies. The flake is a **false positive in the probe**: after the grandchild is SIGTERM'd and reaped, its PID can be **recycled** by an unrelated process inside the 3 s poll window, so `_pid_alive` reports a stranger as alive.

This is load-correlated: green on hosts with `pid_max=4194304` (sequential allocation can't recycle within 3 s — verified 2640/2640 under `-n auto` + 120× repeat), intermittent on busy CI with smaller `pid_max` and `-n auto` fork churn.

## Fix

Read death from a **heartbeat file only the grandchild writes**, mirroring the sibling `test_timeout_group_kill_stops_ingroup_grandchild_heartbeat`. A recycled PID is a different process that won't grow our file, so the check is identity-bound and immune to PID reuse. The clean-exit trigger is unchanged.

Related prior hardening: #1655 / #1656 (zombie handling in `_pid_alive`).

ktinubu:
## Problem

`tests/pipeline/test_subprocess_stream.py::TestPostExitSweep::test_clean_exit_sweep_reaps_ingroup_grandchild` flakes intermittently on CI (`post-exit sweep missed the in-group grandchild`) while staying green locally.

## Root cause

The test confirmed the swept grandchild was dead via a **bare-PID liveness probe** (`_pid_alive(grandchild_pid)` on a pid read from a pidfile). The SUT is correct — `check_call_streamed_async`'s `finally` unconditionally `os.killpg(pgid, SIGTERM)`, and the grandchild (no `setsid`, no SIGTERM handler) is always an in-group member, so it always dies. The flake is a **false positive in the probe**: after the grandchild is SIGTERM'd and reaped, its PID can be **recycled** by an unrelated process inside the 3 s poll window, so `_pid_alive` reports a stranger as alive.

This is load-correlated — green on hosts with `pid_max=4194304` (sequential allocation can't recycle within 3 s), intermittent on busy CI with smaller `pid_max` and `-n auto` fork churn.

## Fix

Read death from a **heartbeat file only the grandchild writes**, mirroring the sibling `test_timeout_group_kill_stops_ingroup_grandchild_heartbeat`: assert the file stops growing. A recycled PID is a different process that won't grow our file, so the check is identity-bound and immune to PID reuse. The clean-exit trigger (parent forks, exits 0) is unchanged, so the post-exit sweep remains the only thing that can stop the heartbeat. The two heartbeat tests now share a named `_SWEEP_SETTLE_SECONDS` window, and the now-~4 s test gains `@pytest.mark.slow` to match its sibling.

## Verification

- Local reproduction could not surface it (this host has `pid_max=4194304`): **2640/2640** passed under `-n auto` + 120× repeat, plus 90+ isolated/loaded/parallel runs — consistent with the PID-reuse mechanism being container-`pid_max`-dependent.
- Converted test green in isolation, 20/20 under 32-way CPU load, and full module **27 passed, 1 xfailed**.

## Deliberately not addressed (advisory review WARNs)

- **First-write race on `_wait_for_file`**: the SUT's 2 s post-exit pipe drain (the grandchild holds the inherited pipe open) gates the SIGTERM, so the grandchild writes its first heartbeat ~2 s before the sweep — the race cannot fire in practice.
- **DRY of the heartbeat-freeze idiom** across the two tests: only two occurrences; extracting a helper is out of scope for a deflake.

Fixes #1700

### DIFF
--- a/tests/pipeline/test_subprocess_stream.py
+++ b/tests/pipeline/test_subprocess_stream.py
@@ -52,6 +52,9 @@ _PROMPT_SECONDS = 10.0
 # the child interpreter startup margin on loaded CI runners.
 _POLICY_TIMEOUT_SECONDS = 2.0
 
+# Per-phase heartbeat-sweep window (seconds); must exceed the grandchild's write cadence.
+_SWEEP_SETTLE_SECONDS = 1.0
+
 
 def _streamed(
     cmd: list[str], *, timeout: float | None = None, env: dict[str, str] | None = None
@@ -611,9 +614,9 @@ class TestTimeoutEscalation:
             _streamed(_child(script, leak_marker), timeout=_POLICY_TIMEOUT_SECONDS)
 
         _wait_for_file(heartbeat)
-        time.sleep(1.0)
+        time.sleep(_SWEEP_SETTLE_SECONDS)
         size_after_kill = heartbeat.stat().st_size
-        time.sleep(1.0)
+        time.sleep(_SWEEP_SETTLE_SECONDS)
         assert heartbeat.stat().st_size == size_after_kill, "grandchild still heartbeating"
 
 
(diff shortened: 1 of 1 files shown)


ktinubu:
## Verification Results

Behavioral verification of PR #1701 — run in a worktree at the pushed HEAD `f6087e5`. The promise: the test no longer false-flakes (it is identity-bound, immune to PID reuse) **and** it still genuinely detects a leaked grandchild.

- [x] **Adversarial: the test detects a real missed sweep (not a tautology)** *(Level 1 — end-to-end: real `sys.executable -c` child, real `os.fork` grandchild, real SUT, with only the post-exit sweep neutralized)*
  Neutralized the SUT's group signal, then ran the test — it MUST fail:
  ```bash
  $ sed -i 's/        os.killpg(pgid, sig)/        _ = (pgid, sig)  # VERIFY: sweep neutralized/' \
      src/synth_setter/pipeline/subprocess_stream.py
  $ grep -n "VERIFY: sweep neutralized" src/synth_setter/pipeline/subprocess_stream.py
  249:        _ = (pgid, sig)  # VERIFY: sweep neutralized
  $ uv run pytest tests/pipeline/test_subprocess_stream.py -o addopts="" -p no:randomly \
      -k clean_exit_sweep_reaps_ingroup_grandchild -q
  tests/pipeline/test_subprocess_stream.py:679: AssertionError      # heartbeat kept growing
  FAILED tests/pipeline/test_subprocess_stream.py::TestPostExitSweep::test_clean_exit_sweep_reaps_ingroup_grandchild
  1 failed, 27 deselected, 1 error in 7.84s
  $ git checkout -- src/synth_setter/pipeline/subprocess_stream.py   # reverted clean
  ```
  With the sweep gone, the grandchild keeps heartbeating, the size-stability assertion (L679) fires, and the autouse `leak_marker` fixture additionally reports the leak (the trailing `error`). The assertion has teeth.

- [x] **Deflaked test is stable under load** *(Level 2 — invoke the test directly; promise is "no longer flakes". Reproduces the CI condition: 32-way CPU saturation + repeat)*
  ```bash
  $ for i in $(seq 1 $(nproc)); do yes > /dev/null & done   # 32-way load
  $ uv run pytest tests/pipeline/test_subprocess_stream.py -o addopts="" -p no:randomly \
      -k clean_exit_sweep_reaps_ingroup_grandchild --count=10 -q
  10 passed, 270 deselected in 41.79s
  ```

- [x] **No collateral breakage in the module (incl. the sibling that now shares `_SWEEP_SETTLE_SECONDS` and the xfail)** *(Level 2 — run the full module with slow enabled)*
  ```bash
  $ uv run pytest tests/pipeline/test_subprocess_stream.py -o addopts="" -p no:randomly -q
  27 passed, 1 xfailed in 42.82s
  ```

### Notes

- **PID-reuse immunity is structural, not directly exercised.** Forcing PID recycling requires a low `pid_max`; this host is `pid_max=4194304` and the sandbox cannot lower it (no `unshare`/sysctl write). The immunity follows from the heartbeat being written only by *this* grandchild — a recycled PID is a different process that cannot grow the file. Check 1 (adversarial) confirms the assertion tracks *this* grandchild's writes specifically.
- The `1 xfailed` is the intentional `test_escaped_setsid_grandchild_is_reaped` (`strict=True`) documenting the known `setsid()` escape gap — unchanged by this PR.

Issue #1677, opened 2026-06-12
fix(testing): flaky test_datasets_from_sampled_params_are_identical — RMS-envelope cosine 0.7228 below 0.8 threshold


ktinubu:
## Summary

`tests/data/vst/test_generate_vst_dataset.py::test_datasets_from_sampled_params_are_identical` flaked in the **VST slow tests (Docker) — deflake** job: a Stage-1-vs-Stage-2 replay pair tripped the RMS-envelope cosine-similarity floor.

```
E   AssertionError: sample 0: rms cosine similarity 0.7228 below 0.8
tests/data/vst/test_generate_vst_dataset.py:416: AssertionError
```

- Failing run: https://github.com/tinaudio/synth-setter/actions/runs/27419535858/job/81041109208 (on PR #1673)
- Threshold: `_RMS_MIN_COSINE = 0.8` (test module constant), observed `0.7228` on `sample 0`
- Assertion site: `_assert_audio_metrics_within_thresholds` (`test_generate_vst_dataset.py:416`)
- The PR it failed on (#1673, a `generate_dataset` config change) does not touch the render path — this is a pre-existing flake surfaced by the deflake rerun, not a regression.

## Root cause (suspected)

This is the same family as the now-closed #489: Surge XT's oscillator phase init is nondeterministic across renders of identical params, so two renders of the same patch are phase-shifted variants of the same waveform. The test deliberately uses phase-robust perceptual metrics (MSS / wMFCC / SOT / RMS-envelope cosine) instead of element-wise equality.

RMS-envelope cosine is the tightest of the four here. For at least one random-sampled patch, the every-other-render amplitude-envelope variance (#489 fingerprint) pushed the per-row cosine to `0.7228`, below the `0.8` floor. Unlike `test_datasets_from_hardcoded_params_are_identical` (fixed, version-controlled patch), this test draws random params each run, so the failing patch is not reproducible from the logs alone — it's whichever Stage-1 candidate happened to land in a phase-sensitive region.

## Why it matters

A randomly-sampled patch can legitimately fall in a region where RMS-envelope variance exceeds the current floor, so the test is inherently flaky on the `sample N` rows whenever the draw is unlucky.

## Possible fixes (to be decided)

- Re-examine whether `_RMS_MIN_COSINE = 0.8` is the right floor for **randomly-sampled** patches (the hardcoded-patch test can keep a tighter bound since its input is pinned).
[...]

PR #1706, opened 2026-06-15
fix(testing): decouple oracle shuffle-probe rms floor from shared bound

### LINKED ISSUE #1703: fix(testing): flaky test_oracle_eval_inline_writes_shuffled_audio_metrics_when_params_uniform — RMS-envelope cosine 0.9472 below 0.95 floor
## Summary

`tests/test_generate_dataset.py::test_oracle_eval_inline_writes_shuffled_audio_metrics_when_params_uniform` is flaky: the oracle round-trip's RMS-envelope cosine (`rms_mean`) tripped the shared `ORACLE_AUDIO_METRIC_BOUNDS.rms_min = 0.95` floor.

```
observed rms_mean ≈ 0.9472  (floor 0.95) — a ~0.3% boundary miss
```

- Assertion site: `tests/test_generate_dataset.py` (`rms_mean > bounds.rms_min`)
- Threshold: shared `ORACLE_AUDIO_METRIC_BOUNDS.rms_min = 0.95` (`tests/evaluation/_oracle_helpers.py:54`)

## Root cause

Same family as #1677 / the closed #489. The test forces `render.param_sample_cadence=shard`, so every row in the shard reuses **one** patch drawn from the **un-seeded** global RNG (`spec.py:318` documents "drawn fresh each run (no seeding)"; `ShardSpec.seed` is computed but never applied to the render path).

`surge/fake_oracle` returns `batch["params"]` verbatim, so the param round-trip is exact (pinned bit-for-bit in `tests/test_train.py`). The `rms_mean` dip is Surge XT's per-voice render jitter (oscillator phase, noise seed) on a single unlucky patch — and because cadence=shard reuses one patch, there is no cross-patch averaging to damp it. So a noise-/transient-heavy draw can pull the mean below 0.95. This is render variance, not a regression.

## Why it matters

A randomly-sampled patch can legitimately land in a phase-sensitive region where RMS-envelope variance exceeds the tight shared floor, so the test is inherently flaky and blocks CI on unlucky draws.

## Fix

Decouple this probe's `rms` floor from the shared `ORACLE_AUDIO_METRIC_BOUNDS.rms_min` (kept tight for `test_train.py`'s fixture-based test) to a looser, corruption-catching `_SHUFFLE_PROBE_RMS_MIN` local to this test. The finite-float plumbing checks and the `mss`/`wmfcc`/`sot` upper bounds (large headroom) are untouched; a corrupted round-trip collapses the envelope cosine toward 0, so the looser floor still catches real breakage.

## Related

- #1677 — flaky `test_datasets_from_sampled_params_are_identical` (same RMS-envelope-cosine family)
- #489 — pedalboard render variance with identical patches (closed; documents the phase nondeterminism)

ktinubu:
## Summary

`tests/test_generate_dataset.py::test_oracle_eval_inline_writes_shuffled_audio_metrics_when_params_uniform` is flaky: its oracle round-trip RMS-envelope cosine (`rms_mean`) tripped the shared `ORACLE_AUDIO_METRIC_BOUNDS.rms_min = 0.95` floor at **0.9472** — a ~0.3% boundary miss.

This decouples that single assertion from the shared bound: `rms_mean` is now floored at a test-local `_SINGLE_PATCH_RMS_MIN = 0.5`. The finite-float plumbing checks and the `mss`/`wmfcc`/`sot` upper bounds stay on the shared `ORACLE_AUDIO_METRIC_BOUNDS`, which remains tight for `test_train.py`'s fixture-based test.

## Why this is flaky, not a bug

- The test forces `render.param_sample_cadence=shard`, so every row in the shard reuses **one** patch drawn from the **un-seeded** global RNG (`spec.py:318` documents "drawn fresh each run (no seeding)"; `ShardSpec.seed` is computed but never applied to the render path).
- `surge/fake_oracle` returns `batch["params"]` verbatim, so the param round-trip is exact (pinned bit-for-bit in `test_train.py`). The `rms_mean` dip is Surge XT's per-voice render jitter (oscillator phase, noise seed) on a single unlucky patch — and because cadence=shard reuses one patch, there is no cross-patch averaging to damp it.
- This is the same family as #1677, where the same RMS-envelope-cosine metric dipped to **0.7228** on an un-seeded single patch. 0.5 leaves margin below that observed tail while a corrupted/silent envelope (cosine ≈ 0) still fails.

## Deliberate decisions

- **Looser floor applies to both `audio/` and `shuffled_audio/` groups.** Under `cadence=shard` both passes are single-patch, so both lack cross-patch averaging and both are subject to the jitter. The tight `rms_min = 0.95` guard for the non-shuffled pass is preserved by the sibling default-cadence test (`test_oracle_eval_inline_writes_bounded_audio_metrics…`) and by `test_train.py:394` — so no real coverage is lost here.
- **`0.5` rather than a tighter ~0.65.** The point of this change is to stop flaking; #1677's empirical tail (~0.72) leaves too little margin for a 0.6–0.65 floor. The strict fidelity guard lives in the deterministic-input tests, not this single-patch probe.
- **Root cause (un-seeded render sampling) left in place.** Seeding the render path so the strict 0.95 floor could stand pipeline-wide is the larger fix; it is out of scope here and tracked separately. `ShardSpec.seed` is computed but never consumed — a good starting point for that work.

## Testing

The affected test is gated behind `@pytest.mark.requires_vst` / `r2` / `slow` and cannot run in CI's CPU-fast loop or locally without R2 + the Surge XT VST. The change is mechanical (a threshold constant + assertion) and passes `ruff`, `ruff-format`, `pydoclint`, and `codespell`.

Fixes #1703

### DIFF
--- a/tests/test_generate_dataset.py
+++ b/tests/test_generate_dataset.py
@@ -48,6 +48,9 @@ from tests.helpers.dummy_shards import stub_renderer
 # only keys in ``metrics.json`` (see ``synth_setter.evaluation.compute_audio_metrics``).
 _ORACLE_AUDIO_METRICS = ("mss", "wmfcc", "sot", "rms")
 
+# Min RMS-envelope cosine for a single-patch render: below #1677 jitter, above silence (~0).
+_SINGLE_PATCH_RMS_MIN = 0.5
+
 _REPO_ROOT = Path(__file__).resolve().parents[1]
 
 # Moduleinfo-only VST3 bundle: extract_renderer_version reads its
@@ -479,10 +482,9 @@ def test_oracle_eval_inline_writes_shuffled_audio_metrics_when_params_uniform(
     ``shuffled_audio/<name>_{mean,std}``.
 
     Asserts each audio metric (mss, wmfcc, sot, rms) produces a finite, bounded
-    value under both the ``audio/`` and ``shuffled_audio/`` prefixes.  Because
-    all samples share one patch, shuffled predictions match the same target as
-    the originals, so the shuffled means satisfy the same
-    ``ORACLE_AUDIO_METRIC_BOUNDS`` envelope.
+    value under both the ``audio/`` and ``shuffled_audio/`` prefixes.  mss/wmfcc/sot
+    use the shared ``ORACLE_AUDIO_METRIC_BOUNDS``; both groups' ``rms`` uses the
+    looser ``_SINGLE_PATCH_RMS_MIN`` (cadence=shard renders one patch; see #1677).
 
     :param cfg_dataset: Composed config; read only for ``r2.bucket`` (cleanup
         purge).
@@ -553,8 +555,6 @@ def test_oracle_eval_inline_writes_shuffled_audio_metrics_when_params_uniform(
                             f"(split={split}, metrics={metrics})"
                         )
 
-            # Uniform params → shuffled pred matches the same target; means satisfy
-            # the same oracle envelope as the non-shuffled pass.
             for group in ("audio", "shuffled_audio"):
                 assert metrics[f"{metric_prefix}{group}/mss_mean"] < bounds.mss_max, (
                     split,
@@ -568,8 +568,9 @@ def test_oracle_eval_inline_writes_shuffled_audio_metrics_when_params_uniform(
                     split,
                     metrics,
                 )
-                assert metrics[f"{metric_prefix}{group}/rms_mean"] > bounds.rms_min, (
+                assert metrics[f"{metric_prefix}{group}/rms_mean"] > _SINGLE_PATCH_RMS_MIN, (
                     split,
+                    _SINGLE_PATCH_RMS_MIN,
                     metrics,
                 )
     finally:


ktinubu:
## Verification Results

Both VST (`/usr/lib/vst3/Surge XT.vst3`) and R2 (`rclone lsd r2:` ✓) are reachable in this devcontainer, so the affected test was run end-to-end rather than skipped.

- [x] **The shuffle-probe test passes end-to-end with the decoupled rms floor** *(Level 1 — runs the real `generate_dataset` CLI subprocess: Surge XT render → R2 upload → inline oracle eval → `shuffled_audio/*` metrics → the changed assertion, on a freshly drawn un-seeded patch)*
  ```bash
  $ uv run pytest "tests/test_generate_dataset.py::test_oracle_eval_inline_writes_shuffled_audio_metrics_when_params_uniform" \
      -v -o "addopts=--color=yes --strict-markers"
  tests/test_generate_dataset.py::...PASSED [100%]
  ======================== 1 passed in 471.30s (0:07:51) =========================
  ```

- [x] **No longer flaky across repeated un-seeded draws — 4/4 passes** *(Level 1 — the initial run plus 3 back-to-back reruns, each drawing a genuinely different patch since the Surge param draw uses the un-seeded global RNG, independent of pytest-randomly's seed)*
  ```bash
  $ for i in 1 2 3; do uv run pytest "tests/test_generate_dataset.py::test_oracle_eval_inline_writes_shuffled_audio_metrics_when_params_uniform" -q; done
  run 1: PASS   # 1 passed in 426.68s
  run 2: PASS   # 1 passed in 409.16s
  run 3: PASS   # 1 passed in 431.90s
  # + initial run: 1 passed in 471.30s  →  4/4 green
  ```
  Margin rationale: the prior failure was `rms_mean ≈ 0.9472` vs the old floor `0.95`; the sibling test #1677 (same RMS-envelope-cosine family) bottomed at `0.7228`. The new floor `_SINGLE_PATCH_RMS_MIN = 0.5` sits ≥0.22 below that empirical tail, while a corrupted/silent envelope yields cosine ≈ 0 and still fails.

- [x] **Tight `rms_min = 0.95` coverage on the non-shuffled path is preserved elsewhere** *(Level 5 — assertion-structure check; the sibling default-cadence test keeps the shared bound, so loosening both groups here loses no coverage)*
  ```bash
  $ grep -n "_SINGLE_PATCH_RMS_MIN" tests/test_generate_dataset.py
  53:_SINGLE_PATCH_RMS_MIN = 0.5
  573:                assert metrics[f"{metric_prefix}{group}/rms_mean"] > _SINGLE_PATCH_RMS_MIN, (
  $ grep -n "rms.*min().*rms_min" tests/test_train.py
  394:        assert per_sample["rms"].min() > bounds.rms_min, (
  ```
  (The default-cadence sibling `test_oracle_eval_inline_writes_bounded_audio_metrics…` and `test_train.py:394` both retain the strict 0.95 floor.)

> Note: this test is excluded from CI's CPU-fast loop (`requires_vst` / `r2` / `slow`), so these local runs are the only behavioral coverage of the change. CI is otherwise fully green, including the Docker+VST dataset-generation job.

PR #1721, opened 2026-06-16
test(training): cover Lance datamodule smoke parity

### LINKED ISSUE #1719: test(training): cover Lance datamodule smoke parity
Track adding train/eval smoke coverage so the Lance datamodule exercises the same entrypoint paths as the Surge H5 datamodule.

ktinubu:
## Summary

- add Lance counterparts for the Surge smoke train and train->predict fixtures
- run the Surge smoke experiment matrix over `datamodule=surge_lance`
- parametrize fast eval entrypoint coverage across H5 and Lance datasets

## Validation

- `pre-commit run --files tests/conftest.py tests/test_eval.py tests/test_train.py`
- `uv run pytest tests/test_train.py::test_train_surge_lance tests/test_train.py::test_train_eval_surge_lance tests/test_eval.py::test_evaluate_predict_mode_merges_audio_metrics_into_metric_dict tests/test_eval.py::test_evaluate_predict_mode_logs_per_sample_metrics_table_to_wandb tests/test_eval.py::test_evaluate_predict_mode_logs_shuffle_permutation_table_to_wandb tests/test_eval.py::test_evaluate_validate_mode_legacy_val_spelling_runs_oracle tests/test_eval.py::test_evaluate_predict_mode_includes_shuffled_audio_metrics_when_subprocess_writes_shuffled_csv tests/test_eval.py::test_evaluate_validate_mode_lance_datamodule_runs_oracle -q`
- `/repo-review-full-no-comments`: PASS, 0 BLOCK / 0 WARN

Fixes #1719

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -8,7 +8,7 @@ import sys
 from collections.abc import Callable, Iterator
 from datetime import UTC, datetime
 from pathlib import Path
-from typing import Any, Literal
+from typing import Any, Literal, NamedTuple
 
 import h5py
 import hdf5plugin  # noqa: F401   side-effect import: registers HDF5_PLUGIN_PATH so h5py can load Blosc2 filters in fixtures
@@ -667,10 +667,15 @@ def cfg_surge_xt_global(
     )
 
 
-def _render_smoke_train_h5_subprocess(train_h5: Path, param_spec_name: str) -> None:
-    """Render the smoke ``train.h5`` via the ``generate_vst_dataset`` subprocess (real VST), failing loud on timeout/non-zero exit/missing output.
+def _render_smoke_train_subprocess(output_path: Path, param_spec_name: str) -> None:
+    """Render the smoke ``train`` shard via the ``generate_vst_dataset`` subprocess (real VST), failing loud on timeout/non-zero exit/missing output.
 
-    :param train_h5: Destination ``train.h5`` path; its parent must already exist.
+    The writer is dispatched on ``output_path``'s suffix by ``generate_vst_dataset``
+    (``.h5`` -> HDF5, ``.lance`` -> Lance), so this one renderer backs both the h5
+    and Lance real-VST smoke fixtures.
+
+    :param output_path: Destination shard path (``train.h5`` or ``train.lance``); its
+        parent must already exist.
     :param param_spec_name: Key into :data:`synth_setter.data.vst.param_specs` and
         :data:`synth_setter.data.vst.preset_paths` selecting spec and preset.
     """
@@ -681,7 +686,7 @@ def _render_smoke_train_h5_subprocess(train_h5: Path, param_spec_name: str) -> N
     generate_dataset_args += [
         sys.executable,
         "src/synth_setter/data/vst/generate_vst_dataset.py",
-        str(train_h5),
+        str(output_path),
         f"--plugin_path={PLUGIN_PATH}",
         f"--preset_path={preset_paths[param_spec_name]}",
         f"--param_spec_name={param_spec_name}",
@@ -722,7 +727,7 @@ def _render_smoke_train_h5_subprocess(train_h5: Path, param_spec_name: str) -> N
             f"(child stdout/stderr printed above; rerun with `pytest -s` if captured)",
             pytrace=False,
         )
-    assert train_h5.exists(), "Dataset generation failed to produce train.h5 fixture"
+    assert output_path.exists(), "Dataset generation failed to produce train fixture"
 
 
 def _smoke_fake_render_cfg(param_spec_name: str) -> RenderConfig:
@@ -803,12 +808,17 @@ def _build_surge_smoke_datasets(
     return smoke_dataset_dir
 
 
-def _build_surge_smoke_lance_datasets(tmp_path: Path, param_spec_name: str) -> Path:
+def _build_surge_smoke_lance_datasets(
+    tmp_path: Path,
+    param_spec_name: str,
+    render_train_lance: Callable[[Path, str], None],
+) -> Path:
     """Render the N-sample Surge smoke dataset natively as single-file Lance shards.
 
-    The Lance counterpart to :func:`_build_surge_smoke_datasets`: renders
-    ``train.lance`` through the production :func:`make_lance_dataset` writer, folds
-    its mel rows into ``stats.npz`` via :func:`stream_stats_lance`, and clones the
+    The Lance counterpart to :func:`_build_surge_smoke_datasets`: ``render_train_lance``
+    is the only difference between the real-VST and fake fixtures. It renders
+    ``train.lance`` through the production :func:`make_lance_dataset` writer, then this
+    folds the mel rows into ``stats.npz`` via :func:`stream_stats_lance` and clones the
     split into ``val``/``test``. No HDF5 file is produced — every shard carries the
     exact on-disk format the pipeline's Lance finalize emits.
 
@@ -816,6 +826,7 @@ def _build_surge_smoke_lance_datasets(tmp_path: Path, param_spec_name: str) -> P
         ``tmp_path / "data" / "smoke-lance"``.
     :param param_spec_name: Key into :data:`synth_setter.data.vst.param_specs` and
         :data:`synth_setter.data.vst.preset_paths` selecting spec and preset.
+    :param render_train_lance: Renders ``train.lance`` given ``(train_lance, param_spec_name)``.
 
     :return: Path to the directory holding ``{train,val,test}.lance`` and ``stats.npz``.
     """
@@ -825,7 +836,7 @@ def _build_surge_smoke_lance_datasets(tmp_path: Path, param_spec_name: str) -> P
     smoke_dataset_dir.mkdir(parents=True, exist_ok=True)
     train_lance = smoke_dataset_dir / "train.lance"
 
-    _render_smoke_train_lance_fake(train_lance, param_spec_name)
+    render_train_lance(train_lance, param_spec_name)
 
     # Sibling stats.npz folded straight from the Lance mel rows; mask degenerate
     # bins as the h5 path's --mask-degenerate-bins flag does for fake-plugin data.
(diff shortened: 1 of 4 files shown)


tests/test_train.py
```diff
@@ -35,10 +36,42 @@
...
+    if is_render:
+        for sample_idx in range(NUM_FIXTURE_SAMPLES):
+            sample_dir = output_dir / f"sample_{sample_idx}"
+            sample_dir.mkdir()
+            for name in ("target.wav", "pred.wav", "spec.png", "params.csv"):
+                (sample_dir / name).write_text("fake")
+    if is_metrics:
+        (output_dir / "aggregated_metrics.csv").write_text(_FAKE_AGGREGATED_METRICS_CSV)
+        (output_dir / "metrics.csv").write_text(_FAKE_METRICS_CSV)
+
```
khaledtin, line 64: **[code-health:warn]** Magic-number duplication / DRY: the constants `_PREDICT_VST_AUDIO_FRAGMENT`, `_COMPUTE_AUDIO_METRICS_FRAGMENT`, and `_FAKE_AGGREGATED_METRICS_CSV` (lines 39-43) duplicate identically-named symbols in `tests/test_eval.py` (lines 38-48), and `_fake_predict_postprocessing_subprocess` is a near-clone of `_fake_postprocessing_subprocess` / `_fake_run_with_metrics` / `_fake_run_with_permutation` in `test_eval.py`. The CSV body values even differ subtly between modules (`0.5/0.02` vs `0.5/0.1`), which is a silent-divergence trap — readers can't tell whether the difference is intentional. Hoist these into a shared helper (e.g. `tests/_eval_fakes.py`) imported by both modules.

tests/test_train.py
```diff
@@ -35,10 +36,42 @@
 # the parametrize lists on the two ``test_train_*_surge_xt`` tests cannot drift apart.
 _ORACLE_EXPERIMENT = "surge/fake_oracle"
 _SURGE_SMOKE_EXPERIMENTS = (_ORACLE_EXPERIMENT, "surge/ffn_full")
+_PREDICT_VST_AUDIO_FRAGMENT = "predict_vst_audio"
+_COMPUTE_AUDIO_METRICS_FRAGMENT = "compute_audio_metrics"
+_FAKE_AGGREGATED_METRICS_CSV = (
+    ",mean,std\nmss,0.5,0.02\nwmfcc,0.25,0.01\nsot,0.1,0.005\nrms,0.9,0.01\n"
+)
```
khaledtin: **[code-health:warn]** `_FAKE_AGGREGATED_METRICS_CSV` uses different numeric values (`mss,0.5,0.02` / `rms,0.9,0.01`) than `test_eval.py:43` (`mss,0.5,0.1` / `rms,0.9,0.01`). The Lance test asserts `mss_mean ≈ 0.5` and `rms_std ≈ 0.01`, both of which hold in either CSV; the `0.02 vs 0.1` divergence isn't load-bearing for the assertions. Either reuse the existing `test_eval._FAKE_AGGREGATED_METRICS_CSV` exactly, or comment why the new copy differs.

tests/test_train.py
```diff
@@ -35,10 +36,42 @@
...
 
 # TODO(#40): add @pytest.mark.ram gate for memory-intensive CPU tests test_train_fast_dev_run
 
 
+def _fake_predict_postprocessing_subprocess(args: list[str], **_kwargs: object) -> None:
+    """Materialize the render and metrics outputs the Lance smoke test asserts.
+
+    :param args: Subprocess argv from the eval postprocessing step.
+    :param **_kwargs: Ignored ``subprocess.run`` keyword arguments.
```
khaledtin: **[code-health:warn]** `_fake_predict_postprocessing_subprocess` reaches into argv with `args[args.index("-m") + 3]` to pull the output dir, then materializes 4 files × N samples plus 2 CSV files. The argv-shape knowledge (`-m`, then module, in_dir, out_dir at offset +3) is restated identically in 3+ places across `test_eval.py`; consolidating with the existing `_fake_postprocessing_subprocess` factory would let this test inherit the contract instead of forking it.

tests/test_train.py
```diff
@@ -471,3 +504,116 @@ def test_train_fast_dev_run_lance_datamodule(cfg_train_lance: DictConfig) -> Non
...
+    monkeypatch.setattr(
+        "synth_setter.cli.eval.as_file",
+        lambda _traversable: nullcontext(Path("/fake/headless-wrapper")),
+    )
+
+    HydraConfig().set_config(cfg_surge_xt_lance_eval)
+    metric_dict, object_dict = evaluate(cfg_surge_xt_lance_eval)
+
+    assert metric_dict["audio/mss_mean"] == pytest.approx(0.5)
+    assert metric_dict["audio/rms_std"] == pytest.approx(0.01)
```
khaledtin: **[code-health:warn]** `test_train_eval_surge_lance` runs ~70 lines and asserts on 4 distinct contracts: trained-step advancement, prediction tensor parity, audio/sample directory structure, and metrics-CSV finiteness. Per the cognitive-load principle, consider splitting at least the "predictions roundtrip" and "audio + metrics outputs" halves into two tests sharing the same train step. As-is the test is hard to bisect when one assertion fails.

tests/test_train.py
```diff
@@ -471,3 +504,116 @@ def test_train_fast_dev_run_lance_datamodule(cfg_train_lance: DictConfig) -> Non
...
+
+        if experiment_name == _ORACLE_EXPERIMENT:
+            target_params = torch.load(
+                predictions_dir / f"target-params-{sample_idx}.pt", weights_only=True
+            )
+            assert torch.equal(pred, target_params), (
+                f"oracle pred-{sample_idx}.pt != target-params-{sample_idx}.pt"
+            )
+
+    audio_dir = tmp_path / "audio"
```
khaledtin, line 610: **[code-health:warn]** The `expected_names` list comprehension `sorted(f"{prefix}-{sample_idx}.pt" for prefix in ("pred", "target-audio", "target-params") for sample_idx in range(NUM_FIXTURE_SAMPLES))` and the subsequent loop over `predictions_dir.iterdir()` re-derive the layout the production code writes. Pull the prefix tuple into a module-level constant (`_PREDICTION_PT_PREFIXES = ("pred", "target-audio", "target-params")`) so the assertion's expected shape is named, and reuse it in the iterate-and-load loop below to avoid the repeated string template.

tests/test_eval.py
```diff
@@ -266,6 +270,27 @@ def _compose_fake_oracle_eval_cfg(
     return cfg
 
 
+def _compose_parametrized_fake_oracle_eval_cfg(
```
khaledtin, line 274: **[code-health:warn]** The dataset parametrize ids (`"h5"` / `"lance"`) are good, but the parametrize tuple `_FAKE_ORACLE_DATASETS` and the per-test `(dataset_fixture, datamodule)` argspec carry an implicit invariant (lance ↔ "surge_lance", h5 ↔ None) that's never asserted. A small dataclass / NamedTuple with a single source of truth (e.g. `DatasetVariant(name, fixture, datamodule_group)`) would prevent the two columns from drifting out of sync in a future addition.

tests/conftest.py
```diff
@@ -531,17 +534,17 @@ def _build_surge_xt_smoke_cfg(
...
     """
     with initialize_config_module(version_base="1.3", config_module="synth_setter.configs"):
         cfg = compose(
             config_name="train.yaml",
             return_hydra_config=True,
-            overrides=[
-                f"experiment={experiment}",
-                "callbacks=[default_surge,eval_surge]",
-            ],
+            overrides=[f"experiment={experiment}", "callbacks=[default_surge,eval_surge]"]
```
khaledtin: **[code-health:warn]** `_build_surge_xt_smoke_cfg` now takes 4 params and grows an inline conditional override `+ ([f"datamodule={datamodule}"] if datamodule else [])`. The ≤3 args / no-flag-args principle is borderline; the `Optional[str]` parameter is effectively a flag selecting between two compose paths. Acceptable here, but a follow-up could express the choice as `datamodule_group: Literal["surge", "surge_lance"] = "surge"` to make the contract explicit and remove the `None` branch.

tests/conftest.py
```diff
@@ -921,6 +924,49 @@ def cfg_surge_xt(
...
+def cfg_surge_xt_lance_global(param_spec_name: str, experiment_name: str) -> DictConfig:
+    """Build a one-step Surge training config over the Lance datamodule.
+
+    :param param_spec_name: Param spec name for model width and callbacks.
+    :param experiment_name: Hydra ``experiment=...`` override.
+    :return: Resolved CPU ``datamodule=surge_lance`` smoke config.
+    """
+    return _build_surge_xt_smoke_cfg(
+        accelerator="cpu",
+        param_spec_name=param_spec_name,
```
khaledtin, line 1052: **[code-health:warn]** `cfg_surge_xt_lance_eval` duplicates `cfg_surge_xt_eval` almost verbatim, differing only in (a) the global cfg it copies from, (b) the dataset-root fixture, (c) `predict_file` suffix (`.lance` vs `.h5`), and (d) `rerender_target` (False vs True). DRY: extract the shared overrides into a helper or a single parametrized fixture indirected on the dataset format. The `rerender_target=False` divergence is also undocumented — add a comment naming the reason if it's load-bearing.

khaledtin (commented):
Re-post of repo-review-full WARN findings from PR #1721 as individual unresolved inline threads. The original orchestrator returned the code-health review inline instead of routing through `post_review.py`; this re-post anchors each WARN to its file:line so the threads stay visible and require explicit resolution. Source skill: code-health. No BLOCK findings.

tests/test_train.py
ktinubu: Fixed in a4dc3ac1 by moving the shared eval postprocessing fake, worker fragments, aggregated metrics CSV, and per-sample metrics CSV generation into `tests/helpers/eval_fakes.py`, then reusing that helper from both entrypoint test modules.

tests/test_eval.py
ktinubu: Fixed in a4dc3ac1 by replacing the parallel parametrization tuple with `_FakeOracleDataset`, so the fixture name and datamodule override travel as one structured variant.

tests/conftest.py
ktinubu: Fixed in a4dc3ac1 by making `_build_surge_xt_smoke_cfg` take an explicit `datamodule_group` with a default of `surge`, and always composing the datamodule override directly.

ktinubu:
Readiness note for a4dc3ac1: Copilot produced no post-push findings. I manually re-requested Copilot review at 2026-06-16T19:19:14Z via the reviewers API; after the follow-up wait, both `pulls/1721/comments` and `pulls/1721/reviews` had no Copilot entries newer than that request.

Issue #1742, opened 2026-06-16
Phase 4: Delete hand-rolled samplers, batch_size=None, and the HDF5 path


ktinubu:
# Phase 4: Delete hand-rolled samplers, `batch_size=None`, and the HDF5 path

> **Type**: Phase · **Parent Epic**: #1738 · **Domain**: code-health · **Milestone**: training v1.0.0
> **Design doc**: `lance-dataloader-migration.md` §3 (non-goals), §8, §9
> **Blocked by**: Phase 2 (and Phase 3 for the full HDF5 drop) · **Blocks**: —

## Description

With the map-style Lance path the default and proven (Phases 1–3), remove the
obsolete machinery that existed only to work around HDF5's slow random access.
This is the payoff phase: it deletes code and dependencies, done last when
nothing imports the old paths. Land it as one or a few reviewable cuts, each
keeping the suite green.

## Scope / tasks

- [ ] **Task 4.1 — Remove custom samplers + `batch_size=None`.** Delete
  `ShiftedBatchSampler`, `WithinChunkShuffledSampler`, `ShuffledSampler` and their
  tests (`surge_datamodule.py` L275–390; `tests/data/test_surge_datamodule.py`
  sampler tests). Remove the `batch_size=None` dataloader pattern and the
  `(start,stop)`/tuple branch in `_index_dataset`. The map dataloader from
  Phase 2 is the only training path.

- [ ] **Task 4.2 — Collapse the adapter indirection.** Remove the
  `ShardFile`/`ShardColumn` protocols and the `LanceShardFile`/`LanceColumn`
  h5py-shaped adapter (`lance_datamodule.py`), folding any still-needed logic
  (writable-copy guard, getpid reopen) into `LanceMapDataset`. Remove the
  batch-indexed `VSTDataset.__getitem__` and `_index_dataset` once unused.

- [ ] **Task 4.3 — Drop HDF5.** Remove the HDF5 dataset class / `_open` h5py
  branch, the `import h5py` / `hdf5plugin` side-effect import, the `.h5`
  `shard_suffix` defaults, and `h5py`/`hdf5plugin` from `pyproject.toml` +
  `uv.lock`. Update any configs/docs that reference `.h5`. (Gated on Phase 3 so
  cloud training no longer depends on the h5 path in any environment.)

- [ ] **Task 4.4 — Remove the config switch.** Delete the `loader`/`dataset_impl`
[...]

Issue #1739, opened 2026-06-16
Phase 1: Extract pure prepare_batch (no-op refactor, pinned by tests)


ktinubu:
# Phase 1: Extract pure `prepare_batch` (no-op refactor, pinned by tests)

> **Type**: Phase · **Parent Epic**: #1738 · **Domain**: training · **Milestone**: training v1.0.0
> **Design doc**: `lance-dataloader-migration.md` §4, §8, §9
> **Blocked by**: — · **Blocks**: Phase 2, Phase 4

## Description

Pull the per-batch logic out of `VSTDataset.__getitem__` into a standalone **pure
function** `prepare_batch`, and have the *current* `__getitem__` call it so
behavior is byte-identical. This is deliberately a no-op refactor: it touches the
only thing in the migration that affects batch *semantics*, isolates it, and pins
it with a test. Every later phase is then pure plumbing that cannot silently
change what a batch contains.

This phase ships alone. After it, both the h5 and Lance paths route through the
same function, and batch semantics are frozen for the rest of the epic.

## Scope / tasks

- [ ] **Task 1.1 — Extract `prepare_batch`.** New pure function (suggested home:
  `src/synth_setter/data/batch.py` or a module-level function in
  `surge_datamodule.py`):

  ```python
  def prepare_batch(
      raw: dict[str, np.ndarray | None],   # {"param_array", "mel_spec"?, "music2latent"?, "audio"?}
      *,
      mean: np.ndarray | None,
      std: np.ndarray | None,
      rescale_params: bool,
      ot: bool,
      generator: torch.Generator,
  ) -> dict[str, torch.Tensor | None]:      # {"mel_spec","m2l","params","noise","audio"}
      ...
  ```

  It must own exactly today's per-batch steps, in the same order
  (`surge_datamodule.py` L236–272): mel normalization `(mel - mean) / std`; param
  rescale `[0,1] → [-1,1]`; `noise = randn_like(params)`; OT/Hungarian match
  (`_hungarian_match(noise, params, mel, audio)`); `.contiguous()` on outputs;
  `None` passthrough for unread features.

- [ ] **Task 1.2 — Thread an explicit `generator`.** Replace bare
  `torch.randn_like(...)` (and any RNG inside the OT path) with calls that accept
  a `torch.Generator`, so randomness is reproducible and pinnable. The current
  `__getitem__` constructs/holds a generator and passes it in; default seeding
  must reproduce today's distribution (document the seed source).

[...]

Issue #1738, opened 2026-06-16
feat(training): Lance-native dataloader — drop the h5 shim, pre-stage from R2, clean DDP


ktinubu:
**Design doc**: `docs/design/lance-dataloader-migration.md`

## Description

The Lance read path is currently an HDF5-shaped compatibility shim. Nearly all of
its hand-rolled machinery (`batch_size=None`, batch-indexed `__getitem__`, the
`ShiftedBatchSampler` / `WithinChunkShuffledSampler` / `ShuffledSampler` family)
exists to dodge HDF5's slow random access — a constraint that does not apply to
Lance. This epic replaces the shim with a Lance-native, map-style dataloader,
makes R2 → local pre-staging the default for cloud training, and then deletes the
obsolete code (including HDF5).

Two problems this fixes, both raised during design review:

1. **Performance** — today each batch issues one Lance scan *per column* and pays
   an Arrow→NumPy→Torch conversion per column. We move to one projected `take`
   per batch via PyTorch's `__getitems__`, with per-batch math in a pure
   `collate_fn`.
2. **DDP correctness + simplicity** — today a custom sampler + `batch_size=None`
   relies on Lightning silently rewriting the sampler under `strategy: ddp`
   (untested; drops the sampler's per-epoch offset; silent-duplication risk if
   replacement ever fails to engage). Map-style + `shuffle=True` hits Lightning's
   clean, special-cased `DistributedSampler` path and we add an explicit
   disjoint-shard test.

Full research, the full option matrix, and the explicit rejection of
`SafeLanceDataset` (with the pinned-source evidence) live in the design doc
§4–§7. **This epic does not change model-facing batch semantics** — keys, shapes,
dtypes, normalization, rescale, noise, and OT matching stay byte-identical; that
invariant is the spine of the test plan.

## Why now

Training is moving to ~TB-scale data on Cloudflare R2. On object storage,
map-style `shuffle=True` is the *worst* access pattern (maximally scattered range
GETs), so the storage decision (pre-stage to node-local disk, exploiting R2's
zero egress) is coupled to the dataloader rewrite and worth doing as one epic.

## Phases

- [ ] **Phase 1** — Extract pure `prepare_batch` (no-op refactor, pinned by tests) · *blocks 2, 4*
- [ ] **Phase 2** — Map-style `LanceMapDataset` + `__getitems__` projected read + collate (behind a config switch) · *blocked by 1 · blocks 4*
[...]

PR #1744, opened 2026-06-16
docs(training): add Lance-native dataloader migration design doc

ktinubu:
## Why

Epics in this repo require a committed design doc, and the Lance-native dataloader
epic (#1738) plus its four phases (#1739–#1742) cite this document by section.
This lands the read/dataloader-side design so the phase issues have a stable,
grounded reference. It is the sibling to the storage/write-side
`docs/design/lance-dataset-api-migration.md`.

## What changed

- Adds `docs/design/lance-dataloader-migration.md` (docs only; no code).
- Captures the decision to move the Lance read path from the HDF5-shaped shim to
  a map-style `LanceMapDataset` with a single projected `__getitems__` `take`, a
  pure `collate_fn`, R2 → node-local pre-staging, and removal of the hand-rolled
  samplers / `batch_size=None` / HDF5 path.
- Records the rejected alternatives (notably `SafeLanceDataset`) with
  pinned-source evidence, the DDP-correctness analysis, and the four-phase plan
  with its dependency graph.

## Test plan

Docs-only change — no runtime behavior. Verified by:

- `pre-commit` (mdformat, codespell, trailing-whitespace, EOF) green on the file.
- `/repo-review-full-no-comments` run to convergence: **0 BLOCK, 0 WARN** across
  code-health, comment-hygiene, and lance-review. Every cited repo line number
  and Lance/pylance 7.0.0 API claim was verified against installed source and
  live upstream docs (`SafeLanceDataset.__getitems__`, `LanceDataset` =
  `IterableDataset`, `ShardedBatchSampler` signature + k=256 reservoir,
  `to_numpy_ndarray()` on `FixedShapeTensorArray`, `data_storage_version="2.2"`).

Refs #1738

### DIFF
--- a/docs/design/lance-dataloader-migration.md
+++ b/docs/design/lance-dataloader-migration.md
@@ -1,0 +1,369 @@
+# Design Doc: Lance-Native Dataloader Migration
+
+> **Status**: Draft
+> **Author**: KT (khaled@tinaudio.com)
+> **Last Updated**: 2026-06-16
+> **Tracking**: #1738 (feat(training): Lance-native dataloader)
+> **Related**: [lance-dataset-api-migration.md](lance-dataset-api-migration.md) (storage/write API; this doc is the *read*/dataloader sibling)
+
+______________________________________________________________________
+
+### Index
+
+| §   | Section                  | What it covers                                                            |
+| --- | ------------------------ | ------------------------------------------------------------------------- |
+| §1  | Context                  | Why the current Lance read path exists and what it costs                  |
+| §2  | Current architecture     | The h5-shaped shim, grounded in code                                      |
+| §3  | Goals / non-goals        | What we are and aren't changing                                           |
+| §4  | Decision                 | Map-style dataset + `__getitems__` projected read + pure collate          |
+| §5  | Tradeoffs considered     | map vs iterable; SafeLanceDataset; shuffle buffer; the full option matrix |
+| §6  | Network storage decision | Pre-stage R2 → local; why streaming loses for our shape                   |
+| §7  | DDP analysis             | DDP correctness risk in the current code and why the new design fixes it  |
+| §8  | Phase plan               | The four phases and their dependency graph                                |
+| §9  | Testing strategy         | Cross-cutting test philosophy; per-phase detail lives in the phase issues |
+| §10 | References               | All external + internal grounding links                                   |
+
+______________________________________________________________________
+
+## §1 Context
+
+The Lance read path was ported from the HDF5 design. Almost every hand-rolled
+piece in the current datamodule — the `batch_size=None` dataloader, the
+batch-indexed `__getitem__`, and the `ShiftedBatchSampler` /
+`WithinChunkShuffledSampler` / `ShuffledSampler` family — exists to work around a
+single HDF5 limitation: **slow random access**. The workarounds keep reads
+*contiguous* so h5py stays fast, then claw back approximate shuffling on top of
+that contiguity.
+
+Lance's headline property is fast random access (its own docs market "100x
+faster random access" vs Parquet — see §10). So the premise those workarounds
+were built on no longer holds. This migration deletes the reason they exist
+rather than porting them to Lance.
+
+Two goals drive the work: (1) **performance** — stop paying the h5-shaped
+per-column read tax and the row-dict conversion tax; (2) **DDP correctness +
+simplicity** — stop relying on Lightning silently rewriting a custom sampler,
+and get clean rank sharding for free.
+
+## §2 Current architecture (grounded)
+
+- `VSTDataset` is **batch-indexed**: one `__getitem__` returns a whole batch, and
+  the dataloader uses `batch_size=None`
+  (`src/synth_setter/data/surge_datamodule.py`, `__getitem__` ~L227–272,
+  `train_dataloader` ~L527–543).
+- Each batch reads **one column at a time**. `__getitem__` calls
+  `self.dataset_file["mel_spec"]`, then `["param_array"]`, then optionally
+  `["music2latent"]` / `["audio"]` — each a *separate* read
+  (`surge_datamodule.py` L236–256).
+- The Lance adapter (`src/synth_setter/data/lance_datamodule.py`) makes a Lance
+  column look like an `h5py.Dataset`. Contiguous slices go through
+  `dataset.scanner(columns=[name], offset=…, limit=…)`; stepped/fancy indices go
+  through `dataset.take(indices, columns=[name])` (`LanceColumn.__getitem__`
+  L48–74). So each logical column is a separate Lance scan.
+- `LanceShardFile.live_dataset()` already reopens the Lance dataset when
+  `os.getpid()` changes — i.e. **worker-safe lazy reopen on fork is already
+  implemented** (L108–122). This matters in §5: it is the one thing
+  `SafeLanceDataset` would otherwise give us, and we already have it.
+- Training uses `ShiftedBatchSampler`, which yields `(start, stop)` tuples
+  (L362–390), so the *training* path already hits the contiguous-scan fast path.
+- Shards are written in Lance format `data_storage_version="2.2"` — the
+  `LANCE_DATA_STORAGE_VERSION = "2.2"` module constant
+  (`src/synth_setter/pipeline/data/lance_shard.py` L21) passed as the
+  `data_storage_version` write kwarg (L111/L142) — which has good random
+  access + projection. Rows are large: mel ≈ 400 KB, audio ≈ 1.4 MB,
+  music2latent ≈ 21 KB (from fake-mode shapes, `surge_datamodule.py`
+  `_get_fake_item`).
+
+**The two costs.** (1) N separate Lance scans per batch instead of one projected
+read. (2) Arrow → NumPy → Torch conversion per column. Both are artifacts of the
+h5-shaped interface, not of Lance.
+
+## §3 Goals / non-goals
+
+**Goals**
+
+- One projected Lance read per batch (all needed columns in a single scan).
+- Map-style, sample-indexed dataset with standard `DataLoader(batch_size=B, shuffle=True)` — no custom sampler, no `batch_size=None`.
+- DDP via Lightning's default `DistributedSampler` injection, verified by test.
+- Pre-staging R2 → local disk as the default for cloud training.
+- Eventually drop HDF5 and the adapter indirection entirely.
+
+**Non-goals**
+
+- No new Lance features (versioning, compaction, vector indices) — that's the
+  storage-API epic ([lance-dataset-api-migration.md](lance-dataset-api-migration.md)).
+- No change to model-facing batch contents/semantics. The keys, shapes, dtypes,
+  normalization, rescale, noise, and OT matching must be byte-identical.
+- No backward compat with the legacy single-file `.lance` shim once cut over.
+
+## §4 Decision
+
+Move to a **map-style, sample-indexed** `Dataset` whose `__getitems__(indices)`
+issues **one projected `take`** for all required columns, and move per-batch
+math (mel norm, param rescale, noise, OT/Hungarian match) into a **pure
+`collate_fn`**.
+
+The key enabler is PyTorch's `__getitems__` (note the plural). PyTorch's
+map-style fetcher calls `__getitems__(list_of_indices)` when the dataset defines
+it, instead of calling `__getitem__` once per sample — so we get standard
+DataLoader batching/shuffling **and** a single batched Lance read, with no
+custom sampler (see §10: PyTorch data docs and `_MapDatasetFetcher`).
+
+End state:
+
+```python
+class LanceMapDataset(torch.utils.data.Dataset):
+    def __init__(self, uri, columns, storage_options=None):
+        self.uri, self.columns = uri, columns
+        self._storage_options = storage_options  # persisted so the fork-time reopen keeps R2 creds
+        # Cached once; assumes an immutable staged split (no concurrent
+        # compaction/append), like LanceShardFile.
+        self._len = lance.dataset(uri, storage_options=storage_options).count_rows()
+        self._ds = None                         # reopened after fork (getpid guard in _live)
+    def __len__(self): return self._len
+    def __getitems__(self, indices):            # single Lance take instead of N __getitem__ calls
+        ds = self._live()                       # reopen on fork (reuse existing pattern)
+        tbl = ds.take(indices, columns=self.columns)
+        # to_numpy_ndarray(), not to_numpy() — Lance returns FixedShapeTensorArray for tensor columns.
+        return {c: tbl.column(c).combine_chunks().to_numpy_ndarray() for c in self.columns}
+    def __getitem__(self, i):                   # rare fallback; DataLoader uses __getitems__ for batches
+        # __getitems__ returns a batch dict (leading dim = len(indices)); unwrap the single element.
+        return {c: v[0] for c, v in self.__getitems__([i]).items()}
+```
+
+The two index paths return different shapes — `__getitems__` a batch dict (rows
+stacked along a leading dim), `__getitem__` a single unwrapped sample — so a
+`collate_fn` must never re-stack a `__getitems__` result.
+
+`collate_fn` is the relocated, pure version of today's per-batch logic (norm,
+rescale, noise, OT). One contract to flag for implementers: because
+`__getitems__` already returns a whole batch (a dict of stacked arrays),
+PyTorch's `_MapDatasetFetcher` passes that dict straight to `collate_fn` — so
+`collate` receives a single batch dict, **not** a `list[dict]` of samples, and
+`prepare_batch` is written to accept it as-is (no per-sample aggregation, no
+double-stacking). That pass-through is a torch-internal detail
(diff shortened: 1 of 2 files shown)


ktinubu:
## Verification Results

Docs-only PR adding `docs/design/lance-dataloader-migration.md`. The `design/` tree is excluded from the mkdocs site (`mkdocs.yml` `exclude_docs`), so the verifiable promises are (a) the doc's internal links resolve and (b) its addition does not break the strict site build. Cross-reference accuracy (repo line numbers + pylance 7.0.0 API claims) was verified separately by `/repo-review-full-no-comments` to **0 BLOCK / 0 WARN** (sentinel `45433f8e`).

- [x] **Every relative link in the doc resolves to a real file** *(Level 2 — runs a real filesystem link-check; a failure means a dead in-repo link)*
  ```bash
  $ python3 - <<'PY'
  import re, pathlib
  doc = pathlib.Path("docs/design/lance-dataloader-migration.md")
  base = doc.parent; text = doc.read_text()
  bad=[]; checked=[]
  for href in re.findall(r'\[[^\]]+\]\(([^)]+)\)', text):
      if href.startswith(("http://","https://","#","mailto:")): continue
      target = href.split("#")[0]
      if not target: continue
      ok = (base/target).resolve().exists(); checked.append((href,ok))
      if not ok: bad.append(href)
  for h,ok in checked: print(f"  {'OK ' if ok else 'DEAD'}  {h}")
  print(f"checked={len(checked)} dead={len(bad)}"); assert not bad
  print("PASS")
  PY
    OK   lance-dataset-api-migration.md
    OK   lance-dataset-api-migration.md
    OK   lance-dataset-api-migration.md
  checked=3 dead=0
  PASS
  ```

- [x] **`mkdocs build --strict` succeeds with the doc present, and the doc stays excluded from the site** *(Level 1 — exercises the real site build end-to-end; strict mode fails on broken refs/nav)*
  ```bash
  $ uv run mkdocs build --strict 2>&1 | tail -4; echo "exit: ${PIPESTATUS[0]}"
  INFO    -  The following pages exist in the docs directory, but are not included in the "nav" configuration:
    - pr-readiness-loop.md
    - testing/mutmut.md
  INFO    -  Documentation built in 0.72 seconds
  exit: 0
  ```
  The new `design/lance-dataloader-migration.md` is **not** among the "not in nav" pages — confirming it is correctly excluded via `design/` and cannot silently break the strict build.

- [ ] **Cross-reference accuracy (repo L-numbers + Lance/pylance 7.0.0 API claims)** *(verified by `/repo-review-full-no-comments` round 5 → 0 BLOCK / 0 WARN against installed pylance source + live upstream docs; not re-run here as a standalone command)*

docs/design/lance-dataloader-migration.md
```diff
@@ -0,0 +1,369 @@
...
+piece in the current datamodule — the `batch_size=None` dataloader, the
+batch-indexed `__getitem__`, and the `ShiftedBatchSampler` /
+`WithinChunkShuffledSampler` / `ShuffledSampler` family — exists to work around a
+single HDF5 limitation: **slow random access**. The workarounds keep reads
+*contiguous* so h5py stays fast, then claw back approximate shuffling on top of
+that contiguity.
+
+Lance's headline property is fast random access (its own docs market "100x
+faster random access" vs Parquet — see §10). So the premise those workarounds
+were built on no longer holds. This migration deletes the reason they exist
```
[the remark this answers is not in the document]
ktinubu: Fixed in 13348acd: changed "its own docs market" → "its own docs claim" so the verb takes the quoted figure as its object.

docs/design/lance-dataloader-migration.md
```diff
@@ -0,0 +1,369 @@
...
+node** (`prepare_data_per_node=True` by default), which is exactly what we want
+for node-local staging — provided `dataset_root` is a node-local path, not a
+shared network mount. Phase 3 verifies this.
+
+## §7 DDP analysis (load-bearing)
+
+The DDP risk in the *current* code has nothing to do with Lance. The training
+dataloader is `DataLoader(train_dataset, batch_size=None, sampler=ShiftedBatchSampler(...))`, the DDP config uses `strategy: ddp` on 4
+devices (`configs/trainer/ddp.yaml`), and nothing sets `use_distributed_sampler`
+→ Lightning's default `True`.
```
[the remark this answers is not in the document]
ktinubu: Fixed in c3fd9df5: prefixed to `src/synth_setter/configs/trainer/ddp.yaml` (no repo-root `configs/` exists; now consistent with the other src-relative refs and doc-map.yaml).

docs/design/lance-dataloader-migration.md
```diff
@@ -0,0 +1,369 @@
...
+- Cloudflare R2 zero egress (pre-staging is free): https://www.cloudflare.com/developer-platform/products/r2/
+
+**Internal (repo, grounded)**
+
+- `src/synth_setter/data/surge_datamodule.py` — `VSTDataset.__getitem__` per-column reads (L236–256); samplers (L275–390); `train_dataloader`/`batch_size=None` (L527–543); `prepare_data` (L462–472).
+- `src/synth_setter/data/lance_datamodule.py` — `LanceColumn.__getitem__` (L48–74); `LanceShardFile.live_dataset` getpid reopen (L108–122).
+- `src/synth_setter/pipeline/r2_io.py` — `r2_storage_options` (L188); `download_dir_no_overwrite`.
+- `src/synth_setter/pipeline/data/lance_shard.py` — `LANCE_DATA_STORAGE_VERSION = "2.2"` constant (L21), applied as the `data_storage_version` write kwarg (L111/L142).
+- `docs/design/lance-dataset-api-migration.md` — storage/write-API sibling; "training dataloader keeps rclone local-first" decision.
+- `configs/trainer/ddp.yaml` — `strategy: ddp`, `devices: 4`; no `use_distributed_sampler` override.
```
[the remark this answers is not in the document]
ktinubu: Fixed in c3fd9df5: prefixed to `src/synth_setter/configs/trainer/ddp.yaml` (no repo-root `configs/` exists; now consistent with the other src-relative refs and doc-map.yaml).

docs/design/lance-dataloader-migration.md
```diff
@@ -0,0 +1,369 @@
...
+  `LANCE_DATA_STORAGE_VERSION = "2.2"` module constant
+  (`src/synth_setter/pipeline/data/lance_shard.py` L21) passed as the
+  `data_storage_version` write kwarg (L111/L142) — which has good random
+  access + projection. Rows are large: mel ≈ 400 KB, audio ≈ 1.4 MB,
+  music2latent ≈ 21 KB (from fake-mode shapes, `surge_datamodule.py`
+  `_get_fake_item`).
+
+**The two costs.** (1) N separate Lance scans per batch instead of one projected
+read. (2) Arrow → NumPy → Torch conversion per column. Both are artifacts of the
+h5-shaped interface, not of Lance.
```
khaledtin: **[synth-setter:block]** Incorrect grounding: the doc states shards are written with `data_storage_version="2.2"` and that `LANCE_DATA_STORAGE_VERSION = "2.2"`. The actual source constant is `LANCE_DATA_STORAGE_VERSION = "2.1"` (`src/synth_setter/pipeline/data/lance_shard.py:22`), and `"2.2"` appears nowhere in the repo — the source comment even notes `"2.1" equals the current default`. This is load-bearing: the whole §2 cost analysis (random-access + projection behavior) rests on the on-disk format version. Fix to `"2.1"` here and in the §10 References block.

docs/design/lance-dataloader-migration.md
```diff
@@ -0,0 +1,369 @@
...
+(`torch/utils/data/_utils/fetch.py`), so Phase 2 pins it with a test that fails
+loudly if the `uv.lock`-pinned torch ever changes the collation path. The
+dataloader becomes boring on purpose:
+`DataLoader(ds, batch_size=B, shuffle=True, num_workers=n, collate_fn=collate, pin_memory=True)`.
+
+## §5 Tradeoffs considered
+
+### 5.1 Map-style vs iterable
+
+| Axis               | Map-style (chosen)                                 | Iterable (`lance.torch.data.LanceDataset`)                       |
```
khaledtin, line 154: **[ml-pipeline:block]** The headline dataloader sketch sets `shuffle=True` with `num_workers=n`, but the design never specifies a `worker_init_fn` or how the per-rank/per-epoch seed is set. Under DDP, Lightning's `DistributedSampler.set_epoch(epoch)` MUST be called every epoch or every epoch replays the identical permutation — a silent correctness bug of exactly the class §7 claims to eliminate. `set_epoch` is mentioned nowhere; the design fixes the sampler-replacement footgun but leaves the epoch-seeding obligation unspecified. Pin it explicitly and assert it in the Phase 2/§9 DDP test.

docs/design/lance-dataloader-migration.md
```diff
@@ -0,0 +1,369 @@
...
+End state:
+
+```python
+class LanceMapDataset(torch.utils.data.Dataset):
+    def __init__(self, uri, columns, storage_options=None):
+        self.uri, self.columns = uri, columns
+        self._storage_options = storage_options  # persisted so the fork-time reopen keeps R2 creds
+        # Cached once; assumes an immutable staged split (no concurrent
+        # compaction/append), like LanceShardFile.
+        self._len = lance.dataset(uri, storage_options=storage_options).count_rows()
```
khaledtin, line 121: **[ml-pipeline:block]** The `LanceMapDataset` sketch is the design's contract object, but the only shape/dtype documentation for `__getitems__`/`__getitem__` lives in §4 prose, not as a method contract. Because the two index paths return different leading-dim shapes (the documented `collate_fn` foot-gun, lines 140–142), state the per-method shape/dtype contract on the methods themselves (e.g. `__getitems__` → `dict[str, ndarray]` of shape `(len(indices), *col_shape)`; `__getitem__` → single unwrapped sample) so the implementer cannot get it wrong.

khaledtin (changes requested):
Multi-skill review of PR #1744 — 5 parallel passes (code-health, synth-setter-project-standards, comment-hygiene, ml-data-pipeline, lance-review). Each BLOCK below is posted as an individual unresolved inline thread; WARNs are collapsed into the `## Advisory (WARN) findings` section to keep the inline-thread list short. Findings on files outside the diff are anchored to the line in the diff that causes the staleness or rolled into the review body.

## Advisory (WARN) findings

- **[code-health:warn]** docs/design/lance-dataloader-migration.md:123 — The `LanceMapDataset` "end state" snippet uses compound tuple assignment (`self.uri, self.columns = uri, columns`) and a statement on the `def __len__` line; since Phase 2 implementers will copy this block, model the repo's one-statement-per-line / separate-line `return` style.
- **[code-health:warn]** docs/design/lance-dataloader-migration.md:131 — The snippet calls `self._live()` but `_live` is never defined, yet the design's §5.2/§7 correctness argument rests on its fork/`getpid`-reopen behavior. Inline a 3-line `_live()` stub (getpid guard + reopen threading `self._storage_options`) or point at the `LanceShardFile.live_dataset()` pattern it mirrors.
- **[code-health:warn]** docs/design/lance-dataloader-migration.md:127 — `__init__` opens `lance.dataset(...)` only for `count_rows()` then discards it (`self._ds = None`); a half-sentence clarifying the count-only open is intentionally discarded and the live handle is created lazily in `_live` would remove a leak-vs-intent ambiguity.
- **[synth-setter:warn]** docs/design/lance-dataloader-migration.md:79 — Wrong line citation: the constant is at `lance_shard.py` L22, not L21 (L21 is the comment above it); the same mis-cite repeats in the References block. Write-kwarg lines are 112/143, not 111/142 (within `~` tolerance).
- **[synth-setter:warn]** docs/design/lance-dataloader-migration.md:372 — References-section restates the same `"2.2"`/L21 errors; correct to `"2.1"`, `L22`, `L112/L143` here too.
- **[synth-setter:warn]** docs/design/lance-dataloader-migration.md:124 — Credential note for implementers: persisting `storage_options` (plaintext R2 `access_key_id`/`secret_access_key` from `r2_storage_options()`) as a long-lived instance attribute replicated into every worker is only needed on the streaming fallback; for the default local path (§6) it should be `None`/absent. Add one sentence so Phase 2 doesn't wire creds in unconditionally (also: ml-pipeline).
- **[comment-hygiene:warn]** docs/design/lance-dataloader-migration.md:128 — Redundant inline comments: L128 `# reopened after fork (getpid guard in _live)` and L132 `# reopen on fork (reuse existing pattern)` state the same fact; keep L128 and tighten the call-site to add the local fact (e.g. `# getpid-guarded handle`) (also: code-health).
- **[ml-pipeline:warn]** docs/design/lance-dataloader-migration.md:122 — `columns`/`storage_options` enter via constructor args but the doc never states they originate from Hydra/YAML config (config-externalized); say the projection list and R2 options come from the datamodule config so an implementer doesn't inline a literal column list.
- **[ml-pipeline:warn]** docs/design/lance-dataloader-migration.md:148 — The design leans on a torch-internal (`_MapDatasetFetcher` passing the `__getitems__` dict straight to `collate_fn`) and pins it with a test; state the failure mode (shape corruption / double-stacking vs. exception) and assertion target (collate receives a dict, not `list[dict]`) so the pinning test actually catches a regression.
- **[ml-pipeline:warn]** docs/design/lance-dataloader-migration.md:336 — The read-equivalence test asserts value equality but never binds "projection minimality" to an observable (only `self.columns` scanned). Since §5.2 makes no-over-read the core reason to reject `SafeLanceDataset`, add a test that fails when an unprojected `take` sneaks in, else the central perf claim stays unverified.
- **[ml-pipeline:warn]** docs/design/lance-dataloader-migration.md:134 — `combine_chunks().to_numpy_ndarray()` is asserted for `FixedShapeTensorArray` columns but no contract/fallback is given for non-fixed-shape columns (variable-length `audio`, scalar `param_array`); state which column types the path assumes or the implementer hits a conversion error on the first non-tensor column.
- **[lance:warn]** docs/design/lance-dataloader-migration.md:357 — §10 cites the Lance repo as `github.com/lance-format/lance` (resolves, currently-canonical after a recent rename), but most cross-references still use `lancedb/lance`; `main`-branch deep links to `data.py`/`sampler.py` will drift — pin to a commit-SHA permalink.
- **[lance:warn]** docs/design/lance-dataloader-migration.md:356 — Could not independently confirm the `cache=True` kwarg name (line 248) against a live signature within the lookup budget; `IterableDataset`, `SafeLanceDataset`, and `batch_readahead=8` all verified verbatim. Quick grounding check on the exact `cache` kwarg before merge.

docs/design/lance-dataloader-migration.md
```diff
@@ -0,0 +1,369 @@
...
+Under that default, Lightning **rewrites the dataloader's sampler** for DDP. Its
+behavior is version-dependent and is exactly the thing we should not rely on
+silently:
+
+- Lightning special-cases plain `SequentialSampler` / `RandomSampler` and
+  cleanly swaps in a `DistributedSampler`; for *custom* samplers it instead
+  wraps/replaces with a warning ("Replacing custom sampler with distributed
+  version …"). See §10 (Lightning trainer docs + issue #5271).
+- With a custom `ShiftedBatchSampler` + `batch_size=None`, the most likely
+  outcome is correct *sharding* (each integer index → one contiguous batch via
```
khaledtin, line 266: Consistency nit in the load-bearing §7. This says the likely DDP outcome is "correct *sharding* (each **integer index** → one contiguous batch via `__getitem__`)", but §2 (and the actual code) establish that `ShiftedBatchSampler` is a `BatchSampler` subclass whose `__iter__` yields `(start, stop)` **tuples** — `__getitem__` receives a tuple and slices, there is no integer index on the training path. Since this section's whole point is the precise sampler-rewrite mechanism, the imprecise "integer index" framing undercuts it. Suggest: "each `(start, stop)` index → one contiguous batch via `__getitem__`". Verified: `surge_datamodule.py:362` (`class ShiftedBatchSampler(torch.utils.data.BatchSampler)`) and `:388` (`yield (i * self.batch_size + offset, ...)`).

docs/design/lance-dataloader-migration.md
```diff
@@ -0,0 +1,369 @@
+# Design Doc: Lance-Native Dataloader Migration
+
+> **Status**: Draft
```
khaledtin, line 3: Status mismatch (doc-lifecycle, holistic). The header marks this **Draft**, but the doc is being committed specifically as the *stable grounding reference* that four already-filed phase issues (#1739–#1742, all OPEN and citing this doc by section) point at, and §4 records a finalized **Decision** with rejected alternatives pinned. A reader arriving from a phase issue sees "Draft" and reasonably distrusts the decision as still-in-flux. Recommend promoting to `Accepted`/`Approved` on merge (or note explicitly that the decision is accepted while wording is draft) so the committed status matches its load-bearing role.

khaledtin (commented):
Personal overall doc review (holistic pass, separate from the multi-skill review). Two additional findings; both verified against the tree. Grounding is otherwise strong — sampler names (`ShiftedBatchSampler`/`WithinChunkShuffledSampler`/`ShuffledSampler`), the `(start, stop)` tuple yield, and all five cited issues (#1738–#1742, matching titles) check out.

ktinubu:
Closing after stale-PR triage: superseded by the as-built Lance loader implementation in #1783 and #1834, followed by legacy-loader removal in #2075.

Issue #1740, opened 2026-06-16
Phase 2: Map-style LanceMapDataset + __getitems__ projected read + pure collate


ktinubu:
# Phase 2: Map-style `LanceMapDataset` + `__getitems__` projected read + pure collate

> **Type**: Phase · **Parent Epic**: #1738 · **Domain**: training · **Milestone**: training v1.0.0
> **Design doc**: `lance-dataloader-migration.md` §4, §5, §7, §9
> **Blocked by**: Phase 1 · **Blocks**: Phase 4

## Description

Introduce a **new** sample-indexed, map-style Lance dataset whose
`__getitems__(indices)` issues **one projected `take`** for all required
columns, wire it to a standard `DataLoader` whose `collate_fn` is the pure
`prepare_batch` from Phase 1, and gate cutover behind a config switch so the old
path still runs for comparison. This single phase captures both wins at once —
the projected read **and** the map-style/DDP fix — without building a throwaway
batch-indexed intermediate.

Do **not** mutate the existing `LanceVSTDataset`; add `LanceMapDataset`
alongside it so both can run and be compared by the read-equivalence test. Flip
the default only once the test matrix is green.

## Scope / tasks

- [ ] **Task 2.1 — `LanceMapDataset` (map-style, sample-indexed).** New class
  (`src/synth_setter/data/lance_datamodule.py` or a new `lance_map.py`):
  - `__len__` returns **row count** (not batch count).
  - `__getitems__(indices: list[int]) -> dict[str, np.ndarray]` does
    `ds.take(indices, columns=self.columns)` — a single projected read — then
    `column(c).combine_chunks().to_numpy_ndarray()` per column, copying out of
    read-only Arrow buffers (reuse the writable-copy guard from
    `LanceColumn.__getitem__`).
  - `__getitem__(i)` delegates to `__getitems__([i])` (rare path; PyTorch uses
    `__getitems__` for batched fetch — see design §10).
  - **Worker-safe lazy open**: open `lance.dataset(uri)` lazily on first read,
    reopen when `os.getpid()` changes. Reuse the exact pattern already in
    `LanceShardFile.live_dataset()` (L108–122) — do not import
    `SafeLanceDataset`.
  - Column projection is driven by the `read_mel` / `read_m2l` / `read_audio`
    flags: `columns = ["param_array"] (+ "mel_spec") (+ "music2latent") (+ "audio")`.
  - Accept optional `storage_options` (for tests that point at R2; production
[...]

PR #1745, opened 2026-06-17
refactor(training): extract pure prepare_batch from VSTDataset

### LINKED ISSUE #1739: Phase 1: Extract pure prepare_batch (no-op refactor, pinned by tests)

ktinubu:
## Why

Phase 1 of the Lance-native dataloader epic (#1738). Before any storage/sampling
changes, isolate the one thing in the migration that affects batch *semantics* —
the per-batch math — into a pure function and pin it with goldens, so every later
phase is plumbing that provably cannot change what a batch contains.

## What changed

- Extracts the per-batch math (mel normalization, param rescale `[0,1]→[-1,1]`,
  seeded noise, optional Hungarian/OT matching, `.contiguous()`, `None`
  passthrough) out of `VSTDataset.__getitem__` into a pure module-level
  `prepare_batch(raw, *, mean, std, rescale_params, ot, generator)`.
- Adds a `RawBatch` TypedDict for the read-column input contract.
- Threads an explicit `torch.Generator` through the noise draw so the batch is
  pinnable; `__getitem__` now assembles the raw column dict and delegates.
- **Incidental bug fix:** `__len__` counted rows off `dataset_file["audio"]`,
  which is optional on Lance shards and would `KeyError` before any read; it now
  counts off the always-present `param_array`.
- Samplers, `batch_size=None`, fake mode, and `repeat_first_batch` are untouched.

## Reviewer note — RNG seeding surface

The noise draw moves from a global `torch.randn_like` (governed by Lightning
`seed_everything`) to a per-dataset, entropy-seeded `torch.Generator`. The golden
tests pin the per-batch math bit-for-bit under matched seeds, so batch
*semantics* are unchanged — but the production *default* seeding surface changes
(noise is no longer governed by the global seed, and forked workers currently
share the generator and draw correlated noise). Per-worker re-seeding via
`worker_init_fn` is explicitly Phase 2 scope (#1740, Task 2.2), where the
seeding/DDP surface is finalized. Flagging here for reviewer awareness.

## Test plan

`tests/data/test_prepare_batch.py` (new) + new tests in
`tests/data/test_surge_datamodule.py`. 90 passed (serial, `-m "not slow"`):

- **Pinning/golden** — `prepare_batch` output matches a frozen reference of the
  pre-refactor body bit-for-bit under a fixed `torch.Generator` seed; outputs are
  `float32` + `.contiguous()`.
- **Purity** — same seed → identical output; `raw` inputs are not mutated.
- **Feature flags / rescale / OT** — modality slots match `read_*` flags; rescale
  toggle maps `[0,1]→[-1,1]`; `ot=True` pairing matches `_hungarian_match` and a
  reference-free brute-force OT-optimality check; `ot=False` passes through.
- **End-to-end no-op** — `test_getitem_unchanged_after_extraction` compares the
  post-refactor `__getitem__` to a pre-refactor golden, proving the rewire is a
  true no-op.

Closes #1739
Part of #1738

### DIFF
--- a/src/synth_setter/data/surge_datamodule.py
+++ b/src/synth_setter/data/surge_datamodule.py
@@ -1,7 +1,7 @@
 import random
 from collections.abc import Iterator, Sequence
 from pathlib import Path
-from typing import ClassVar, Literal, Protocol
+from typing import ClassVar, Literal, NotRequired, Protocol, TypedDict
 
 import h5py
 import hdf5plugin  # noqa: F401  # side-effect import: registers HDF5 blosc filters for shard I/O
@@ -17,6 +17,89 @@ from synth_setter.pipeline import r2_io
 # Registry key whose spec width sizes fake-mode batches and seeds the
 # datamodule default when no ``param_spec_name`` is configured.
 _DEFAULT_PARAM_SPEC_NAME = "surge_xt"
+
+
+# DOC601/DOC603: pydoclint can't read sphinx ``:ivar:`` docs, so TypedDict keys
+# are documented in the docstring body instead.
+class RawBatch(TypedDict):  # noqa: DOC601, DOC603
+    """One batch of read shard columns, the input to :func:`prepare_batch`.
+
+    Shapes are ``(batch, ...)``: ``param_array`` is ``(batch, num_params)`` and
+    always present; ``mel_spec`` is ``(batch, channels, n_mels, n_frames)``,
+    ``music2latent`` is ``(batch, latent_dim, n_frames)``, ``audio`` is
+    ``(batch, channels, samples)``. Each optional key is ``None`` when its read
+    flag is off.
+    """
+
+    param_array: np.ndarray
+    mel_spec: NotRequired[np.ndarray | None]
+    music2latent: NotRequired[np.ndarray | None]
+    audio: NotRequired[np.ndarray | None]
+
+
+def prepare_batch(
+    raw: RawBatch,
+    *,
+    mean: np.ndarray | None,
+    std: np.ndarray | None,
+    rescale_params: bool,
+    ot: bool,
+    generator: torch.Generator,
+) -> dict[str, torch.Tensor | None]:
+    """Turn one batch of read columns into model-ready tensors.
+
+    The single owner of per-batch semantics: mel normalization, param rescale,
+    seeded noise, and optional Hungarian matching, in that order. Pure — it
+    reads only ``raw`` and the passed args and never writes back into them, so a
+    fixed ``generator`` seed makes the whole batch reproducible. The h5 and Lance
+    read paths both route here so batch contents stay format-independent.
+
+    :param raw: Read shard columns; see :class:`RawBatch` for keys and shapes.
+        ``param_array`` ``(batch, num_params)`` is required; a missing or
+        ``None`` optional modality drops to ``None`` in the output.
+    :param mean: Mel mean to subtract, broadcasting over the batch axis, or
+        ``None`` to skip normalization.
+    :param std: Mel std to divide by, broadcasting over the batch axis, or
+        ``None`` to skip normalization.
+    :param rescale_params: Whether to map params from ``[0, 1]`` to ``[-1, 1]``.
+    :param ot: Whether to Hungarian-match noise to params (and permute the
+        conditioning rows to follow).
+    :param generator: RNG for the noise draw; seed it to pin the batch.
+    :returns: ``{"mel_spec", "m2l", "params", "noise", "audio"}`` with
+        ``float32`` contiguous tensors, ``None`` for unread modalities; ``params``
+        and ``noise`` are ``(batch, num_params)``.
+    """
+    audio_raw = raw.get("audio")
+    audio = torch.from_numpy(audio_raw).to(dtype=torch.float32) if audio_raw is not None else None
+
+    mel_raw = raw.get("mel_spec")
+    if mel_raw is not None:
+        if mean is not None and std is not None:
+            mel_raw = (mel_raw - mean) / std
+        mel_spec = torch.from_numpy(mel_raw).to(dtype=torch.float32)
+    else:
+        mel_spec = None
+
+    m2l_raw = raw.get("music2latent")
+    m2l = torch.from_numpy(m2l_raw).to(dtype=torch.float32) if m2l_raw is not None else None
+
+    param_array = raw["param_array"]
+    if rescale_params:
+        param_array = param_array * 2 - 1
+    params = torch.from_numpy(param_array).to(dtype=torch.float32)
+    # empty_like(params) matches params' device/dtype/shape and normal_ accepts
+    # the generator (randn_like does not), so noise can't land on a foreign device.
+    noise = torch.empty_like(params).normal_(generator=generator)
+    if ot:
+        noise, params, mel_spec, audio = _hungarian_match(noise, params, mel_spec, audio)
+
+    return dict(
+        mel_spec=mel_spec.contiguous() if mel_spec is not None else None,
+        m2l=m2l.contiguous() if m2l is not None else None,
+        params=params.contiguous(),
+        noise=noise.contiguous(),
+        audio=audio.contiguous() if audio is not None else None,
+    )
 
 
 class ShardColumn(Protocol):
@@ -59,10 +142,8 @@ class ShardFile(Protocol):
         ...
 
 
-# DOC601/DOC603: pydoclint cannot see sphinx ``:ivar:`` class-attribute docs (the
-# underlying docstring_parser drops them from its attribute list), so a documented
-# class with class-level annotations is unavoidably flagged; mean/std/dataset_file are
-# described in the class docstring body instead.
+# DOC601/DOC603: pydoclint can't read sphinx ``:ivar:`` docs, so class-level
+# annotations are documented in the docstring body instead.
 class VSTDataset(torch.utils.data.Dataset):  # noqa: DOC601, DOC603
     """Batch-indexed dataset over one shard of VST renders (HDF5 by default).
 
@@ -71,7 +152,9 @@ class VSTDataset(torch.utils.data.Dataset):  # noqa: DOC601, DOC603
     ``ot`` is set). Fake mode synthesises
     random tensors of the configured width without opening any file. ``mean`` and ``std``
     hold the optional saved mel statistics applied during reads (``None`` when unloaded);
-    ``dataset_file`` is the open :class:`ShardFile` handle (``None`` in fake mode).
+    ``dataset_file`` is the open :class:`ShardFile` handle (``None`` in fake mode);
+    ``generator`` is the entropy-seeded RNG threaded into :func:`prepare_batch` for the
+    per-batch noise draw (seed it to pin a batch).
     Storage-format subclasses override ``_open`` to read non-HDF5 shards.
     """
 
@@ -111,6 +194,13 @@ class VSTDataset(torch.utils.data.Dataset):  # noqa: DOC601, DOC603
         self.batch_size = batch_size
         self.ot = ot
 
+        # A bare torch.Generator() has a fixed default seed, so .seed() is needed
+        # to actually randomize production noise across runs; tests call
+        # generator.manual_seed(...) to pin a batch. num_workers>0 forks share
+        # this state and draw correlated noise; per-worker re-seed is Phase 2.
+        self.generator = torch.Generator()
+        self.generator.seed()
+
         self.read_audio = read_audio
         self.read_mel = read_mel
         self.read_m2l = read_m2l
@@ -147,8 +237,6 @@ class VSTDataset(torch.utils.data.Dataset):  # noqa: DOC601, DOC603
         :param dataset_file: Shard path used to locate the sibling ``stats.npz``.
         :raises FileNotFoundError: If the expected ``stats.npz`` is missing.
         """
-        # for /path/to/train.h5 we would expect to find /path/to/stats.npz
-        # if not, we throw an error
         stats_file = VSTDataset.get_stats_file_path(dataset_file)
         if not stats_file.exists():
             raise FileNotFoundError(
@@ -179,10 +267,15 @@ class VSTDataset(torch.utils.data.Dataset):  # noqa: DOC601, DOC603
         if self.fake:
             return 10000
 
-        return self.dataset_file["audio"].shape[0] // self.batch_size
+        # param_array is the one always-present column; audio is optional on
+        # Lance shards, so counting rows off it would KeyError before any read.
+        return self.dataset_file["param_array"].shape[0] // self.batch_size
 
     def _get_fake_item(self) -> dict[str, torch.Tensor | None]:
         """Synthesise one random batch matching the configured feature flags.
+
+        Bypasses :func:`prepare_batch`: fake batches are already tensors and need
+        no normalization or OT, and their noise is intentionally unseeded.
 
         :returns: Mapping of feature names to random tensors (or ``None`` when unread).
         """
@@ -233,43 +326,36 @@ class VSTDataset(torch.utils.data.Dataset):  # noqa: DOC601, DOC603
         if self.fake:
             return self._get_fake_item()
 
-        if self.read_audio:
-            audio = self._index_dataset(self.dataset_file["audio"], idx)
-            audio = torch.from_numpy(audio).to(dtype=torch.float32)
-        else:
-            audio = None
-
-        if self.read_mel:
-            mel_spec = self._index_dataset(self.dataset_file["mel_spec"], idx)
-            if self.mean is not None and self.std is not None:
-                mel_spec = (mel_spec - self.mean) / self.std
-            mel_spec = torch.from_numpy(mel_spec).to(dtype=torch.float32)
-        else:
-            mel_spec = None
-
-        if self.read_m2l:
-            m2l = self._index_dataset(self.dataset_file["music2latent"], idx)
-            m2l = torch.from_numpy(m2l).to(dtype=torch.float32)
-        else:
-            m2l = None
-
-        param_array = self._index_dataset(self.dataset_file["param_array"], idx)
-        if self.rescale_params:
-            param_array = param_array * 2 - 1
-        param_array = torch.from_numpy(param_array).to(dtype=torch.float32)
-        noise = torch.randn_like(param_array)
-        if self.ot:
-            noise, param_array, mel_spec, audio = _hungarian_match(
-                noise, param_array, mel_spec, audio
-            )
-
-        return dict(
-            mel_spec=mel_spec.contiguous() if mel_spec is not None else None,
-            m2l=m2l.contiguous() if m2l is not None else None,
-            params=param_array.contiguous(),
-            noise=noise.contiguous(),
-            audio=audio.contiguous() if audio is not None else None,
+        # Every key is always present; an off read flag stores ``None`` rather
+        # than omitting the key, so prepare_batch never relies on key-absence.
+        raw: RawBatch = {
+            "param_array": self._index_dataset(self.dataset_file["param_array"], idx),
+            "audio": self._read_column("audio", idx, self.read_audio),
+            "mel_spec": self._read_column("mel_spec", idx, self.read_mel),
+            "music2latent": self._read_column("music2latent", idx, self.read_m2l),
+        }
+        return prepare_batch(
+            raw,
+            mean=self.mean,
+            std=self.std,
+            rescale_params=self.rescale_params,
+            ot=self.ot,
+            generator=self.generator,
         )
+
+    def _read_column(
+        self, name: str, idx: int | Sequence[int] | np.ndarray, read: bool
+    ) -> np.ndarray | None:
+        """Slice one optional shard column, or skip it when its read flag is off.
+
+        :param name: Column name within the shard.
+        :param idx: Batch index, or a ``(start, stop)`` pair, or a sequence of rows.
+        :param read: Whether this column's modality is enabled.
+        :returns: The selected rows, or ``None`` when ``read`` is false.
+        """
+        if not read:
+            return None
+        return self._index_dataset(self.dataset_file[name], idx)
 
 
 class WithinChunkShuffledSampler(torch.utils.data.Sampler):
@@ -319,7 +405,6 @@ class WithinChunkShuffledSampler(torch.utils.data.Sampler):
 
         indices = np.concatenate(indices, axis=0)
 
-        # shuffle by rows
         np.random.shuffle(indices)
 
         for row in indices:
@@ -534,12 +619,7 @@ class VSTDataModule(LightningDataModule):
             batch_size=None,
             num_workers=self.num_workers,
             pin_memory=self.pin_memory,
-            # sampler=WithinChunkShuffledSampler(
-            #     self.batch_size, len(self.train_dataset), 4
-            # ),
             sampler=ShiftedBatchSampler(self.batch_size, len(self.train_dataset)),
-            # sampler=ShuffledSampler(self.batch_size, len(self.train_dataset)),
-            # shuffle=True,
         )
 
     def val_dataloader(self) -> DataLoader:
--- a/tests/data/test_prepare_batch.py
+++ b/tests/data/test_prepare_batch.py
@@ -1,0 +1,415 @@
+"""Pinning tests for :func:`synth_setter.data.surge_datamodule.prepare_batch`.
+
+Each test compares against :func:`_reference_prepare_batch`, an independent
+golden that applies the batch math (mel norm, param rescale, seeded noise,
+Hungarian match) drawing noise from the global RNG via ``torch.manual_seed``. A
+seeded ``torch.Generator`` reproduces the seeded global RNG bit-for-bit on CPU
+(true for current PyTorch; the equivalence is the comparison's only cross-version
+assumption), so the golden and ``prepare_batch`` are directly comparable.
+"""
+
+from __future__ import annotations
+
+import itertools
+from typing import cast
+
+import numpy as np
+import pytest
+import torch
+
+from synth_setter.data.ot import _hungarian_match
+from synth_setter.data.surge_datamodule import RawBatch, prepare_batch
+
+_NUM_PARAMS = 5
+_BATCH = 8
+_MEL_SHAPE = (_BATCH, 2, 4, 6)
+_AUDIO_SHAPE = (_BATCH, 2, 16)
+_M2L_SHAPE = (_BATCH, 3, 7)
+
+
+def _unwrap(maybe_tensor: torch.Tensor | None) -> torch.Tensor:
+    """Assert ``maybe_tensor`` is non-``None`` and narrow it for pyright.
+
+    :param maybe_tensor: The dict value to narrow.
+    :returns: The same tensor, typed as non-``None``.
+    """
+    assert maybe_tensor is not None
+    return maybe_tensor
+
+
+def _unwrap_array(maybe_array: np.ndarray | None) -> np.ndarray:
+    """Assert ``maybe_array`` is non-``None`` and narrow it for pyright.
+
+    :param maybe_array: The ``raw`` dict value to narrow.
+    :returns: The same array, typed as non-``None``.
+    """
+    assert maybe_array is not None
+    return maybe_array
+
+
+def _reference_prepare_batch(
+    raw: RawBatch,
+    *,
+    mean: np.ndarray | None,
+    std: np.ndarray | None,
+    rescale_params: bool,
+    ot: bool,
+    seed: int,
+) -> dict[str, torch.Tensor | None]:
+    """Apply the batch math independently of ``prepare_batch`` as the golden.
+
+    Draws noise from the global RNG seeded by ``seed``; a seeded
+    ``torch.Generator`` in ``prepare_batch`` reproduces this bit-for-bit on CPU.
+
+    :param raw: Read shard columns; see :class:`RawBatch`.
+    :param mean: Mel mean, or ``None`` to skip normalization.
+    :param std: Mel std, or ``None`` to skip normalization.
+    :param rescale_params: Whether to map params ``[0, 1] -> [-1, 1]``.
+    :param ot: Whether to Hungarian-match noise to params.
+    :param seed: Global-RNG seed for the noise draw.
+    :returns: ``{"mel_spec", "m2l", "params", "noise", "audio"}`` tensors.
+    """
+    audio_raw = raw.get("audio")
+    if audio_raw is not None:
+        audio = torch.from_numpy(audio_raw).to(dtype=torch.float32)
+    else:
+        audio = None
+
+    mel_raw = raw.get("mel_spec")
+    if mel_raw is not None:
+        mel_spec = mel_raw
+        if mean is not None and std is not None:
+            mel_spec = (mel_spec - mean) / std
+        mel_spec = torch.from_numpy(mel_spec).to(dtype=torch.float32)
+    else:
+        mel_spec = None
+
+    m2l_raw = raw.get("music2latent")
+    if m2l_raw is not None:
+        m2l = torch.from_numpy(m2l_raw).to(dtype=torch.float32)
+    else:
+        m2l = None
+
+    param_raw = _unwrap_array(raw["param_array"])
+    if rescale_params:
+        param_raw = param_raw * 2 - 1
+    param_array = torch.from_numpy(param_raw).to(dtype=torch.float32)
+    with torch.random.fork_rng():
+        torch.manual_seed(seed)
+        noise = torch.randn_like(param_array)
+    if ot:
+        noise, param_array, mel_spec, audio = _hungarian_match(noise, param_array, mel_spec, audio)
+
+    return dict(
+        mel_spec=mel_spec.contiguous() if mel_spec is not None else None,
+        m2l=m2l.contiguous() if m2l is not None else None,
+        params=param_array.contiguous(),
+        noise=noise.contiguous(),
+        audio=audio.contiguous() if audio is not None else None,
+    )
+
+
+def _make_raw(
+    *,
+    read_mel: bool = True,
+    read_m2l: bool = False,
+    read_audio: bool = False,
+    seed: int = 7,
+) -> RawBatch:
+    """Build a deterministic ``raw`` batch for the given read flags.
+
+    :param read_mel: Whether to include a ``mel_spec`` array.
+    :param read_m2l: Whether to include a ``music2latent`` array.
+    :param read_audio: Whether to include an ``audio`` array.
+    :param seed: Seed for the NumPy generator backing every array.
+    :returns: ``raw`` batch with ``param_array`` always present.
+    """
+    rng = np.random.default_rng(seed)
+    raw: RawBatch = {
+        "param_array": rng.random((_BATCH, _NUM_PARAMS)).astype(np.float32),
+    }
+    if read_mel:
+        raw["mel_spec"] = rng.standard_normal(_MEL_SHAPE).astype(np.float32)
+    if read_m2l:
+        raw["music2latent"] = rng.standard_normal(_M2L_SHAPE).astype(np.float32)
+    if read_audio:
+        raw["audio"] = rng.standard_normal(_AUDIO_SHAPE).astype(np.float32)
+    return raw
+
+
+def test_prepare_batch_is_pure_and_pinned() -> None:
+    """``prepare_batch`` reproduces the frozen pre-refactor golden bit-for-bit."""
+    seed = 0
+    raw = _make_raw(read_mel=True, read_audio=True, seed=3)
+    mean = np.zeros((2, 4, 6), dtype=np.float32)
+    std = np.full((2, 4, 6), 2.0, dtype=np.float32)
+
+    golden = _reference_prepare_batch(
+        raw, mean=mean, std=std, rescale_params=True, ot=True, seed=seed
+    )
+    out = prepare_batch(
+        raw,
+        mean=mean,
+        std=std,
+        rescale_params=True,
+        ot=True,
+        generator=torch.Generator().manual_seed(seed),
+    )
+
+    for key in ("mel_spec", "params", "noise", "audio"):
+        # atol=rtol=0: every step (affine mel norm, x*2-1, seeded noise, integer
+        # row permutation from Hungarian) is exact, so bit-equality must hold.
+        torch.testing.assert_close(out[key], golden[key], atol=0.0, rtol=0.0)
+        assert _unwrap(out[key]).is_contiguous()
+        assert _unwrap(out[key]).dtype == torch.float32
+    assert out["m2l"] is None
+
+
+@pytest.mark.parametrize("ot", [False, True])
+def test_prepare_batch_same_seed_yields_identical_output(ot: bool) -> None:
+    """Two calls with the same seed produce identical outputs (deterministic purity).
+
+    :param ot: Whether to exercise the Hungarian-match path; ``ot=False`` isolates
+        determinism from any per-call generator consumption inside the matcher.
+    """
+    raw = _make_raw(seed=11)
+    first = prepare_batch(
+        raw,
+        mean=None,
+        std=None,
+        rescale_params=True,
+        ot=ot,
+        generator=torch.Generator().manual_seed(42),
+    )
+    second = prepare_batch(
+        raw,
+        mean=None,
+        std=None,
+        rescale_params=True,
+        ot=ot,
+        generator=torch.Generator().manual_seed(42),
+    )
+    for key in ("params", "noise", "mel_spec"):
+        torch.testing.assert_close(first[key], second[key], atol=0.0, rtol=0.0)
+    # The unread slots must stay None across calls, not silently materialize.
+    assert first["audio"] is None and second["audio"] is None
+    assert first["m2l"] is None and second["m2l"] is None
+
+
+def test_prepare_batch_does_not_mutate_raw_inputs() -> None:
+    """``prepare_batch`` performs no in-place writes to the ``raw`` arrays."""
+    raw = _make_raw(read_mel=True, read_audio=True, seed=5)
+    snapshot = {
+        "param_array": _unwrap_array(raw["param_array"]).copy(),
+        "mel_spec": _unwrap_array(raw.get("mel_spec")).copy(),
+        "audio": _unwrap_array(raw.get("audio")).copy(),
+    }
+    prepare_batch(
+        raw,
+        mean=np.zeros((2, 4, 6), dtype=np.float32),
+        std=np.ones((2, 4, 6), dtype=np.float32),
+        rescale_params=True,
+        ot=True,
+        generator=torch.Generator().manual_seed(0),
+    )
+    np.testing.assert_array_equal(_unwrap_array(raw["param_array"]), snapshot["param_array"])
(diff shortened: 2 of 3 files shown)


ktinubu:
## Verification Results

The promise of this PR is a **function-level no-op**: extracting `prepare_batch` and rewiring `__getitem__` must leave batch output byte-identical. Per design §9 ("parity is at the function level, not the stream level") the correct altitude is the targeted goldens, not a full training run. All runs serial (`-n0 -m "not slow"`); the full `tests/data/` dir hangs under xdist on slow/VST tests.

- [x] **Post-refactor `__getitem__` output is byte-identical to a pre-refactor golden** *(Level 2 — invokes the real `__getitem__` and compares to a captured pre-refactor reference; a failure would mean the extraction changed batch contents, which is the one thing this PR must not do)*
  ```bash
  $ uv run pytest "tests/data/test_surge_datamodule.py::TestGetitemNoOpAfterExtraction" -n0 -q --no-header -p no:cacheprovider -v
  tests/data/test_surge_datamodule.py::TestGetitemNoOpAfterExtraction::test_getitem_unchanged_after_extraction[0] PASSED [ 50%]
  tests/data/test_surge_datamodule.py::TestGetitemNoOpAfterExtraction::test_getitem_unchanged_after_extraction[1] PASSED [100%]
  ============================== 2 passed in 0.33s ===============================
  ```

- [x] **`prepare_batch` is pure and pinned (output, purity, feature flags, rescale, OT)** *(Level 2 — runs the extracted function against a frozen reference under a fixed `torch.Generator`; failure = the per-batch math drifted)*
  ```bash
  $ uv run pytest tests/data/test_prepare_batch.py -n0 -q --no-header -p no:cacheprovider
  ============================== 14 passed in 0.74s ===============================
  ```
  Covers: golden equivalence, same-seed determinism, no-mutation-of-`raw`, modality-slot/flag mapping, rescale `[0,1]→[-1,1]`, OT match vs `_hungarian_match`, and a reference-free brute-force OT-optimality check.

- [x] **`__len__` bug fix: an audio-less shard counts rows off `param_array` (no `KeyError`)** *(Level 2 — adversarial; the PR changes `__len__` from the optional `audio` column to always-present `param_array`. This exercises the audio-less path that previously would `KeyError`)*
  ```bash
  $ uv run pytest "tests/data/test_surge_datamodule.py::TestVSTDatasetH5Mode::test_len_counts_param_array_rows_without_audio_column" -n0 -q --no-header -p no:cacheprovider -v
  tests/data/test_surge_datamodule.py::TestVSTDatasetH5Mode::test_len_counts_param_array_rows_without_audio_column PASSED [100%]
  ============================== 1 passed in 0.30s ===============================
  ```

- [ ] **Full training-run reproducibility under the new seeding surface** *(Level 1 — not run here. The noise RNG moved from global `torch.randn_like` to a per-dataset `torch.Generator` (see PR "Reviewer note"). This PR does not claim end-to-end training parity — the function-level goldens are the contract — and per-worker/global-seed reconciliation is Phase 2 (#1740) scope. To check manually: `uv run pytest tests/test_train.py -m "not slow"` and a seeded `train` smoke run, weighing the documented seeding change.)*

**Note (test-hardening recommendation, non-blocking):** `prepare_batch` outputs are asserted for dtype/contiguity/bounds but not finiteness. Recommend adding a NaN/Inf assertion on `noise`/`mel_spec` in `tests/data/test_prepare_batch.py` to harden the pure-transform contract (flagged in pre-PR review, ml-test).

src/synth_setter/data/surge_datamodule.py
```diff
@@ -111,6 +192,11 @@ def __init__(
         self.batch_size = batch_size
         self.ot = ot
 
+        # Entropy-seeded so production noise stays random; tests seed it to pin a
+        # batch. num_workers>0 forks share this state and draw correlated noise,
+        # matching the pre-refactor global ``randn_like`` (re-seed is future work).
+        self.generator = torch.Generator()
+
```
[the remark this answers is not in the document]
ktinubu: Good catch — confirmed empirically that two fresh `torch.Generator()` share the same fixed default seed (`67280421310721`), so it was *not* entropy-seeded as the comment claimed. Fixed in 259d631b: added `self.generator.seed()` to draw a nondeterministic OS seed; tests still pin via `manual_seed`. Per-worker re-seed remains Phase 2 (#1740).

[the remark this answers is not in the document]
ktinubu: Good catch — confirmed empirically that two fresh `torch.Generator()` share the same fixed default seed (`67280421310721`), so it was *not* entropy-seeded as the comment claimed. Fixed in 259d631b: added `self.generator.seed()` to draw a nondeterministic OS seed; tests still pin via `manual_seed`. Per-worker re-seed remains Phase 2 (#1740).

src/synth_setter/data/surge_datamodule.py
```diff
@@ -19,6 +19,87 @@
...
+    m2l_raw = raw.get("music2latent")
+    m2l = torch.from_numpy(m2l_raw).to(dtype=torch.float32) if m2l_raw is not None else None
+
+    param_array = raw["param_array"]
+    if rescale_params:
+        param_array = param_array * 2 - 1
+    params = torch.from_numpy(param_array).to(dtype=torch.float32)
+    noise = torch.randn(params.shape, generator=generator, dtype=torch.float32)
+    if ot:
+        noise, params, mel_spec, audio = _hungarian_match(noise, params, mel_spec, audio)
```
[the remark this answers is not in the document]
ktinubu: Fixed in 259d631b. (`randn_like` doesn't accept a `generator`, so the equivalent device-safe form is `torch.empty_like(params).normal_(generator=generator)` — matches params' device/dtype/shape and still uses the seeded generator. Goldens still pass bit-for-bit.)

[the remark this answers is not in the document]
ktinubu: Fixed in 259d631b. (`randn_like` doesn't accept a `generator`, so the equivalent device-safe form is `torch.empty_like(params).normal_(generator=generator)` — matches params' device/dtype/shape and still uses the seeded generator. Goldens still pass bit-for-bit.)

tests/data/test_prepare_batch.py
```diff
@@ -0,0 +1,410 @@
...
+        m2l = None
+
+    param_raw = _unwrap_array(raw["param_array"])
+    if rescale_params:
+        param_raw = param_raw * 2 - 1
+    param_array = torch.from_numpy(param_raw).to(dtype=torch.float32)
+    torch.manual_seed(seed)
+    noise = torch.randn_like(param_array)
+    if ot:
+        noise, param_array, mel_spec, audio = _hungarian_match(noise, param_array, mel_spec, audio)
```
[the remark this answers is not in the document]
ktinubu: Fixed in 259d631b: the golden noise draw is now inside `with torch.random.fork_rng():` so `torch.manual_seed` can't leak global RNG state into later tests.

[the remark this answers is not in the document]
ktinubu: Fixed in 259d631b: the golden noise draw is now inside `with torch.random.fork_rng():` so `torch.manual_seed` can't leak global RNG state into later tests.

tests/data/test_surge_datamodule.py
```diff
@@ -649,9 +640,92 @@ def test_returned_dict_always_exposes_full_key_set(self, single_h5: Path) -> Non
...
+        m2l = None
+
+    param_array = dataset._index_dataset(shard["param_array"], idx)
+    if dataset.rescale_params:
+        param_array = param_array * 2 - 1
+    param_array = torch.from_numpy(param_array).to(dtype=torch.float32)
+    torch.manual_seed(seed)
+    noise = torch.randn_like(param_array)
+    if dataset.ot:
+        noise, param_array, mel_spec, audio = _hungarian_match(noise, param_array, mel_spec, audio)
```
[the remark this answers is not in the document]
ktinubu: Fixed in 259d631b: wrapped the golden noise draw in `torch.random.fork_rng()` so the seed no longer leaks into subsequent tests.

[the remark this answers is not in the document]
ktinubu: Fixed in 259d631b: wrapped the golden noise draw in `torch.random.fork_rng()` so the seed no longer leaks into subsequent tests.

tests/data/test_prepare_batch.py
```diff
@@ -0,0 +1,410 @@
...
+        rescale_params=True,
+        ot=True,
+        generator=torch.Generator().manual_seed(0),
+    )
+    noise = _unwrap(out["noise"])
+    params = _unwrap(out["params"])
+    chosen_cost = torch.linalg.vector_norm(noise - params, dim=1).sum().item()
+    for perm in itertools.permutations(range(_BATCH)):
+        cost = torch.linalg.vector_norm(noise[list(perm)] - params, dim=1).sum().item()
+        assert chosen_cost <= cost + 1e-5
```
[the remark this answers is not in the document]
ktinubu: Fixed in 259d631b: the brute-force optimality check now uses a 5-row batch (120 permutations) instead of 8 (40320). The pairing is batch-size-independent, so it still exercises the contract while staying fast in the default suite.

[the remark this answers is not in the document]
ktinubu: Fixed in 259d631b: the brute-force optimality check now uses a 5-row batch (120 permutations) instead of 8 (40320). The pairing is batch-size-independent, so it still exercises the contract while staying fast in the default suite.

src/synth_setter/data/surge_datamodule.py
```diff
@@ -59,10 +142,8 @@ def close(self) -> None:
         ...
```
khaledtin, line 142: *(anchored at line 142; finding is on line 136, outside the diff hunks)*

**[ml-pipeline:block]** Per-worker noise correlation regression. `self.generator = torch.Generator(); self.generator.seed()` is constructed/seeded once in `__init__` (parent process). With `num_workers > 0` PyTorch forks the dataset into each worker, so every worker inherits the identical generator state and draws the same noise sequence per batch — correlated noise across workers. The pre-refactor `torch.randn_like` (global RNG) was auto-seeded per worker (`base_seed + worker_id`) and did not have this problem. Production datamodule configs run multi-worker (`vst.yaml`/`kosc.yaml`=11, `ksin.yaml`/`fm.yaml`=8, `nsynth.yaml`/`fsd.yaml`=4). The in-code comment defers the fix to Phase 2, but this should be gated on a `worker_init_fn` re-seed (e.g. from `get_worker_info().seed`) before merge, or scoped so multi-worker configs aren't affected.

khaledtin, line 142: *(anchored at line 142; finding is on line 137, outside the diff hunks)*

**[synth-setter:block]** P31: behavior change to `train` ships without a matching `tests/test_train.py` update. The real-read noise path changed from `torch.randn_like` (global RNG) to a `self.generator` re-seeded from OS entropy in `__init__`. `train` (cli/train.py) calls `L.seed_everything(cfg.seed, workers=True)`, which previously made per-batch noise reproducible from `cfg.seed`; after this change the production noise draw ignores `cfg.seed` entirely, so the same `train` config now yields different results for the same input — a behavioral change to an operator-facing entrypoint. The default `datamodule/vst.yaml` (`num_workers: 11`) is what `train` composes, yet no leg was added/updated in `tests/test_train.py`. The new `tests/data/` unit/golden tests do not satisfy P31; the gate is the e2e driver.

khaledtin, line 142: *(anchored at line 142; finding is on line 136, outside the diff hunks)*

**[synth-setter:block]** Pipeline-invariant / determinism regression. With `num_workers=11` the `DataLoader` forks the dataset after `__init__`, so every worker inherits the identical entropy-seeded `self.generator` state and draws correlated/identical noise across workers — the comment itself admits this ("num_workers>0 forks share this state and draw correlated noise; per-worker re-seed is Phase 2"). Knowingly shipping correlated training noise under the default multi-worker config degrades the trained net. Either thread `cfg.seed` into the generator (preserving the old reproducibility) or implement the per-worker re-seed before merge; at minimum pin it with a `test_train.py` leg.

khaledtin, line 142: *(anchored at line 142; finding is on line 132, outside the diff hunks)*

**[ml-test:warn]** The production seeding contract is the riskiest behavior introduced and has no test. `self.generator.seed()` reseeds from entropy at construction; the comment flags that `num_workers>0` forks share generator state and draw correlated noise. Add: (a) a test that two `VSTDataset` instances without manual seeding produce different noise (a bare `torch.Generator()` has a fixed default seed, so a missing `.seed()` call would silently make every run identical and no current test would catch it); and (b) a regression-pinning test or explicit xfail documenting the known worker-correlation behavior so Phase 2 has a contract to flip. The only current `self.generator` coverage goes through `manual_seed` in goldens, bypassing the `.seed()` path.

khaledtin, line 142: *(anchored at line 142; finding is on line 136, outside the diff hunks)*

**[code-health:warn]** `self.generator.seed()` seeds from OS entropy, diverging from sibling datamodules (`kosc`/`ksin`/`fm`) which all use `manual_seed(self.seed)` for reproducibility (consistency). The inline comment justifies the choice and flags the `num_workers>0` correlated-noise gap as Phase 2, so it's deliberate, but the inconsistency with three neighboring datamodules is non-obvious and a future reader may "fix" it the wrong way. Consider a `seed` constructor param (defaulting to entropy) to align the family.

src/synth_setter/data/surge_datamodule.py
```diff
@@ -19,6 +19,89 @@
...
+        ``None`` optional modality drops to ``None`` in the output.
+    :param mean: Mel mean to subtract, broadcasting over the batch axis, or
+        ``None`` to skip normalization.
+    :param std: Mel std to divide by, broadcasting over the batch axis, or
+        ``None`` to skip normalization.
+    :param rescale_params: Whether to map params from ``[0, 1]`` to ``[-1, 1]``.
+    :param ot: Whether to Hungarian-match noise to params (and permute the
+        conditioning rows to follow).
+    :param generator: RNG for the noise draw; seed it to pin the batch.
+    :returns: ``{"mel_spec", "m2l", "params", "noise", "audio"}`` with
```
khaledtin, line 68: **[code-health:warn]** `prepare_batch` takes 6 args (1 positional + 5 keyword-only), exceeding the ~3-arg guideline, and `rescale_params`/`ot` are behavior-selecting boolean flag-args. Mitigated by keyword-only enforcement and the pure-function design, and the args mirror `VSTDataset` attributes 1:1, so it's acceptable as-is — but the function now owns the maximum surface a single function should carry. Future additions should go into a config/params dataclass rather than a 7th argument.

src/synth_setter/data/surge_datamodule.py
```diff
@@ -71,7 +152,9 @@ class VSTDataset(torch.utils.data.Dataset):  # noqa: DOC601, DOC603
     ``ot`` is set). Fake mode synthesises
     random tensors of the configured width without opening any file. ``mean`` and ``std``
     hold the optional saved mel statistics applied during reads (``None`` when unloaded);
-    ``dataset_file`` is the open :class:`ShardFile` handle (``None`` in fake mode).
+    ``dataset_file`` is the open :class:`ShardFile` handle (``None`` in fake mode);
+    ``generator`` is the entropy-seeded RNG threaded into :func:`prepare_batch` for the
+    per-batch noise draw (seed it to pin a batch).
     Storage-format subclasses override ``_open`` to read non-HDF5 shards.
```
khaledtin, line 161: **[ml-pipeline:warn]** `__len__` now counts rows off `param_array` instead of `audio` (a correct fix for audio-less Lance shards, well covered by a regression test), but it silently assumes `param_array`'s row count equals every other column's. There is no length/shape-consistency check across columns at open time, so a mis-built shard with mismatched column lengths would over/under-report `__len__` and drop rows or `IndexError` mid-epoch. Consider an open-time invariant check since this is the one always-present column the whole length contract now hangs on.

src/synth_setter/data/surge_datamodule.py
```diff
@@ -111,6 +194,13 @@ def __init__(
         self.batch_size = batch_size
         self.ot = ot
 
+        # A bare torch.Generator() has a fixed default seed, so .seed() is needed
```
khaledtin: **[synth-setter:warn]** `_get_fake_item` still draws noise from the global RNG (`torch.randn_like`) while the real path now uses `self.generator`. Fake-mode noise stays `seed_everything`-reproducible but real-mode no longer is — an inconsistency that means `fake=true` e2e legs in `test_train.py` cannot catch the real-path regression. Document why fake mode deliberately bypasses the generator; the reproducibility divergence between the two modes is not currently called out.

src/synth_setter/data/surge_datamodule.py
```diff
@@ -19,6 +19,89 @@
...
+    mel_raw = raw.get("mel_spec")
+    if mel_raw is not None:
+        if mean is not None and std is not None:
+            mel_raw = (mel_raw - mean) / std
+        mel_spec = torch.from_numpy(mel_raw).to(dtype=torch.float32)
+    else:
+        mel_spec = None
+
+    m2l_raw = raw.get("music2latent")
+    m2l = torch.from_numpy(m2l_raw).to(dtype=torch.float32) if m2l_raw is not None else None
```
khaledtin, line 84: **[code-health:warn]** The `param_array * 2 - 1` rescale magic constants (and `+1e-5` tolerance in the OT optimality test) are inlined unnamed. Minor: these are the exact pre-refactor literals (no new magic introduced) and the docstring spells out the `[0,1] -> [-1,1]` semantics, so extraction would add little. Pre-existing; noting for awareness.

src/synth_setter/data/surge_datamodule.py
```diff
@@ -19,6 +19,89 @@
...
+
+    param_array = raw["param_array"]
+    if rescale_params:
+        param_array = param_array * 2 - 1
+    params = torch.from_numpy(param_array).to(dtype=torch.float32)
+    # empty_like(params) matches params' device/dtype/shape and normal_ accepts
+    # the generator (randn_like does not), so noise can't land on a foreign device.
+    noise = torch.empty_like(params).normal_(generator=generator)
```
khaledtin, line 92: **[python-style:warn]** `prepare_batch` returns via `dict(mel_spec=..., ...)` rather than a dict literal. No hard Google-style rule and it matches the pre-refactor code, so it is consistent; a dict literal would be marginally clearer for a fixed-key return. Non-blocking.

tests/data/test_prepare_batch.py
```diff
@@ -0,0 +1,415 @@
...
+    untouched = _unwrap(
+        prepare_batch(
+            raw,
+            mean=None,
+            std=None,
+            rescale_params=False,
+            ot=False,
+            generator=torch.Generator().manual_seed(0),
+        )["mel_spec"]
+    )
```
khaledtin, line 277: **[ml-test:warn]** The golden references draw noise from the global RNG (`torch.manual_seed` + `torch.randn_like`) and rely on a seeded `torch.Generator` reproducing it bit-for-bit on CPU. This is an undocumented PyTorch internal that can change across versions, and `atol=rtol=0` would turn any such change into a confusing mass failure across ~6 tests. The goldens could instead draw noise the same way `prepare_batch` does — `torch.randn(shape, generator=torch.Generator().manual_seed(seed))` (verified bit-identical) — eliminating the cross-version coupling while keeping the goldens independent of the production path.

tests/data/test_prepare_batch.py
```diff
@@ -0,0 +1,415 @@
...
+    snapshot = _unwrap_array(raw.get("mel_spec")).copy()
+    prepare_batch(
+        raw,
+        mean=None,
+        std=None,
+        rescale_params=True,
+        ot=True,
+        generator=torch.Generator().manual_seed(0),
+    )
+    np.testing.assert_array_equal(_unwrap_array(raw.get("mel_spec")), snapshot)
```
khaledtin, line 439: *(anchored at line 415; finding is on line 519, outside the diff hunks)*

**[ml-test:warn]** `test_prepare_batch_normalizes_mel_only_when_mean_and_std_set` exercises both-set and both-`None`, but not the mixed cases (`mean` set / `std=None` and vice versa). The production guard is `if mean is not None and std is not None`, so a mutation to `or` would pass every current test. A parametrized mixed-arg case would close this mutation gap.

khaledtin, line 439: *(anchored at line 415; finding is on line 622, outside the diff hunks)*

**[tdd-impl:warn]** `test_prepare_batch_ot_false_passes_through_unpermuted` builds expected noise with `torch.randn(shape, generator=...)` while `prepare_batch` draws via `torch.empty_like(params).normal_(generator=...)`. These are bit-identical on the current PyTorch and the `atol=rtol=0` assertion depends on that. Unlike the Generator-vs-global-RNG equivalence (documented in the module docstring), this `randn` vs `empty_like().normal_()` equivalence is undocumented — a second, unstated cross-version assumption. A one-line comment would harden the test.

tests/data/test_prepare_batch.py
```diff
@@ -0,0 +1,415 @@
...
+    )
+    assert torch.allclose(normalized, torch.full_like(normalized, (3.0 - 1.0) / 2.0))
+
+    untouched = _unwrap(
+        prepare_batch(
+            raw,
+            mean=None,
+            std=None,
+            rescale_params=False,
+            ot=False,
```
khaledtin, line 274: **[synth-setter:warn]** The golden-comparison strategy rests on the documented assumption that a seeded `torch.Generator` reproduces the seeded global RNG bit-for-bit on CPU. This invariant is undertested — a future PyTorch where `Generator().manual_seed(s)` and `torch.manual_seed(s)` diverge would make every golden test silently wrong rather than fail loudly. Consider a direct equivalence assertion as an explicit guard rather than burying it in a docstring caveat.

tests/data/test_surge_datamodule.py
```diff
@@ -787,9 +852,8 @@ def test_offset_is_constant_within_one_epoch(self) -> None:
...
     def test_offset_can_vary_across_epochs(self) -> None:
         """A fresh ``iter()`` redraws the offset — over many epochs we see >1 distinct value."""
-        # Seed Python's random + numpy so the test is deterministic but still
-        # exercises the redraw. Save/restore the global states so other tests
-        # in the same pytest-xdist worker don't see a leaked seed.
+        # Seed global RNGs for determinism; save/restore so xdist workers don't
+        # see a leaked state.
         py_state = random.getstate()
         np_state = np.random.get_state()
         try:
```
khaledtin, line 967: *(anchored at line 859; finding is on line 864, outside the diff hunks)*

**[tdd-impl:warn]** `test_getitem_unchanged_after_extraction` is the only behavior test exercising the refactor's integration point (`__getitem__` → `prepare_batch` → `_read_column`) against a real HDF5 shard, and it's parametrized over only `idx ∈ {0, 1}`. The refactor's point was format-independence, yet no test drives `prepare_batch` through `_read_column`/`_index_dataset` for the list-of-rows or `(start, stop)`-pair index forms that `_read_column`'s docstring advertises. Add a list/sequence-index case so the new slicing contract is pinned for all index shapes it documents.

tests/data/test_surge_datamodule.py
```diff
@@ -746,11 +816,6 @@ def test_indices_cover_full_range_without_duplicates(self) -> None:
         assert sorted(flat) == list(range(batch_size * num_batches))
```
khaledtin, line 926: **[python-style:warn]** `_reference_getitem` reaches into private members (`dataset.dataset_file`, `dataset._index_dataset`) to rebuild the golden. Acceptable for a white-box pinning test that mirrors the production read path, but it couples the test to private internals — if `_index_dataset` is renamed the golden silently diverges. Worth a comment noting the intentional white-box coupling.

tests/data/test_surge_datamodule.py
```diff
@@ -588,9 +580,8 @@ def test_ot_true_calls_hungarian_match_with_all_four_tensors(self, single_h5: Pa
         mock_match.assert_called_once()
```
khaledtin: **[tdd-impl:warn]** `test_ot_true_calls_hungarian_match_with_all_four_tensors` mocks `_hungarian_match` and asserts on call arity + per-slot shapes (interaction style). Acceptable because the new golden-based OT tests provide real behavior backing — flagging only that this mock-based assertion is now redundant with `test_prepare_batch_ot_true_matches_reference_hungarian` and could be retired.

khaledtin (changes requested):
Multi-skill review of PR #1745 — 7 parallel passes (code-health, synth-setter-project-standards, python-style, tdd-implementation, comment-hygiene, ml-data-pipeline, ml-test). Each finding below (BLOCK or WARN) is posted as an individual unresolved inline thread so it can't be scrolled past without an explicit reply or resolution. Findings on files outside the diff are anchored to the line in the diff that causes the staleness or rolled into the review body.

src/synth_setter/data/surge_datamodule.py
ktinubu: Fixed in e9c2ef8. Each dataloader worker now re-seeds its forked generator copy from `get_worker_info().seed` (`base_seed + worker_id`) on first `__getitem__`. Lazy rather than a `worker_init_fn` deliberately: Lightning auto-adds `pl_worker_init_function` only while `worker_init_fn is None`, so installing our own hook would displace the per-worker re-seed of worker *global* RNGs that fake-mode noise still uses. Pinned by `test_dataloader_two_workers_reproduce_and_decorrelate_noise` (real forked 2-worker DataLoader over HDF5: same global seed reproduces the stream bit-for-bit across epochs, and worker 0's first draw differs from worker 1's) plus a worker-sim unit test of the once-per-worker re-seed.

ktinubu: Fixed in e9c2ef8: added `tests/test_train.py::test_train_same_seed_reproduces_noise_stream`. It injects a `NoiseCaptureCallback` (new `tests/helpers/noise_capture.py`) into the composed `datamodule=surge_lance` train cfg and asserts two `train(cfg)` runs under one `cfg.seed` consume a bit-identical noise stream. Capturing the stream (not loss) is deliberate: `surge_ffn`'s loss ignores `batch["noise"]`, so a loss-equality leg would not gate this path. The leg runs `num_workers=0` because forking workers over Lance deadlocks on the parent's tokio threadpool (pre-existing; worker-safe opening is #1740 Phase 2 scope) — the forked-worker path is e2e-covered over HDF5 by `test_dataloader_two_workers_reproduce_and_decorrelate_noise`.

ktinubu: Fixed in e9c2ef8 with both suggested remedies: `__init__` now seeds the generator from a draw off the global RNG, so `L.seed_everything(cfg.seed, workers=True)` governs production noise exactly as the pre-refactor global `randn_like` did (unseeded runs stay random); forked workers re-seed once from their DataLoader-assigned worker seed (decorrelated, still global-seed-derived); and the new `test_train.py` leg pins the contract end-to-end through `train(cfg)`, with real 2-worker fork coverage in `test_dataloader_two_workers_reproduce_and_decorrelate_noise`.

ktinubu: Fixed in e9c2ef8: (a) `test_generator_unseeded_constructions_get_distinct_seeds` guards against regressing to a bare `torch.Generator()` fixed default seed; (b) the worker-correlation gap is now fixed rather than xfail-documented, pinned by `test_getitem_in_worker_reseeds_generator_once_from_worker_seed`; (c) `test_getitem_noise_same_global_seed_reproduces` covers the production (non-`manual_seed`) path end-to-end through `__getitem__`.

ktinubu: Resolved by the seeding fix in e9c2ef8 — the OS-entropy `.seed()` is gone. The generator now seeds from a global-RNG draw, matching the siblings' seed-governed reproducibility model. Skipped adding a `seed` constructor param: kosc/ksin/fm take one because they synthesize whole samples from it, whereas here the only consumer is the noise draw and Phase 2 (#1740 Task 2.2) finalizes the seeding surface — a param now would be YAGNI churn.

src/synth_setter/data/surge_datamodule.py
ktinubu: Left as-is, per the finding's own assessment (keyword-only, pure, args mirror dataset attributes 1:1). Agreed on the forward rule: the next argument goes into a config/params dataclass rather than a 7th parameter — Phase 2's collate_fn refactor (#1740 Task 2.2) is the natural point, noted there.

src/synth_setter/data/surge_datamodule.py
ktinubu: Left as-is for this PR: shard construction is the trust boundary (the datagen pipeline writes all columns row-locked), and the whole `_open`/read path is replaced in Phase 2 (#1740), where an open-time cross-column invariant fits the new reader once instead of twice. Deferred with a note on #1740 so it isn't lost.

src/synth_setter/data/surge_datamodule.py
ktinubu: Fixed in e9c2ef8: `_get_fake_item`'s docstring now documents why fake noise deliberately draws from the global RNG. The reproducibility divergence the finding flags is also gone — after the seeding fix both modes are `seed_everything`-reproducible (real via the constructor's global-RNG draw + per-worker re-seed; fake via Lightning's `pl_worker_init_function`).

src/synth_setter/data/surge_datamodule.py
ktinubu: Left as-is, per the finding: the literals are the exact pre-refactor values, the docstring spells out the `[0,1] -> [-1,1]` semantics, and a named constant would add indirection without information. Happy to extract if a second consumer appears.

src/synth_setter/data/surge_datamodule.py
ktinubu: Fixed in e9c2ef8: `prepare_batch` (and `_get_fake_item`, for consistency) now return dict literals instead of `dict(...)`.

tests/data/test_prepare_batch.py
ktinubu: Fixed in e9c2ef8: `_reference_prepare_batch` (and `_reference_getitem` in test_surge_datamodule.py) now draw noise via `torch.randn(shape, generator=torch.Generator().manual_seed(seed))` — no global-RNG/`fork_rng` coupling remains in the goldens. The residual cross-API assumption is pinned loudly by the new `test_noise_draw_apis_same_seed_bit_identical`, so a PyTorch change fails one guard test instead of a mass golden diff.

tests/data/test_prepare_batch.py
ktinubu: Fixed in e9c2ef8: the normalization test is parametrized over all four `(mean_set, std_set)` combinations, so an `and -> or` mutation in the production guard now fails the two mixed cases.

ktinubu: Fixed in e9c2ef8: the `randn` vs `empty_like().normal_()` equivalence is now asserted directly by `test_noise_draw_apis_same_seed_bit_identical`, and the `ot=False` test carries a one-line comment pointing at that guard.

tests/data/test_prepare_batch.py
ktinubu: Fixed in e9c2ef8: added `test_noise_draw_apis_same_seed_bit_identical`, a direct three-way bit-equality assertion (seeded-Generator `randn` / `empty_like().normal_()` / seeded global RNG). A future PyTorch divergence now fails this one test loudly instead of silently invalidating the goldens.

tests/data/test_surge_datamodule.py
ktinubu: Fixed in e9c2ef8: `test_getitem_unchanged_after_extraction` is now parametrized over `[0, 1, [0, 2, 5, 7], (2, 6)]`, pinning the golden across the row-list and `(start, stop)` index forms `_index_dataset` documents, on top of the two batch indices.

tests/data/test_surge_datamodule.py
ktinubu: Fixed in e9c2ef8: `_reference_getitem`'s docstring now states the intentional white-box coupling to `dataset.dataset_file` / `_index_dataset` and that renaming those members must update the helper in the same change.

tests/data/test_surge_datamodule.py
ktinubu: Fixed in e9c2ef8: retired `test_ot_true_calls_hungarian_match_with_all_four_tensors` as redundant with the golden `test_prepare_batch_ot_true_matches_reference_hungarian` + the reference-free OT-optimality check. Kept `test_ot_with_disabled_modalities_passes_none_through`, which pins the None-slot contract the goldens don't.

src/synth_setter/data/surge_datamodule.py
```diff
@@ -224,6 +320,23 @@ def _index_dataset(self, ds: ShardColumn, idx: int | Sequence[int] | np.ndarray)
...
+        derived from the global RNG, keeping ``seed_everything`` in charge. Done
+        lazily instead of via ``worker_init_fn`` so Lightning's auto-added
+        ``pl_worker_init_function`` (gated on ``worker_init_fn is None``) stays.
+        """
+        if self._worker_reseed_done:
+            return
+        self._worker_reseed_done = True
+        worker_info = torch.utils.data.get_worker_info()
+        if worker_info is not None:
+            self.generator.manual_seed(worker_info.seed)
```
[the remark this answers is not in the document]
ktinubu: Good catch — real bug: a parent-process read before the fork would have latched the flag and disarmed the worker re-seed. Fixed in 1e17346: `_reseed_generator_in_worker` now checks `get_worker_info()` first and latches only inside a worker, so parent reads keep the re-seed armed. Regression-pinned by `test_parent_read_before_fork_does_not_disarm_worker_reseed` (parent read with real `None` worker info, then a simulated worker read that must re-seed from the worker seed).

src/synth_setter/data/surge_datamodule.py
```diff
@@ -19,6 +19,89 @@
...
+# DOC601/DOC603: pydoclint can't read sphinx ``:ivar:`` docs, so TypedDict keys
+# are documented in the docstring body instead.
+class RawBatch(TypedDict):  # noqa: DOC601, DOC603
+    """One batch of read shard columns, the input to :func:`prepare_batch`.
+
+    Shapes are ``(batch, ...)``: ``param_array`` is ``(batch, num_params)`` and
+    always present; ``mel_spec`` is ``(batch, channels, n_mels, n_frames)``,
+    ``music2latent`` is ``(batch, latent_dim, n_frames)``, ``audio`` is
+    ``(batch, channels, samples)``. Each optional key is ``None`` when its read
+    flag is off.
```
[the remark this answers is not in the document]
ktinubu: Fixed in 098472b: the RawBatch docstring now states an optional key whose read flag is off may be stored as `None` or omitted entirely, matching the `NotRequired[... | None]` types and `prepare_batch`'s `.get()` reads. (`__getitem__` itself always stores explicit `None`s, but the TypedDict contract admits omission and the docs now say so.)

src/synth_setter/data/surge_datamodule.py
```diff
@@ -111,6 +196,12 @@ def __init__(
         self.batch_size = batch_size
         self.ot = ot
 
+        # Seeded from a global-RNG draw so seed_everything governs noise (a bare
+        # Generator() repeats a fixed default seed); workers re-seed on first read.
+        self.generator = torch.Generator()
+        self.generator.manual_seed(int(torch.randint(2**63 - 1, (1,)).item()))
+        self._worker_reseed_done = False
```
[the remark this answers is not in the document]
ktinubu: Fixed in 098472b: the constructor now skips the global-RNG seed draw in fake mode (`if not fake:`), leaving the global stream untouched — the generator is unused there, so it stays at the default seed rather than drawing OS entropy. Pinned by `test_fake_mode_construction_leaves_global_rng_untouched` (fork_rng equality of a post-construction draw), and `test_generator_unseeded_constructions_get_distinct_seeds` now exercises the real-read path it was meant to guard.

src/synth_setter/data/surge_datamodule.py
```diff
@@ -232,44 +348,38 @@ def __getitem__(self, idx: int | Sequence[int] | np.ndarray) -> dict[str, torch.
...
-            param_array = param_array * 2 - 1
-        param_array = torch.from_numpy(param_array).to(dtype=torch.float32)
-        noise = torch.randn_like(param_array)
-        if self.ot:
-            noise, param_array, mel_spec, audio = _hungarian_match(
-                noise, param_array, mel_spec, audio
-            )
+    def _read_column(
+        self, name: str, idx: int | Sequence[int] | np.ndarray, read: bool
+    ) -> np.ndarray | None:
```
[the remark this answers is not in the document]
ktinubu: Fixed in 098472b: `_index_dataset`, `__getitem__`, and `_read_column` now annotate `idx: int | tuple[int, int] | Sequence[int] | np.ndarray`, making the `(start, stop)` form explicit rather than implied by `Sequence[int]`.

tests/helpers/noise_capture.py
```diff
@@ -0,0 +1,45 @@
...
+        :param trainer: Unused; Lightning hook signature.
+        :param pl_module: Unused; Lightning hook signature.
+        :param outputs: Unused; Lightning hook signature.
+        :param batch: The training batch; its ``noise`` entry is captured.
+        :param batch_idx: Unused; Lightning hook signature.
+        """
+        noise = batch["noise"]
+        assert noise is not None
+        # clone(): worker batches arrive in shared memory the loader may reuse.
+        NoiseCaptureCallback.captured.append(noise.detach().clone())
```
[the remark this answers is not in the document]
ktinubu: Fixed in da7a5ba: captures are now `noise.detach().clone().cpu()`, so a GPU training test reusing the callback won't retain device tensors on the class-level list; the bit-equality comparisons are unaffected.

Issue #1697, opened 2026-06-14
feat(data-pipeline): add audio_mp3 preview column to Lance shards


ktinubu:
Add a Lance-only, per-row MP3 preview column (`audio_mp3`, `large_binary`) so dataset viewers can audition a patch without decoding the raw float16 waveform.

Scope:
- Encode each rendered row to an inline MP3 at worker generation time, in the Lance writer's flush (`make_lance_dataset`) — Lance-only, so HDF5/WDS and the resharder are untouched.
- Declare `audio_mp3` (`large_binary`, mime `audio/mpeg`) in the Lance schema beside the fixed-shape tensors; it stays out of `DATASET_FIELD_NAMES`.
- Training datamodule skips the non-tensor column; worker shard validation asserts presence + `large_binary`.
- A lossy convenience artifact, never a training input.

Issue #1682, opened 2026-06-13
feat(data-pipeline): mp3 audio-preview column in finalized Lance dataset


ktinubu:
## Problem

Finalized Lance split files (`train/val/test.lance`) carry only the lossless `audio` tensor column (float16 PCM). Open-source Lance dataset visualizers can't auto-play raw tensor columns as audio, so there's no quick audible preview when browsing a dataset.

## Proposal

During the **finalize** stage, attach an `audio_mp3` column to each finalized split row — a lossy MP3 encoded from that row's `audio`, tagged `mime_type: audio/mpeg`. Lance viewers can then play a per-row preview.

Scope:
- Finalize-only; per-shard files are unchanged.
- The MP3 is a convenience artifact, never a training input — the `audio` tensor column remains the source of truth.
- The training data loader (`LanceShardFile`) must tolerate the new non-tensor column (it projects tensor columns by name).

Refs #1600 (Lance single-file shard output format).

PR #1750, opened 2026-06-17
feat(data-pipeline): add audio_uuid column to preview-columns CLI

ktinubu:
## What

Extends the post-write Lance augmenter introduced in #1718 to backfill a
**second** derived column alongside the MP3 preview, and renames it so the name
reflects both outputs:

- **CLI** `synth-setter-add-mp3-audio` → `synth-setter-add-preview-columns`
- **Module** `add_mp3_audio.py` → `add_preview_columns.py`
- **Function** `add_mp3_audio_column` → `add_preview_columns`

New `audio_uuid` column: a deterministic UUIDv5 fingerprint of each row's
`audio` tensor bytes, under a project-scoped namespace
(`uuid5(NAMESPACE_DNS, "synth-setter.tinaudio.com")`). The same rendered
waveform always maps to the same id — content-addressed and stable across
re-runs — giving rows a stable identity for dedup / cross-dataset reference.
It is stored as a plain Arrow `string` column.

Both columns are committed in a **single** `lance.batch_udf` `add_columns`
transaction (derived in one pass over the audio), so an interrupted run leaves
the dataset on its prior version and re-running is safe. Neither column is a
training input; the `audio` tensor stays the source of truth.

The notebook, doc-map, and data-pipeline design doc are renamed/updated to match.

> **Note — CLI rename:** `synth-setter-add-mp3-audio` shipped one release ago
> (8.39.0, #1718) and is removed here in favor of the clearer combined name.

## Why

The augmenter already walks every row's audio to encode the MP3; deriving the
uuid in the same pass adds a stable row identity at negligible cost, in the same
atomic transaction. It sits alongside the other post-write Lance augmenters and
follows the same §5 stage contract — adds columns to existing shards without
modifying earlier stages.

## Test plan

- 33 unit/integration tests in `tests/pipeline/data/test_add_preview_columns.py`
  (renamed from `test_add_mp3_audio.py`), including new `audio_uuid` coverage:
  determinism for equal bytes, layout-independence, a **pinned** UUID value
  (namespace + hex-input format contract), version-5 canonical form, per-row
  match against the pure helper through a real Lance round-trip, and the
  **partial** half-add guard (an `audio_uuid`-only dataset still raises).
- Full `tests/pipeline/data/` suite green (157 tests); `make format` clean.
- Pre-PR gate: `/repo-review-full-no-comments` — **0 BLOCK** across 9 skills
  (structlog logging and a factually-loose uuid5 comment fixed from the first
  pass; remaining findings are advisory style/coverage nits).

Refs #1697
Refs #1682

### DIFF
--- a/src/synth_setter/pipeline/data/add_preview_columns.py
+++ b/src/synth_setter/pipeline/data/add_preview_columns.py
@@ -5,8 +5,8 @@ Backfills two derived columns onto a dataset written by
 single Lance ``add_columns`` transaction:
 
 - ``audio_mp3`` — each row's ``audio`` tensor encoded to a CBR MP3 (pedalboard),
-  stored as a Lance blob v2 column tagged ``mime_type: audio/mpeg`` so Lance
-  viewers auto-play a per-row preview.
+  stored as an Arrow binary column tagged ``mime_type: audio/mpeg`` so viewers
+  can scan the table through DuckDB without Lance blob support.
 - ``audio_uuid`` — a deterministic UUIDv5 fingerprint of the same ``audio`` bytes,
   so the same rendered waveform always maps to the same id (content-addressed).
 
@@ -99,7 +99,7 @@ def _encode_preview_columns(
     :param batch: Record batch projecting the ``audio`` fixed-shape tensor column.
     :param sample_rate: Playback rate in Hz every row is encoded at.
     :param bitrate_kbps: Constant MP3 bitrate in kbps for every row.
-    :returns: A two-column batch (``audio_mp3`` blob array, ``audio_uuid`` string
+    :returns: A two-column batch (``audio_mp3`` binary array, ``audio_uuid`` string
         array), one cell per input row, in batch row order.
     :raises ValueError: The ``audio`` column is not a fixed-shape tensor column,
         or a row fails to encode (the message names the offending row index).
@@ -108,7 +108,7 @@ def _encode_preview_columns(
     if not isinstance(column, pa.FixedShapeTensorArray):
         raise ValueError(f"{AUDIO_FIELD!r} must be a fixed-shape tensor column, got {column.type}")
     rows = column.to_numpy_ndarray()
-    mp3_blobs = []
+    mp3_payloads = []
     uuids = []
     for row_index, row in enumerate(rows):
         try:
@@ -119,10 +119,10 @@ def _encode_preview_columns(
             ) from exc
         # Append as a pair only after the encode succeeds, so a failed row never
         # leaves the two columns at mismatched lengths.
-        mp3_blobs.append(mp3)
+        mp3_payloads.append(mp3)
         uuids.append(audio_uuid(row))
     return pa.record_batch(
-        [lance.blob_array(mp3_blobs), pa.array(uuids, type=pa.string())],
+        [pa.array(mp3_payloads, type=pa.binary()), pa.array(uuids, type=pa.string())],
         names=[AUDIO_MP3_FIELD, AUDIO_UUID_FIELD],
     )
 
@@ -156,13 +156,13 @@ def add_preview_columns(
     sample_rate = read_shard_metadata(dataset.schema).sample_rate
     preview_schema = pa.schema(
         [
-            lance.blob_field(AUDIO_MP3_FIELD).with_metadata(_MP3_FIELD_METADATA),
+            pa.field(AUDIO_MP3_FIELD, pa.binary()).with_metadata(_MP3_FIELD_METADATA),
             pa.field(AUDIO_UUID_FIELD, pa.string()),
         ]
     )
 
     # output_schema skips Lance's first-batch inference probe (avoids a double
-    # encode) and carries the blob type + mime metadata onto the new columns.
+    # encode) and carries the binary type + mime metadata onto the new columns.
     @lance.batch_udf(output_schema=preview_schema)
     def _to_preview(batch: pa.RecordBatch) -> pa.RecordBatch:
         return _encode_preview_columns(batch, sample_rate, bitrate_kbps)
--- a/src/synth_setter/pipeline/data/add_preview_columns.py
+++ b/src/synth_setter/pipeline/data/add_preview_columns.py
@@ -25,25 +25,25 @@ import click
 import lance
 import numpy as np
 import pyarrow as pa
+import structlog
 from pedalboard.io import AudioFile
 
 from synth_setter.data.vst.shapes import AUDIO_FIELD
 from synth_setter.pipeline import r2_io
 from synth_setter.pipeline.data.lance_shard import read_shard_metadata
 
+logger = structlog.get_logger(__name__)
+
 AUDIO_MP3_FIELD = "audio_mp3"
 AUDIO_UUID_FIELD = "audio_uuid"
 
 DEFAULT_MP3_BITRATE_KBPS = 128
 
-# Viewer hint so Lance UIs offer per-row playback; a repo convention, not a
-# Lance-defined contract. Arrow field metadata is a bytes->bytes map.
+# Viewer hint for Lance UIs — a repo convention, not a Lance-defined contract.
 _MP3_FIELD_METADATA: dict[bytes, bytes] = {b"mime_type": b"audio/mpeg"}
 
-# Project-scoped UUIDv5 namespace, derived deterministically from a DNS name so
-# the constant is reproducible and self-documenting (no opaque literal). UUIDv5
-# hashes (namespace, name) with SHA-1, so a fixed namespace makes audio_uuid a
-# pure function of the audio bytes.
+# DNS-derived namespace so the constant is reproducible; a fixed namespace makes
+# audio_uuid a pure function of the audio bytes (same input -> same id).
 _AUDIO_UUID_NAMESPACE = uuid.uuid5(uuid.NAMESPACE_DNS, "synth-setter.tinaudio.com")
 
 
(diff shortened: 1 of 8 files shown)


ktinubu:
## Verification Results

Exercised the **installed console script** end-to-end on a real Lance dataset built with the same schema `generate-dataset`/`finalize-dataset` emit (`audio`/`mel_spec`/`param_array` tensors + embedded `ShardMetadata`). Commands are reproducible against the PR branch HEAD (`9af7196`).

- [x] **`synth-setter-add-preview-columns` adds both `audio_mp3` and `audio_uuid` to a Lance dataset in one pass** *(Level 1 — promise is the end-to-end column add; ran the real entry point, not the library function)*
  ```
  $ uv run synth-setter-add-preview-columns "$URI" --bitrate-kbps 128
  [info] added_preview_columns columns=['audio_mp3', 'audio_uuid'] rows=4 uri=.../shard-000000.lance
  added 'audio_mp3' and 'audio_uuid' columns to .../shard-000000.lance   (exit=0)

  schema_after: ['audio', 'mel_spec', 'param_array', 'audio_mp3', 'audio_uuid']
  mp3_mime:     {b'mime_type': b'audio/mpeg'}
  decoded mp3[0]: rate=16000 channels=2 frames=5184   # blob is a real, decodable MP3 at the metadata sample rate
  ```

- [x] **`audio_uuid` is a content-addressed UUIDv5 fingerprint of the audio bytes, preserved through the Lance round-trip** *(Level 1 — adversarial: recomputed the uuid from the *stored* audio column and compared to the *stored* uuid column; a random/non-deterministic or wrong-source id would diverge)*
  ```
  stored_uuids: ['db809a55-…cb43', 'a8a63cf7-…188f', 'ce4f5f4e-…63c3a', 'bdf45eec-…0688']
  recomputed:   ['db809a55-…cb43', 'a8a63cf7-…188f', 'ce4f5f4e-…63c3a', 'bdf45eec-…0688']
  assert stored == recomputed   -> PASS
  assert len(set(stored)) == 4  -> PASS   # distinct waveforms -> distinct ids
  ```

- [x] **CLI renamed: the new entry point resolves and the old one is gone** *(Level 2 — queried the installed `console_scripts` entry points; promise is the rename, not end-to-end)*
  ```
  $ uv run python -c "from importlib.metadata import entry_points; ..."
  add-preview-columns: ['synth_setter.pipeline.data.add_preview_columns:main']
  add-mp3-audio (should be empty): []
  ```

- [x] **Idempotency / crash-safety guard: re-running on an already-augmented dataset fails fast** *(Level 2 — adversarial: the §5 contract claims re-runs are safe; invoked the CLI a second time and checked the exit code + message)*
  ```
  $ uv run synth-setter-add-preview-columns "$URI"
  Error: dataset at .../shard-000000.lance already has preview column(s): ['audio_mp3', 'audio_uuid']
  rerun_exit=1
  ```
  The partial half-add case (`audio_uuid`-only dataset) is pinned by `test_add_preview_columns_partial_existing_column_raises`.

CI (full suite, lint, pre-commit) runs separately on the PR — not duplicated here.

src/synth_setter/pipeline/data/add_preview_columns.py
```diff
@@ -62,68 +75,105 @@ def encode_audio_to_mp3(audio: np.ndarray, sample_rate: int, bitrate_kbps: int)
...
     :param bitrate_kbps: Constant MP3 bitrate in kbps for every row.
-    :returns: One Lance blob array of per-row MP3 bytes, in batch row order.
+    :returns: A two-column batch (``audio_mp3`` blob array, ``audio_uuid`` string
+        array), one cell per input row, in batch row order.
     :raises ValueError: The ``audio`` column is not a fixed-shape tensor column,
         or a row fails to encode (the message names the offending row index).
     """
     column = batch.column(AUDIO_FIELD)
     if not isinstance(column, pa.FixedShapeTensorArray):
         raise ValueError(f"{AUDIO_FIELD!r} must be a fixed-shape tensor column, got {column.type}")
```
khaledtin, line 109: **Missing test: the non-tensor-column guard.** This `raise` is never exercised — every existing test either projects a real `FixedShapeTensorArray` (so the branch is skipped) or fails earlier in `add_preview_columns` on the schema check. It's reachable only by calling the helper directly, and it's a real contract (the message names the offending type). Suggest a focused direct-call test:

```python
def test__encode_preview_columns_non_tensor_audio_column_raises() -> None:
    """A non-tensor ``audio`` column is rejected before any encode is attempted."""
    batch = pa.record_batch([pa.array([[1.0, 2.0], [3.0, 4.0]])], names=[AUDIO_FIELD])

    with pytest.raises(ValueError, match="fixed-shape tensor"):
        _encode_preview_columns(batch, _SAMPLE_RATE, 128)
```

Verified against this branch: raises `'audio' must be a fixed-shape tensor column, got list<item: double>`. (`_encode_preview_columns` would need adding to the import block.)

src/synth_setter/pipeline/data/add_preview_columns.py
```diff
@@ -62,68 +75,105 @@ def encode_audio_to_mp3(audio: np.ndarray, sample_rate: int, bitrate_kbps: int)
...
     rows = column.to_numpy_ndarray()
-    encoded = []
+    mp3_blobs = []
+    uuids = []
     for row_index, row in enumerate(rows):
         try:
-            encoded.append(encode_audio_to_mp3(row, sample_rate, bitrate_kbps))
+            mp3 = encode_audio_to_mp3(row, sample_rate, bitrate_kbps)
         except (ValueError, RuntimeError, OSError) as exc:
             raise ValueError(
```
khaledtin, line 117: **Missing test: the per-row encode-failure wrapper.** This `except` → `raise` is the only place the offending **row index** and shape are surfaced, yet nothing exercises it (a valid dataset never trips it). That row-index message is a contract a future reader will rely on when a real shard has one bad row, so it deserves a behaviour test. Drive a real batch and force one row to fail:

```python
def test__encode_preview_columns_row_failure_names_row_index(mocker) -> None:
    """A row that fails to encode raises a ValueError naming that row's index."""
    rows = _sine_rows()  # (3, 1, T) float16
    tt = pa.fixed_shape_tensor(pa.float16(), list(rows.shape[1:]))
    storage = pa.array(rows.reshape(_ROWS, -1).tolist(), pa.list_(pa.float16(), rows[0].size))
    batch = pa.record_batch([pa.FixedShapeTensorArray.from_storage(tt, storage)], names=[AUDIO_FIELD])
    mocker.patch(
        "synth_setter.pipeline.data.add_preview_columns.encode_audio_to_mp3",
        side_effect=[b"ok", RuntimeError("boom"), b"ok"],
    )

    with pytest.raises(ValueError, match="row 1"):
        _encode_preview_columns(batch, _SAMPLE_RATE, 128)
```

Verified against this branch: raises `failed to encode audio row 1 (shape (1, 8)): boom`. This is the one spot where mocking the SUT's collaborator is justified — there's no cheap way to make pedalboard fail on exactly one row, and the assertion is on the *real* wrapped message, not on the mock.

khaledtin (commented):
## Test review — coverage gaps that are *real*, not artifacts

I looked at the Codecov patch report (73.08%, 7 missing lines, all in `add_preview_columns.py`) through the `/code-health` and `/tdd-implementation` lenses before suggesting anything, to avoid padding.

**Most of the "uncovered" block is a tracing artifact, not a gap.** Running coverage locally flags lines **107–124 and 168** — i.e. the *entire* body of `_encode_preview_columns` plus the `_to_preview` UDF call. But those lines genuinely execute: `test_add_preview_columns_adds_uuid_column_matching_per_row_audio` reads the columns back and asserts they equal `audio_uuid(row)` per row. They read as "missing" only because Lance runs the `batch_udf` inside its native/Rust `add_columns` execution, which `coverage.py` can't trace. **Adding a happy-path test purely to colour those lines green would be coverage for its own sake — please don't.**

That said, *two error branches inside that function have no test at all* — they're unreachable through a real Lance round-trip (the projection always yields a `FixedShapeTensorArray` and every sine row encodes), so only a direct-call unit test exercises them. Both are behaviour contracts worth pinning (the second one even carries a row-index message that a reviewer would expect to be tested). I prototyped both against the branch and confirmed they hit the intended lines and produce the exact messages. Specifics inline.

Deliberately keeping this to **2 suggestions** — the remaining uncovered lines don't represent untested behaviour.

_(Two console scripts shipped under `[project.scripts]` last release; this PR renames `synth-setter-add-mp3-audio` → `synth-setter-add-preview-columns`. Not a test concern, but worth a line in release notes since it's a breaking CLI removal.)_

src/synth_setter/pipeline/data/add_preview_columns.py
ktinubu: Added in b8acb4d as `test__encode_preview_columns_non_tensor_audio_column_raises` — a direct call with a `list<double>` `audio` column, asserting `match="fixed-shape tensor"` (and `_encode_preview_columns` imported into the test block, as you noted). Verified the raised message is `'audio' must be a fixed-shape tensor column, got list<item: double>`.

src/synth_setter/pipeline/data/add_preview_columns.py
ktinubu: Added in b8acb4d as `test__encode_preview_columns_row_failure_names_offending_row_index`, asserting `match="row 1"` on the real wrapped message. One deviation from the snippet: it injects the mid-batch failure with `monkeypatch` (a small iterator side-effect: `b"ok"`, `RuntimeError("boom")`, `b"ok"`) rather than `mocker` — `pytest-mock` isn't a dependency here and the rest of this suite uses `monkeypatch`. Agreed this is the one spot where stubbing the collaborator is justified, so the assertion stays on the real message, not the stub.

PR #1755, opened 2026-06-28
internal-feat(train): log resolved Hydra cfg at DEBUG

### LINKED ISSUE #33: Add debug logging of resolved config in train()
## Summary
Log the full resolved Hydra config at DEBUG level early in `train()` (in `src/train.py`), right after seed setup and before instantiation begins.

## Motivation
When debugging test failures or unexpected training behavior, it's useful to see the exact resolved config without needing to add temporary print statements.

## Proposed change
Add a single `log.debug()` call around line 61 of `src/train.py`:

```python
log.debug(f"Training config:\n{OmegaConf.to_yaml(cfg, resolve=True)}")
```

This keeps output quiet in normal runs but visible with `log_cli_level=DEBUG` in tests or `-v` flags.

SuperCapper:
## Why

Developers debugging training failures or unexpected behavior have no way to inspect the fully-resolved Hydra config without manually inserting `print(OmegaConf.to_yaml(cfg))` and re-running. Running with `log_cli_level=DEBUG` should be enough to see the full composed config — all overrides applied, all `${...}` interpolations expanded — without touching the source.

Closes #33

## What changed

- In `train()`, inserted `log.debug("Resolved Hydra config:\n%s", _cfg_yaml)` immediately after `L.seed_everything` and before the first `log.info` / object instantiation
- Resolution is attempted with `resolve=True` first; a bare `except Exception` (matching the `# noqa: BLE001` pattern already in `_log_model_artifact`) falls back to `resolve=False` when test-fixture configs contain unresolvable internal Hydra keys — so a resolution failure never aborts training
- Added `test_train_debug_logs_resolved_cfg_before_datamodule` in `tests/test_train.py`: patches `OmegaConf.to_yaml` and `log.debug` in the train module, asserts `to_yaml` is called with `resolve=True`, and verifies `log.debug` receives `"Resolved Hydra config"` as its format string

## Test plan

- Confirmed the debug call appears **after** `seed_everything` and **before** `Instantiating datamodule` in the log when running with `log_cli_level=DEBUG` — the position matters so the full config (including any seed-dependent state) is captured before any object is built
- Confirmed normal runs (without debug logging) produce no extra output — `log.debug` is silent unless debug logging is explicitly enabled
- Deliberately did not test with GPU or multirun: the change is rank-zero-only (via `RankedLogger`) and purely additive, so DDP and multirun paths are unaffected

## Out of scope

Logging the resolved config at other lifecycle points (eval, predict) — not requested by the issue.

---

- [x] One logical change; any breaking changes are called out above.

### DIFF
--- a/src/synth_setter/cli/train.py
+++ b/src/synth_setter/cli/train.py
@@ -190,6 +190,12 @@ def train(cfg: DictConfig) -> tuple[dict[str, Any], dict[str, Any]]:
     if cfg.get("seed"):
         L.seed_everything(cfg.seed, workers=True)
 
+    try:
+        _cfg_yaml = OmegaConf.to_yaml(cfg, resolve=True)
+    except Exception:  # noqa: BLE001 — resolution failures must not abort training
+        _cfg_yaml = OmegaConf.to_yaml(cfg)
+    log.debug("Resolved Hydra config:\n%s", _cfg_yaml)
+
     log.info(f"Instantiating datamodule <{cfg.datamodule._target_}>")
     datamodule: LightningDataModule = hydra.utils.instantiate(cfg.datamodule)
 
--- a/tests/test_train.py
+++ b/tests/test_train.py
@@ -11,6 +11,7 @@ import os
 from collections.abc import Callable
 from contextlib import nullcontext
 from pathlib import Path
+from unittest.mock import ANY, patch
 
 import numpy as np
 import pandas as pd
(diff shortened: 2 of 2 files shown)


SuperCapper:
Hi @ktinubu — this is my first contribution to this project. I have fixed issue #33 by adding a `log.debug` call in `train()` that emits the fully-resolved Hydra config immediately after seed setup and before any object is instantiated. Normal runs are unaffected — the output is only visible when debug logging is enabled (`log_cli_level=DEBUG`). Happy to make any adjustments!

